Issue #68, opened 2018-02-26
Use strongly typed object for Launch options


Meir017:
Right now the launch options is a generic Dictionary, it would be easier to use if it would be a strongly typed object

The options properties are described here - https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/puppeteer/index.d.ts#L1205

would you accept a pull-request?

PR #71, opened 2018-02-27
Improve/launch options

Meir017:
related to #68

### DIFF
--- a/lib/PuppeteerSharp.Tests/TestConstants.cs
+++ b/lib/PuppeteerSharp.Tests/TestConstants.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Collections.Generic;
 
 namespace PuppeteerSharp.Tests
 {
@@ -11,13 +10,13 @@ namespace PuppeteerSharp.Tests
         public const int ChromiumRevision = 526987;
         public static readonly string EmptyPage = $"{ServerUrl}/empty.html";
 
-        public static readonly Dictionary<string, object> DefaultBrowserOptions = new Dictionary<string, object>()
+        public static LaunchOptions DefaultBrowserOptions() => new LaunchOptions
         {
-            { "slowMo", Convert.ToInt32(Environment.GetEnvironmentVariable("SLOW_MO") ?? "0") },
-            { "headless", Convert.ToBoolean(Environment.GetEnvironmentVariable("HEADLESS") ?? "true") },
-            { "args", new[] { "--no-sandbox" }},
-            { "timeout", 0},
-            { "keepAliveInterval", 120}
+            SlowMo = Convert.ToInt32(Environment.GetEnvironmentVariable("SLOW_MO")),
+            Headless = Convert.ToBoolean(Environment.GetEnvironmentVariable("HEADLESS") ?? "true"),
+            Args = new[] { "--no-sandbox" },
+            Timeout = 0,
+            KeepAliveInterval = 120
         };
     }
 }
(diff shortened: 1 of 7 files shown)


lib/PuppeteerSharp/LaunchOptions.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        public bool Dumpio { get; set; }
+
+        public string UserDataDir { get; set; }
+
+        public IDictionary<string, string> Env { get; } = new Dictionary<string, string>();
+
+        public bool Devtools { get; set; }
+
+        public int KeepAliveInterval { get; set; } = 30;
```
Meir017: I didn't find this in the source code of https://github.com/GoogleChrome/puppeteer/blob/master/lib/Launcher.js#L150 but I didn't want to break the code

lib/PuppeteerSharp.Tests/TestConstants.cs
```diff
@@ -11,13 +10,13 @@ public static class TestConstants
...
 
-        public static readonly Dictionary<string, object> DefaultBrowserOptions = new Dictionary<string, object>()
+        public static LaunchOptions DefaultBrowserOptions() => new LaunchOptions
         {
-            { "slowMo", Convert.ToInt32(Environment.GetEnvironmentVariable("SLOW_MO") ?? "0") },
-            { "headless", Convert.ToBoolean(Environment.GetEnvironmentVariable("HEADLESS") ?? "true") },
-            { "args", new[] { "--no-sandbox" }},
-            { "timeout", 0},
-            { "keepAliveInterval", 120}
+            SlowMo = Convert.ToInt32(Environment.GetEnvironmentVariable("SLOW_MO")),
```
Meir017: `Convert.ToInt32` return 0 if argument is `null`

lib/PuppeteerSharp/LaunchOptions.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        public bool HandleSIGINT { get; set; } = true;
+
+        public bool HandleSIGTERM { get; set; } = true;
+
+        public bool HandleSIGHUP { get; set; } = true;
+
+        public int Timeout { get; set; } = 30_000;
+
+        public bool Dumpio { get; set; }
```
kblok: How about `DumpIO`?
See: [.NET example](https://github.com/aspnet/KestrelHttpServer/blob/3fc69dc71f309fe59be48e0d5cf8592c10bfd120/test/Kestrel.Core.Tests/HttpResponseStreamTests.cs#L109) And [Code analysis rule](https://msdn.microsoft.com/en-us/library/ms182256.aspx).

lib/PuppeteerSharp/LaunchOptions.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        public bool Headless { get; set; } = true;
+
+        public string ExecutablePath { get; set; }
+
+        public int SlowMo { get; set; }
+
+        public string[] Args { get; set; } = Array.Empty<string>();
+
+        public bool HandleSIGINT { get; set; } = true;
```
kblok: Let's add only properties we're using right now in the code. There are many guys here which are not being used.

lib/PuppeteerSharp/LaunchOptions.cs
kblok: `KeepAliveInterval` is being used by the WebSocket. https://github.com/kblok/puppeteer-sharp/blob/2dc3661299a8574ba5abb7ba020d55629df7079c/lib/PuppeteerSharp/Connection.cs#L196

30 seconds should be enough. The unit test is setting this value to 120 seconds because you might need to pause and debug the code.

PR #75, opened 2018-03-01
Use csharp markdown code for code in README

Meir017:


kblok:
Yeah, it's great! Thank you!

Issue #76, opened 2018-03-02
Implement puppeteer.launch tests using the evaluate function


kblok:
Implement these tests in https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/test/test.js#L109

* `Puppeteer.launch.should reject all promises when browser is closed`
* `userDataDir option should restore state`
* `userDataDir option should restore cookies`
* `headless should be able to read cookies written by headful`

PR #79, opened 2018-03-02
Implemented Puppeteer.launch tests

Meir017:
solves #3 

This pull request is me just working out loud
I'll remove the `[WIP]` prefix when I'm done with it

### DIFF
--- a/lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
+++ b/lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
@@ -1,4 +1,7 @@
-﻿using System.Threading.Tasks;
+﻿using System;
+using System.Collections.Generic;
+using System.Net;
+using System.Threading.Tasks;
 using Xunit;
 
 namespace PuppeteerSharp.Tests.Puppeteer
@@ -36,6 +39,31 @@ namespace PuppeteerSharp.Tests.Puppeteer
             var response = await page.GoToAsync("https://www.google.com");
             Assert.Equal(response.Status.ToString(), "OK");
 
+            await browser.CloseAsync();
+        }
+
+        [Fact]
+        public async Task NetworkRedirectsShouldReportSecurityDetails()
+        {
+            var options = TestConstants.DefaultBrowserOptions();
+            options.IgnoreHTTPSErrors = true;
+
+            var browser = await PuppeteerSharp.Puppeteer.LaunchAsync(options, TestConstants.ChromiumRevision);
+            var page = await browser.NewPageAsync();
+
+            // TODO:
+            // httpsServer.setRedirect('/plzredirect', '/empty.html');
+
+            var responses = new List<Response>();
+            page.ResponseCreated += (sender, e) => responses.Add(e.Response);
+
+            await page.GoToAsync(TestConstants.HttpsPrefix + "/plzredirect");
+            Assert.Equal(2, responses.Count);
+            Assert.Equal(HttpStatusCode.Redirect, responses[0].Status);
+            var securityDetails = responses[0].SecurityDetails;
+            Assert.Equal("TLS 1.2", securityDetails.Protocol);
+
+            await page.CloseAsync();
             await browser.CloseAsync();
         }
     }
--- a/lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
+++ b/lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
@@ -55,7 +55,7 @@ namespace PuppeteerSharp.Tests.Puppeteer
             var responses = new List<Response>();
             page.ResponseCreated += (sender, e) => responses.Add(e.Response);
 
-            await page.GoToAsync(TestConstants.ServerUrl + "/plzredirect");
+            await page.GoToAsync(TestConstants.HttpsPrefix + "/plzredirect");
             Assert.Equal(2, responses.Count);
             Assert.Equal(HttpStatusCode.Redirect, responses[0].Status);
             var securityDetails = responses[0].SecurityDetails;
--- a/lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
+++ b/lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
@@ -40,7 +40,7 @@ namespace PuppeteerSharp.Tests.Puppeteer
             await browser.CloseAsync();
         }
 
-        [Fact]
+        [Fact(Skip = "test is not part of v1.0.0")]
         public async Task NetworkRedirectsShouldReportSecurityDetails()
         {
             var options = TestConstants.DefaultBrowserOptions();
@@ -98,7 +98,7 @@ namespace PuppeteerSharp.Tests.Puppeteer
             Assert.True(Directory.GetFiles(userDataDir).Length > 0);
             await browser.CloseAsync();
             Assert.True(Directory.GetFiles(userDataDir).Length > 0);
-            Directory.Delete(userDataDir, true);
+            // Directory.Delete(userDataDir, true);
         }
 
         [Fact]
@@ -106,13 +106,13 @@ namespace PuppeteerSharp.Tests.Puppeteer
         {
             var userDataDir = Launcher.GetTemporaryDirectory();
             var options = TestConstants.DefaultBrowserOptions();
-            options.Args = options.Args.Concat( new[] { $"--user-data-dir={userDataDir}" }).ToArray();
-            
+            options.Args = options.Args.Concat(new[] { $"--user-data-dir={userDataDir}" }).ToArray();
+
             var browser = await PuppeteerSharp.Puppeteer.LaunchAsync(options, TestConstants.ChromiumRevision);
             Assert.True(Directory.GetFiles(userDataDir).Length > 0);
             await browser.CloseAsync();
             Assert.True(Directory.GetFiles(userDataDir).Length > 0);
-            Directory.Delete(userDataDir, true);
+            // Directory.Delete(userDataDir, true);
         }
 
         [Fact]
--- a/lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
+++ b/lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.ComponentModel;
+using System.IO;
 using System.Net;
 using System.Threading.Tasks;
 using Xunit;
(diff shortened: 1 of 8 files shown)


Meir017:
#3

lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
```diff
@@ -38,5 +41,30 @@ public PuppeteerLaunchTests()
 
             await browser.CloseAsync();
         }
+
+        [Fact]
+        public async Task NetworkRedirectsShouldReportSecurityDetails()
+        {
+            var options = TestConstants.DefaultBrowserOptions();
+            options.IgnoreHTTPSErrors = true;
```
kblok: Let's try to use objects initializers when possible. I know it's only one prop, but just to keep the style.

Meir017: It's similar to Line 20 - https://github.com/kblok/puppeteer-sharp/pull/79/files/8a5d65d0f466541e0dececaa968dfd4b14022beb#diff-d88690bd1c588dd71b96f61f6c5c2b1fR20 and it's not an initialization, it's just a call to a static method so I'm not sure how we can use objects initializers

lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
```diff
@@ -38,5 +41,30 @@ public PuppeteerLaunchTests()
...
+
+            var browser = await PuppeteerSharp.Puppeteer.LaunchAsync(options, TestConstants.ChromiumRevision);
+            var page = await browser.NewPageAsync();
+
+            // TODO:
+            // httpsServer.setRedirect('/plzredirect', '/empty.html');
```
Meir017: not sure how to do this

lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
kblok: That's true :) :+1:

lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
kblok: The test project is using the TestServer to test a local website. You should run `dotnet ./bin/Debug/netcoreapp2.0/PuppeteerSharp.TestServer.dll` to get the server running.

As the TestServer is an ASP.NET Core project, you could create there a Route `plzredirect` (or a better name) which redirects to `empty.html`.

lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
```diff
@@ -66,5 +63,19 @@ public PuppeteerLaunchTests()
...
             await browser.CloseAsync();
         }
+
+        [Fact]
+        public async Task ShouldRejectAllPromisesWhenBrowserIsClosed()
+        {
+            var browser = await PuppeteerSharp.Puppeteer.LaunchAsync(TestConstants.DefaultBrowserOptions(), TestConstants.ChromiumRevision);
+            var page = await browser.NewPageAsync();
+
+            var neverResolves = page.EvaluateHandle("() => new Promise(r => {})");
```
kblok: I missed this guy on this list https://github.com/kblok/puppeteer-sharp/issues/76.

If this works that would be great because `EvaluateHandle` is working, and I wasn't expecting that till v1.0. But it is far from working, lets ignore this test and move it to that list, so we cover the `EvaluateHandle` as a whole.

lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
```diff
@@ -38,5 +42,53 @@ public PuppeteerLaunchTests()
...
+        }
+
+        [Fact]
+        public async Task ShouldRejectIfExecutablePathIsInvalid()
+        {
+            var options = TestConstants.DefaultBrowserOptions();
+            options.ExecutablePath = "random-invalid-path";
+
+            var exception = await Assert.ThrowsAsync<Exception>(() => PuppeteerSharp.Puppeteer.LaunchAsync(options, TestConstants.ChromiumRevision));
+            Assert.Equal("Failed to create connection", exception.Message);
```
Meir017: the exception is thrown from https://github.com/kblok/puppeteer-sharp/blob/master/lib/PuppeteerSharp/Launcher.cs#L122, this is because the `WaitForEndpoint` method calls the `Start` method on the process object

lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
```diff
@@ -38,5 +42,53 @@ public PuppeteerLaunchTests()
...
+            Assert.IsType<Win32Exception>(exception.InnerException);
```
kblok: This should work also on MacOS (it's Core) I'll take a look tomorrow and see what exception I'd get there

kblok: I'm getting a `Win32Exception` on Mac (weird)  but the message is different. I'm getting a "No such file or directory".

I think that maybe checking for `new FileInfo(chromeExecutable).Exists` would be cheaper than creating a process and waiting for a fail there.

kblok:
Something it's off with this branch, I added you remote in my GitHub repo but I cannot see this branch, odd...

Meir017:
maybe you need to git-fetch first from my remote?

Is the test `should return the default chrome arguments` really needed?
 https://github.com/GoogleChrome/puppeteer/blob/master/test/test.js#L270

lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
```diff
@@ -38,5 +42,53 @@ public PuppeteerLaunchTests()
...
+            options.IgnoreHTTPSErrors = true;
+
+            var browser = await PuppeteerSharp.Puppeteer.LaunchAsync(options, TestConstants.ChromiumRevision);
+            var page = await browser.NewPageAsync();
+            
+            var responses = new List<Response>();
+            page.ResponseCreated += (sender, e) => responses.Add(e.Response);
+
+            await page.GoToAsync(TestConstants.HttpsPrefix + "/plzredirect");
+            Assert.Equal(2, responses.Count);
```
Meir017: the test fails on this line, I commented out this line and it passed, not sure what the second response is supposed to be

kblok:
@Meir017 I know that `should return the default chrome arguments` is quite dumb, but the goal is to show we cover the same Puppeteer covers, let's do it.

lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
kblok: I tested that on Puppeteer, and Puppeteer is returning 2 responses, one with HTTP status 302 (redirect) and the other the empty page.
```JS
(async ()=>{
    const fs = require('fs');
    const path = require('path');
    
    const headless = (process.env.HEADLESS || 'true').trim().toLowerCase() === 'true';
    const slowMo = parseInt((process.env.SLOW_MO || '0').trim(), 10);
    const executablePath = process.env.CHROME;
    
    const defaultBrowserOptions = {
        executablePath,
        slowMo,
        headless,
        args: ['--no-sandbox', '--disable-dev-shm-usage']
    };

    const puppeteer = require("Puppeteer");

    const options = Object.assign({ignoreHTTPSErrors: true}, defaultBrowserOptions);
    const browser = await puppeteer.launch(options);
    const page = await browser.newPage();
    //httpsServer.setRedirect('/plzredirect', '/empty.html');
    const responses =  [];
    page.on('response', response => responses.push(response));
    await page.goto('http://127.0.0.1:8907/plzredirect');
    expect(responses.length).toBe(2);
    expect(responses[0].status()).toBe(302);
    const securityDetails = responses[0].securityDetails();
    expect(securityDetails.protocol()).toBe('TLS 1.2');
    await page.close();
    await browser.close();
})();
```

screen shot 2018-03-04 at 9 05 36 am


Something important @Meir017 is that we're working on Puppeteer v1.0, so it's important to read the test.js in that tag https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/test/test.js#L109

I think this is the best approach because it would be a mess if we try to chase Puppeteer's guys on their master branch. So the approach would be implementing all tests we have on the v.1.0.0 tag and then move on to 1.1 and so on.

Though I think there is something off the response there, this is not on the test list on the issue. #3. But if you want to go and check what's missing (maybe some stupid mistake) go ahead and dig into this.

Meir017:
in v1.0.0 the test `Network redirects should report SecurityDetails` does not exist yet, should I skip it?

kblok:
@Meir017 Yeah, don't remove the code, just ignore the test.

Meir017:
Any update on this?

kblok:
I'll take a look, sorry I didn't notice that you removed the WIP mark.

I'm working on this right now https://github.com/kblok/puppeteer-sharp/pull/82. It turns out that AppVeyor was right, as she always is, and we were leaving lots of chromium processes opened.

I think we will also need to make Browser implement IDisposable so we can use it in a `using` so we can be sure that the browser will always be closed.

Meir017:
I think also the `Page` class should implement `IDisposable`

lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
```diff
@@ -34,5 +39,87 @@ public class PuppeteerLaunchTests
...
+        {
+            var userDataDir = Launcher.GetTemporaryDirectory();
+            var options = TestConstants.DefaultBrowserOptions();
+            options.UserDataDir = userDataDir;
+
+            var browser = await PuppeteerSharp.Puppeteer.LaunchAsync(options, TestConstants.ChromiumRevision);
+            Assert.True(Directory.GetFiles(userDataDir).Length > 0);
+            await browser.CloseAsync();
+            Assert.True(Directory.GetFiles(userDataDir).Length > 0);
+            // Directory.Delete(userDataDir, true);
```
kblok: Let's remove this line if we don't need it

lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
```diff
@@ -34,5 +39,87 @@ public class PuppeteerLaunchTests
...
+        {
+            var userDataDir = Launcher.GetTemporaryDirectory();
+            var options = TestConstants.DefaultBrowserOptions();
+            options.Args = options.Args.Concat(new[] { $"--user-data-dir={userDataDir}" }).ToArray();
+
+            var browser = await PuppeteerSharp.Puppeteer.LaunchAsync(options, TestConstants.ChromiumRevision);
+            Assert.True(Directory.GetFiles(userDataDir).Length > 0);
+            await browser.CloseAsync();
+            Assert.True(Directory.GetFiles(userDataDir).Length > 0);
+            // Directory.Delete(userDataDir, true);
```
kblok: Same here

lib/PuppeteerSharp/Launcher.cs
```diff
@@ -215,6 +219,7 @@ private void CleanUp()
             if (_chromeProcess.Id != 0 && Process.GetProcessById(_chromeProcess.Id) != null)
             {
                 _chromeProcess.Kill();
+                _chromeProcess.WaitForExit();
```
kblok: Could you talk a little bit about this?
I'm working on these lines now and I'd like to know what you've discovered so far.

Meir017: I had an issue with deleting the chrome profile directory in the test, I was receiving an access denied error. It looked like the `Kill` method had a small delay before completing

Meir017: That's why I commented out the delete directory line in the tests

kblok:
I'm merging this guy. I'll take a look at those Delete folders with my process changes if they won't work we'll remove those.

PR #80, opened 2018-03-03
Implement PuppeteerLoaderFixture

kblok:
This class will help us download chromium and set up a WebServer on a CI server.

Wanna take a look @Meir017 ?

### DIFF
--- a/lib/PuppeteerSharp.Tests/PuppeteerLoaderFixture.cs
+++ b/lib/PuppeteerSharp.Tests/PuppeteerLoaderFixture.cs
@@ -1,0 +1,107 @@
+﻿using System;
+using System.Diagnostics;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace PuppeteerSharp.Tests
+{
+    public class PuppeteerLoaderFixture : IDisposable
+    {
+        Process _webServerProcess = null;
+
+        public PuppeteerLoaderFixture()
+        {
+            SetupAsync().GetAwaiter().GetResult();
+        }
+
+        public void Dispose()
+        {
+            int processId = 0;
+
+            try
+            {
+                processId = _webServerProcess.Id;
+            }
+            catch (InvalidOperationException)
+            {
+                //We might get an InvalidOperationException if the process is in an invalid state
+            }
+            catch (Exception ex)
+            {
+                throw new Exception("Unable to kill webserver", ex);
+            }
+
+            if (processId != 0 && Process.GetProcessById(processId) != null)
+            {
+                _webServerProcess.Kill();
+            }
+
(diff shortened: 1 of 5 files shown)


lib/PuppeteerSharp.Tests/PuppeteerLoaderFixture.cs
```diff
@@ -0,0 +1,107 @@
...
+            var downloaderTask = Downloader.CreateDefault().DownloadRevisionAsync(TestConstants.ChromiumRevision);
+            var serverTask = StartWebServerAsync();
+
+            await Task.WhenAll(downloaderTask, serverTask);
+
+        }
+
+        private async Task StartWebServerAsync()
+        {
+            var taskWrapper = new TaskCompletionSource<bool>();
```
Meir017: Would it be easier if we ran the server by adding a project-reference to `PuppeteerSharp.TestServer`?

kblok: I've tried that when I first started the Test project a few months ago. The thing is that `host.Run();` locks the thread and I wasn't able to bypass that.

Any ideas?

Meir017: https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs#L59.
could we just call `host.RunAsync` and save the returned `Task` then we could terminate the server by stoping the task, does this make sense?

kblok: I think I tried that, but I'll give it another shot!

Meir017: Maybe merge this and create a task to fix (if needed) it later in order to get the CI working on all branches?

Meir017:
how is the CI configured? I don't see an `appveyor.yml` file?

lib/PuppeteerSharp.Tests/PuppeteerLoaderFixture.cs
kblok: I tried RunAsync, the problem is that the same process can't run the tests and serve the page at the same time. I'll leave this at it is now, you can play with it if you want to.

kblok:
@Meir017 It CI is configured in AppVeyor, I don't think we need the yml file (yet)

Meir017:
I noticed that the ci runs are taking around 1 Hour to finish, could it be because the server process is still open? I'm guessing that the 1 Hour is the default test run timeout

kblok:
I'll take a look

PR #82, opened 2018-03-06
Fix process leaks

kblok:
Processes are not being killed when the browser is closed

### DIFF
--- a/lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
+++ b/lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
@@ -34,5 +34,21 @@ namespace PuppeteerSharp.Tests.Puppeteer
 
             await browser.CloseAsync();
         }
+
+        [Fact]
+        public async Task ChromeShouldBeClosed()
+        {
+            var options = TestConstants.DefaultBrowserOptions();
+            var launcher = new Launcher();
+            var browser = await launcher.LaunchAsync(options, TestConstants.ChromiumRevision);
+            var page = await browser.NewPageAsync();
+
+            var response = await page.GoToAsync("https://www.google.com");
+            Assert.Equal(response.Status.ToString(), "OK");
+
+            await browser.CloseAsync();
+
+            Assert.Equal(true, launcher.IsChromeClosed);
+        }
     }
 }
--- a/lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
+++ b/lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
@@ -98,7 +98,7 @@ namespace PuppeteerSharp.Tests.Puppeteer
             Assert.True(Directory.GetFiles(userDataDir).Length > 0);
             await browser.CloseAsync();
             Assert.True(Directory.GetFiles(userDataDir).Length > 0);
-            await Launcher.TryDeleteFolder(userDataDir);
+            await Launcher.TryDeleteDirectory(userDataDir);
         }
 
         [Fact]
@@ -112,7 +112,7 @@ namespace PuppeteerSharp.Tests.Puppeteer
             Assert.True(Directory.GetFiles(userDataDir).Length > 0);
             await browser.CloseAsync();
             Assert.True(Directory.GetFiles(userDataDir).Length > 0);
-            await Launcher.TryDeleteFolder(userDataDir);
+            await Launcher.TryDeleteDirectory(userDataDir);
         }
 
         [Fact]
--- a/lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
+++ b/lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
@@ -9,7 +9,7 @@ using Xunit;
 namespace PuppeteerSharp.Tests.Puppeteer
 {
     [Collection("PuppeteerLoaderFixture collection")]
-    public class PuppeteerLaunchTests
+    public class PuppeteerLaunchTests : PuppeteerBaseTest
     {
         [Fact]
         public async Task ShouldSupportIgnoreHTTPSErrorsOption()
@@ -98,7 +98,7 @@ namespace PuppeteerSharp.Tests.Puppeteer
             Assert.True(Directory.GetFiles(userDataDir).Length > 0);
             await browser.CloseAsync();
             Assert.True(Directory.GetFiles(userDataDir).Length > 0);
-            Directory.Delete(userDataDir, true);
+            await Launcher.TryDeleteFolder(userDataDir);
         }
 
         [Fact]
@@ -112,7 +112,7 @@ namespace PuppeteerSharp.Tests.Puppeteer
             Assert.True(Directory.GetFiles(userDataDir).Length > 0);
             await browser.CloseAsync();
             Assert.True(Directory.GetFiles(userDataDir).Length > 0);
-            Directory.Delete(userDataDir, true);
+            await Launcher.TryDeleteFolder(userDataDir);
         }
 
         [Fact]
--- a/lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
+++ b/lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
@@ -98,6 +98,7 @@ namespace PuppeteerSharp.Tests.Puppeteer
             Assert.True(Directory.GetFiles(userDataDir).Length > 0);
             await browser.CloseAsync();
             Assert.True(Directory.GetFiles(userDataDir).Length > 0);
+            Directory.Delete(userDataDir, true);
         }
 
         [Fact]
@@ -111,6 +112,7 @@ namespace PuppeteerSharp.Tests.Puppeteer
             Assert.True(Directory.GetFiles(userDataDir).Length > 0);
             await browser.CloseAsync();
             Assert.True(Directory.GetFiles(userDataDir).Length > 0);
+            Directory.Delete(userDataDir, true);
         }
 
         [Fact]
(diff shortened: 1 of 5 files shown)


lib/PuppeteerSharp/Launcher.cs
```diff
@@ -198,6 +224,24 @@ private void CleanUp()
...
+                IsChromeClosed = true;
+
+                if (_waitForChromeToClose.Task.Status != TaskStatus.RanToCompletion)
+                {
+                    _waitForChromeToClose.SetResult(true);
+                }
+
+                if (_temporaryUserDataDir != null)
+                {
+                    await Task.Factory.StartNew(path => Directory.Delete((string)path, true), _temporaryUserDataDir);
```
Meir017: line 270 calls this method
on line 275 we delete the directory, so either this line or 275 will fail since one of them will delete the directory

kblok: Good catch. The problem I was having is that the process `Exited` event was not being called, your `WaitForExit` did the trick.

This PR is ready to review again. I'll implement IDisposables in another PR

lib/PuppeteerSharp/Launcher.cs
```diff
@@ -208,18 +257,23 @@ private void CleanUp()
             {
                 await _connection.SendAsync("Browser.close", null);
             }
+
+            await _waitForChromeToClose.Task;
         }
 
         private async Task ForceKillChrome()
```
Meir017: This method could be void since it doesn't do anything async stuff

lib/PuppeteerSharp/Launcher.cs
```diff
@@ -35,15 +36,30 @@ public class Launcher
...
         };
+        #endregion
 
-        private bool _chromeClosed;
+        #region Private members
         private Process _chromeProcess;
         private string _temporaryUserDataDir = null;
         private Connection _connection = null;
         private Timer _timer = null;
+        private LaunchOptions _currentOptions;
```
Meir017: why not name it just `_options`? the `current` prefix makes it look like a static member

lib/PuppeteerSharp/Launcher.cs
```diff
@@ -198,6 +224,29 @@ private void CleanUp()
             _chromeProcess?.RemoveExitedEvent();
         }
 
+        private async Task AfterProcessExit()
+        {
+            if (!IsChromeClosed)
+            {
+                if (_currentOptions.LogProcess)
+                {
+                    Console.WriteLine($"PROCESS COUNT: {--_processCount}");
```
Meir017: this should probably be `Interlocked.Decrement`

lib/PuppeteerSharp/Launcher.cs
Meir017: this should probably be `Interlocked.Increment`

lib/PuppeteerSharp/Launcher.cs
```diff
@@ -123,6 +139,12 @@ public class Launcher
                 var keepAliveInterval = options.KeepAliveInterval;
 
                 _connection = await Connection.Create(browserWSEndpoint, connectionDelay, keepAliveInterval);
+
+                if (options.LogProcess)
+                {
+                    Console.WriteLine($"PROCESS COUNT: {++_processCount}");
```
Meir017: this should probably be `Interlocked.Increment`

lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
```diff
@@ -34,5 +34,21 @@ public class PuppeteerLaunchTests
...
+            var launcher = new Launcher();
+            var browser = await launcher.LaunchAsync(options, TestConstants.ChromiumRevision);
+            var page = await browser.NewPageAsync();
+
+            var response = await page.GoToAsync("https://www.google.com");
+            Assert.Equal(response.Status.ToString(), "OK");
+
+            await browser.CloseAsync();
+
+            Assert.Equal(true, launcher.IsChromeClosed);
```
Meir017: `Assert.True` would be cleaner

lib/PuppeteerSharp/Launcher.cs
Meir017: I added some notes.
you should notice that the close operation is `async` as oppose to `IDisposables.Dispose` which it `void`

kblok: @Meir017 yeah. We'd need to `CloseAsync().GetAwaiter().GetResult()`, do you have a better idea?

kblok:
Great review @Meir017!  I really appreciate it.

lib/PuppeteerSharp/Launcher.cs
Meir017: should good enough for now.
what we need is - https://github.com/dotnet/roslyn/issues/114

Meir017:
The build failed due to access denied when trying to delete the chrome profile directory - https://ci.appveyor.com/project/kblok/puppeteer-sharp/build/1.0.60#L161

lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
```diff
@@ -98,7 +98,7 @@ public class PuppeteerLaunchTests
             Assert.True(Directory.GetFiles(userDataDir).Length > 0);
             await browser.CloseAsync();
             Assert.True(Directory.GetFiles(userDataDir).Length > 0);
-            // Directory.Delete(userDataDir, true);
+            Directory.Delete(userDataDir, true);
```
Meir017: adding
```cs
await Task.Delay(1000);
```
right before this line seems to work, not sure what's not waiting

kblok: yeah... I don't like much that solution. Another solution I read is looping (and failing) until the file is released https://stackoverflow.com/a/37154588/2373249.

I don't care about that delete. The problem is that `Launcher.cs` is also failing to delete the folder here https://github.com/kblok/puppeteer-sharp/blob/b99216161da585f74a1981b466cece9db42ed32b/lib/PuppeteerSharp/Launcher.cs#L250 and we do need to prevent that from happening.

kblok:
Ready to review again @Meir017. I ended up implementing a retry pattern to delete those folders.

lib/PuppeteerSharp/Launcher.cs
```diff
@@ -127,16 +145,66 @@ public class Launcher
...
             catch (Exception ex)
             {
-                await ForceKillChrome();
+                ForceKillChrome();
                 throw new Exception("Failed to create connection", ex);
             }
 
         }
 
+        public static async Task TryDeleteFolder(string folder, int times = 10, TimeSpan? delay = null)
```
Meir017: maybe rename `folder` to `directory`?

kblok: Yeah, much better

kblok: Done @Meir017

lib/PuppeteerSharp/Launcher.cs
```diff
@@ -127,16 +145,66 @@ public class Launcher
...
             catch (Exception ex)
             {
-                await ForceKillChrome();
+                ForceKillChrome();
                 throw new Exception("Failed to create connection", ex);
             }
 
         }
 
+        public static async Task TryDeleteDirectory(string folder, int times = 10, TimeSpan? delay = null)
```
Meir017: can we infer the `folder` directory from the `_options.UserDataDir` or `_temporaryUserDataDir`. I don't think we want this method to be used on any directory. If you remove that argument you could also rename it to something like `TryDeleteUserDataDir`

lib/PuppeteerSharp/Launcher.cs
```diff
@@ -127,16 +145,66 @@ public class Launcher
...
+        {
+            if (times <= 0)
```
Meir017: maybe add a validation that `IsChromeClosed` should be `true`?

lib/PuppeteerSharp/Launcher.cs
kblok: I can buy that. But if you provide a UserDataDir, you should call this method explicitly. We would only remove the folder we create automatically.
Agreed?

lib/PuppeteerSharp/Launcher.cs
kblok: I like it.

lib/PuppeteerSharp/Launcher.cs
```diff
@@ -202,32 +274,61 @@ private void CleanUp()
             _chromeProcess?.RemoveExitedEvent();
         }
 
+        private async Task AfterProcessExit()
+        {
+            if (!IsChromeClosed)
```
Meir017: maybe rewrite as
```cs
if(IsChromeClosed) return;
...
```
to lower the nesting

lib/PuppeteerSharp/Launcher.cs
Meir017: Hmm, didn't think of that. I guess this way is better

kblok: That's why I exposed it as a public method, If you need it you can use it.

lib/PuppeteerSharp/Launcher.cs
kblok: Will do but with braces.

lib/PuppeteerSharp/Launcher.cs
kblok: If we leave this as generic as it is now. Should we check for `IsChromeClosed`?

lib/PuppeteerSharp/Launcher.cs
kblok: Done

lib/PuppeteerSharp/Launcher.cs
Meir017: could it succeed if chrome is open?

Meir017: could it succeed if chrome is open?

kblok: It would fail after retrying 10 times. So we are calling it after closing chrome we wouldn't fail there. If the user calls it, he would get an exception no matter what we do. If we validate IsChromeClosed he'd get an exception and if not he'd get an exception after the 10th retry.

Meir017: If chrome is open there is no point trying to delete since it will fail

kblok: How about now @Meir017?

PR #83, opened 2018-03-08
added Puppeteer.executablePath tests

### LINKED ISSUE #78: Implement Puppeteer.executablePath test
https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/test/test.js#L241

Meir017:
closes #78

### DIFF
--- a/lib/PuppeteerSharp.Tests/Puppeteer/ExecutablePathTests.cs
+++ b/lib/PuppeteerSharp.Tests/Puppeteer/ExecutablePathTests.cs
@@ -3,6 +3,7 @@ using Xunit;
 
 namespace PuppeteerSharp.Tests.Puppeteer
 {
+    [Collection("PuppeteerLoaderFixture collection")]
     public class ExecutablePathTests
     {
         [Fact]
--- a/lib/PuppeteerSharp.Tests/Puppeteer/ExecutablePathTests.cs
+++ b/lib/PuppeteerSharp.Tests/Puppeteer/ExecutablePathTests.cs
@@ -1,0 +1,15 @@
+﻿using System.IO;
+using Xunit;
+
+namespace PuppeteerSharp.Tests.Puppeteer
+{
+    public class ExecutablePathTests
+    {
+        [Fact]
+        public void ShouldWork()
+        {
+            var executablePath = PuppeteerSharp.Puppeteer.GetExecutablePath();
+            Assert.True(File.Exists(executablePath));
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


kblok:
Well not AppVeyor is working :joy:

Meir017:
It's failing because:
image

lib/PuppeteerSharp/Downloader.cs
```diff
@@ -21,6 +21,8 @@ public class Downloader
         };
         private string _downloadHost;
 
+        public const int DefaultRevision = 526987;
```
kblok: I can buy this now. But I created an issue to make it configurable in v1.0 https://github.com/kblok/puppeteer-sharp/issues/84

lib/PuppeteerSharp/Puppeteer.cs
```diff
@@ -1,13 +1,18 @@
...
 {
     public class Puppeteer
     {
         public static string[] DefaultArgs => Launcher.DefaultArgs;
 
+        public static string GetExecutablePath()
+        {
+            var downloader = Downloader.CreateDefault();
+            var revisionInfo = downloader.RevisionInfo(Downloader.CurrentPlatform, Downloader.DefaultRevision);
+            return revisionInfo.ExecutablePath;
```
kblok: This code is in the Launcher class on Puppeteer. Let's honor that.

https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/lib/Launcher.js#L188

Meir017: should that method be static?

kblok: Yes. Let's follow Puppetter API, it's static there.

lib/PuppeteerSharp/Launcher.cs
```diff
@@ -136,6 +136,12 @@ public class Launcher
             }
 
         }
+        internal string GetExecutablePath()
```
kblok: Can we add one line between the bracket on line 138 and this new function?

lib/PuppeteerSharp.Tests/Puppeteer/ExecutablePathTests.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System.IO;
+using Xunit;
+
+namespace PuppeteerSharp.Tests.Puppeteer
+{
+    public class ExecutablePathTests
```
kblok: You need to make this class inherit from `PuppeteerBaseTest` and add the attribute `[Collection("PuppeteerLoaderFixture collection")]`

The problem we have here is that, unluckily, this is the first test running, so we need to download Chromium first, and `PuppeteerBaseTest` does that job for us.

kblok: @Meir017 do you think you will be able to implement this by tomorrow? 
If not I can merge it and implement it in another PR.

PR #86, opened 2018-03-10
Implement IDisposable on Browser and Page

### LINKED ISSUE #85: Implement IDisposable on Browser and Page

kblok:
closes #85

### DIFF
--- a/lib/PuppeteerSharp/Page.cs
+++ b/lib/PuppeteerSharp/Page.cs
@@ -446,7 +446,7 @@ namespace PuppeteerSharp
 
         public async Task CloseAsync()
         {
-            if (_client != null && !_client.Connection.IsClosed)
+            if (!(_client?.Connection?.IsClosed ?? true))
             {
                 await _client.Connection.SendAsync("Target.closeTarget", new
                 {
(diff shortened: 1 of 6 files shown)


kblok:
Ready to review @Meir017.

We this guy and #83 we release v0.3

lib/PuppeteerSharp/Page.cs
```diff
@@ -446,10 +446,13 @@ private Page(Session client, Target target, FrameTree frameTree, bool ignoreHTTP
 
         public async Task CloseAsync()
         {
-            await _client?.Connection?.SendAsync("Target.closeTarget", new
+            if (_client != null && !_client.Connection.IsClosed)
```
Meir017: could `_client.Connection` be `null`?

kblok: It shouldn't be we could double check.

kblok: Would this be clear to you?
`if (!(_client?.Connection?.IsClosed ?? true))`

Meir017: I was thinking of suggesting something like that

kblok: Awesome!

kblok: How about now?

PR #94, opened 2018-03-14
Implement Page.evaluate Feature

### LINKED ISSUE #89: Implement Page.evaluate tests
https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/test/test.js#L342

kblok:
closes #89

### DIFF
--- a/lib/PuppeteerSharp.Tests/Page/EvaluateTests.cs
+++ b/lib/PuppeteerSharp.Tests/Page/EvaluateTests.cs
@@ -79,7 +79,7 @@ namespace PuppeteerSharp.Tests.Page
             using (var page = await Browser.NewPageAsync())
             {
                 dynamic result = await page.EvaluateAsync<dynamic>("() => NaN");
-                throw new NotImplementedException();
+                Assert.Equal(double.NaN, result);
             }
         }
 
@@ -89,7 +89,7 @@ namespace PuppeteerSharp.Tests.Page
             using (var page = await Browser.NewPageAsync())
             {
                 dynamic result = await page.EvaluateAsync<dynamic>("() => -0");
-                throw new NotImplementedException();
+                Assert.Equal(-0, result);
             }
         }
 
@@ -99,7 +99,7 @@ namespace PuppeteerSharp.Tests.Page
             using (var page = await Browser.NewPageAsync())
             {
                 dynamic result = await page.EvaluateAsync<dynamic>("() => Infinity");
-                throw new NotImplementedException();
+                Assert.Equal(double.PositiveInfinity, result);
             }
         }
 
@@ -109,7 +109,7 @@ namespace PuppeteerSharp.Tests.Page
             using (var page = await Browser.NewPageAsync())
             {
                 dynamic result = await page.EvaluateAsync<dynamic>("() => -Infinity");
-                throw new NotImplementedException();
+                Assert.Equal(double.NegativeInfinity, result);
             }
         }
 
@@ -118,7 +118,7 @@ namespace PuppeteerSharp.Tests.Page
         {
             using (var page = await Browser.NewPageAsync())
             {
-                dynamic result = await page.EvaluateAsync<bool>("a, b => Object.is(a, null) && Object.is(b, 'foo')", null, "foo");
+                dynamic result = await page.EvaluateAsync<bool>("(a, b) => Object.is(a, null) && Object.is(b, 'foo')", null, "foo");
                 Assert.True(result);
             }
         }
--- a/lib/PuppeteerSharp.Tests/Page/EvaluateTests.cs
+++ b/lib/PuppeteerSharp.Tests/Page/EvaluateTests.cs
@@ -121,6 +121,5 @@ namespace PuppeteerSharp.Tests.Page
                 Assert.Null(result["a"]);
             }
         }
-
     }
 }
--- a/lib/PuppeteerSharp.Tests/Page/EvaluateTests.cs
+++ b/lib/PuppeteerSharp.Tests/Page/EvaluateTests.cs
@@ -59,8 +59,7 @@ namespace PuppeteerSharp.Tests.Page
                     return page.EvaluateFunctionAsync<object>("() => not.existing.object.property");
                 });
 
-                Assert.NotNull(exception);
-                Assert.True(exception.Message.Contains("not is not defined"));
+                Assert.Contains("not is not defined", exception.Message);
             }
         }
 
@@ -97,7 +96,7 @@ namespace PuppeteerSharp.Tests.Page
         {
             using (var page = await Browser.NewPageAsync())
             {
-                dynamic result = await page.EvaluateFunctionAsync<bool>("(a, b) => Object.is(a, null) && Object.is(b, 'foo')", null, "foo");
+                bool result = await page.EvaluateFunctionAsync<bool>("(a, b) => Object.is(a, null) && Object.is(b, 'foo')", null, "foo");
                 Assert.True(result);
             }
         }
@@ -108,7 +107,7 @@ namespace PuppeteerSharp.Tests.Page
             using (var page = await Browser.NewPageAsync())
             {
                 var result = await page.EvaluateFunctionAsync<Dictionary<string, object>>("() => ({a: undefined})");
-                Assert.False(result.ContainsKey("a"));
+                Assert.Empty(result);
             }
         }
 
--- a/lib/PuppeteerSharp.Tests/Page/EvaluateTests.cs
+++ b/lib/PuppeteerSharp.Tests/Page/EvaluateTests.cs
@@ -68,8 +68,8 @@ namespace PuppeteerSharp.Tests.Page
                 {
                     foo = "bar!"
                 };
-                dynamic result = await page.EvaluateAsync<dynamic>("a => a", obj);
-                Assert.Equal("bar!", result.foo);
+                dynamic result = await page.EvaluateAsync("a => a", obj);
+                Assert.Equal("bar!", result.foo.ToString());
             }
         }
 
--- a/lib/PuppeteerSharp.Tests/Page/EvaluateTests.cs
+++ b/lib/PuppeteerSharp.Tests/Page/EvaluateTests.cs
@@ -9,18 +9,29 @@ namespace PuppeteerSharp.Tests.Page
     public class EvaluateTests : PuppeteerBaseTest
     {
         [Theory]
-        [InlineData("() => 7 * 3", 21)] //ShouldWork
-        [InlineData("() => Promise.resolve(8 * 7)", 56)] //ShouldAwaitPromise
         [InlineData("1 + 5;", 6)] //ShouldAcceptSemiColons
         [InlineData("2 + 5\n// do some math!'", 7)] //ShouldAceptStringComments
-        public async Task BasicIntEvaluationTest(string script, object expected)
+        public async Task BasicIntExressionEvaluationTest(string script, object expected)
         {
             using (var page = await Browser.NewPageAsync())
             {
-                var result = await page.EvaluateAsync<int>(script);
+                var result = await page.EvaluateExpressionAsync<int>(script);
                 Assert.Equal(expected, result);
             }
         }
+
+        [Theory]
+        [InlineData("() => 7 * 3", 21)] //ShouldWork
+        [InlineData("() => Promise.resolve(8 * 7)", 56)] //ShouldAwaitPromise
+        public async Task BasicIntFunctionEvaluationTest(string script, object expected)
+        {
+            using (var page = await Browser.NewPageAsync())
+            {
+                var result = await page.EvaluateFunctionAsync<int>(script);
+                Assert.Equal(expected, result);
+            }
+        }
+
 
         [Fact]
         public async Task ShouldWorkRightAfterFrameNavigated()
@@ -31,7 +42,7 @@ namespace PuppeteerSharp.Tests.Page
             {
                 page.FrameNavigated += (sender, e) =>
                 {
-                    frameEvaluation = e.Frame.EvaluateAsync<int>("() => 6 * 7");
+                    frameEvaluation = e.Frame.EvaluateFunctionAsync<int>("() => 6 * 7");
                 };
 
                 await page.GoToAsync(TestConstants.EmptyPage);
@@ -46,7 +57,7 @@ namespace PuppeteerSharp.Tests.Page
             {
                 var exception = await Assert.ThrowsAsync<EvaluationFailedException>(() =>
                 {
-                    return page.EvaluateAsync<object>("() => not.existing.object.property");
+                    return page.EvaluateFunctionAsync<object>("() => not.existing.object.property");
                 });
 
                 Assert.NotNull(exception);
@@ -63,7 +74,7 @@ namespace PuppeteerSharp.Tests.Page
                 {
                     foo = "bar!"
                 };
-                dynamic result = await page.EvaluateAsync("a => a", obj);
+                dynamic result = await page.EvaluateFunctionAsync("a => a", obj);
                 Assert.Equal("bar!", result.foo.ToString());
             }
         }
@@ -77,7 +88,7 @@ namespace PuppeteerSharp.Tests.Page
         {
             using (var page = await Browser.NewPageAsync())
             {
-                dynamic result = await page.EvaluateAsync(script);
+                dynamic result = await page.EvaluateFunctionAsync(script);
                 Assert.Equal(expected, result);
             }
         }
@@ -87,7 +98,7 @@ namespace PuppeteerSharp.Tests.Page
         {
             using (var page = await Browser.NewPageAsync())
             {
-                dynamic result = await page.EvaluateAsync<bool>("(a, b) => Object.is(a, null) && Object.is(b, 'foo')", null, "foo");
+                dynamic result = await page.EvaluateFunctionAsync<bool>("(a, b) => Object.is(a, null) && Object.is(b, 'foo')", null, "foo");
                 Assert.True(result);
             }
         }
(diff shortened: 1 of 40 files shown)


Meir017:
closes #89

kblok:
All tests are green now @Meir017. I believe this is an important part of the project,  many features to come depends on this guy, so let's give it a good review. I'll give it a self-review tomorrow.

lib/PuppeteerSharp.Tests/Page/EvaluateTests.cs
```diff
@@ -0,0 +1,166 @@
...
+{
+    [Collection("PuppeteerLoaderFixture collection")]
+    public class EvaluateTests : PuppeteerBaseTest
+    {
+        [Fact]
+        public async Task ShouldWork()
+        {
+            using (var page = await Browser.NewPageAsync())
+            {
+                var result = await page.EvaluateAsync<int>("() => 7 * 3");
```
Meir017: I'm guessing using `Fact` instead of `Theory` is to stay true to the original? the only difference between the tests in the input of the method `EvaluateAsync` and the expected value

lib/PuppeteerSharp.Tests/Page/EvaluateTests.cs
```diff
@@ -0,0 +1,166 @@
...
+                Assert.True(result);
+            }
+        }
+
+        [Fact]
+        public async Task ShouldProperlySerializeNullFields()
+        {
+            using (var page = await Browser.NewPageAsync())
+            {
+                dynamic result = await page.EvaluateAsync<dynamic>("() => ({a: undefined})");
```
Meir017: shouldn't this serialize to a dictionary and make sure the key `a` exists with a `null` value?

lib/PuppeteerSharp.Tests/Page/EvaluateTests.cs
```diff
@@ -0,0 +1,166 @@
...
+                Assert.Null(result.a);
+            }
+        }
+
+        [Fact]
+        public async Task ShouldAcceptAString()
```
Meir017: not sure this is relevant to our project since argument is always a string

lib/PuppeteerSharp/EvaluateExceptionStackTrace.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System.Collections.Generic;
+
+namespace PuppeteerSharp
+{
+    public class EvaluateExceptionStackTrace
+    {
+        public IEnumerable<EvaluationExceptionCallFrame> CallFrames { get; internal set; }
```
Meir017: is this missing a `JsonProperty` attribute?

this is only initialized from DeSerialization so it might as well be an Array, not sure about this

lib/PuppeteerSharp/ExecutionContext.cs
```diff
@@ -22,44 +24,86 @@ public ExecutionContext(Session client, ContextPayload contextPayload, Func<dyna
...
+        internal async Task<JSHandle> EvaluateHandleAsync(string script, object[] args)
         {
-            if (!string.IsNullOrEmpty(pageFunction))
+            if (string.IsNullOrEmpty(script))
             {
-                dynamic remoteObject;
+                return null;
+            }
+
+            if (IsFunction(script))
```
Meir017: maybe expose `EvaluateFunctionAsync` & `EvaluateExpressionAsync` instead of trying to implement `IsFunction`?

lib/PuppeteerSharp/ExecutionContext.cs
```diff
@@ -22,44 +24,86 @@ public ExecutionContext(Session client, ContextPayload contextPayload, Func<dyna
...
-                        ["expression"] = pageFunction,
-                        ["contextId"] = _contextId,
-                        ["returnByValue"] = false,
-                        ["awaitPromise"] = true
-                    });
-
-                    return ObjectHandleFactory(remoteObject.result);
-                }
-                catch (Exception ex)
+                    {"contextId", _contextId},
```
Meir017: I see you don't like the C# 6 dictionary initializer syntax ;)

lib/PuppeteerSharp/ExecutionContext.cs
```diff
@@ -22,44 +24,86 @@ public ExecutionContext(Session client, ContextPayload contextPayload, Func<dyna
         public string FrameId { get; internal set; }
         public bool IsDefault { get; internal set; }
 
-        public async Task<dynamic> Evaluate(string pageFunction, params object[] args)
+        public async Task<dynamic> EvaluateAsync(string script, params object[] args)
         {
-            var handle = await EvaluateHandleAsync(pageFunction, args);
+            var handle = await EvaluateHandleAsync(script, args);
             dynamic result = await handle.JsonValue();
```
Meir017: is there a case when `result` isn't a `JToken`? we might not need this overload

lib/PuppeteerSharp/ExecutionContext.cs
```diff
@@ -22,44 +24,86 @@ public ExecutionContext(Session client, ContextPayload contextPayload, Func<dyna
...
+                    Helper.GetExceptionMessage(result.exceptionDetails.ToObject<EvaluateExceptionDetails>()));
             }
 
-            return null;
+            return ObjectHandleFactory(result.result);
+        }
+
+        private object FormatArguments(object[] args)
+        {
+            return args.Select(o => new { value = o });
```
Meir017: this is not the complete implementation of the [`convertArgument`](https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/lib/ExecutionContext.js#L78) method

lib/PuppeteerSharp/Frame.cs
```diff
@@ -99,8 +106,8 @@ internal Task<ElementHandle> AddScriptTag(dynamic options)
 
         internal async Task<string> GetTitleAsync()
         {
-            var result = await EvaluateAsync("document.title");
-            return result.ToString();
+            var result = await EvaluateAsync<string>("document.title");
```
Meir017: this can be written as an arrow function:
```cs
GetTitleAsync() => EvaluateAsync<string>("document.title");
```

lib/PuppeteerSharp/Page.cs
```diff
@@ -456,6 +451,16 @@ private Page(Session client, Target target, FrameTree frameTree, bool ignoreHTTP
             }
         }
 
+        public Task<dynamic> EvaluateAsync(string script, params object[] args)
+        {
+            return _frameManager.MainFrame.EvaluateAsync(script, args);
```
Meir017: convert to an arrow method?

lib/PuppeteerSharp/Page.cs
```diff
@@ -456,6 +451,16 @@ private Page(Session client, Target target, FrameTree frameTree, bool ignoreHTTP
...
+        }
+
+        public Task<T> EvaluateAsync<T>(string script, params object[] args)
+        {
+            return _frameManager.MainFrame.EvaluateAsync<T>(script, args);
```
Meir017: convert to an arrow method?

lib/PuppeteerSharp.Tests/Page/EvaluateTests.cs
kblok: Good idea. I'll try to implement that. I'm a Nunit user so I'm trying to get used to XUnit. I'd like to see if it supports types as parameters for the generics.

lib/PuppeteerSharp.Tests/Page/EvaluateTests.cs
kblok: Agreed

lib/PuppeteerSharp/ExecutionContext.cs
kblok: Love it. By doing that we'll expose to APIs as Puppeteer does.

lib/PuppeteerSharp/ExecutionContext.cs
kblok: The problem is that we can't use `dynamic` when we call `EvaluateAsync<T>`. So when we need a `dynamic` we should call this function.

Thoughts?

lib/PuppeteerSharp/ExecutionContext.cs
kblok: I would wait for test needing that code. What do you think?

lib/PuppeteerSharp/ExecutionContext.cs
Meir017: when would want our response to be of type `dynamic`? we always know what we expect the method to return

lib/PuppeteerSharp/ExecutionContext.cs
Meir017: ok, I wasn't sure if this method has tests

lib/PuppeteerSharp.Tests/Page/EvaluateTests.cs
kblok: I was reviewing this and on Puppeteer's test is returning an empty object https://github.com/GoogleChrome/puppeteer/blob/v1.1.0/test/test.js#L465

I think the name is wrong (even in Puppeteer) this should be called `ShouldProperlyIgnoreUndefinedFields`. I'm creating both tests so we can express the intent here.

lib/PuppeteerSharp/ExecutionContext.cs
kblok: I thought about this and I think we should leave only one entry point here. It has no benefit for the user having 2. She can send us a script and we know what to do there.

lib/PuppeteerSharp/ExecutionContext.cs
kblok: hahah I'm not used to that. On the other hand and don't have to have some files with one style and other files with another style. But yeah, we can refactor that in the future.

lib/PuppeteerSharp/ExecutionContext.cs
kblok: Here, for instance, https://github.com/kblok/puppeteer-sharp/pull/94/files#diff-a700c62632ba5e31465dc34bc7e2d89dR72

I wouldn't like to have to create a class to access `result.foo`

lib/PuppeteerSharp/Page.cs
kblok: If I use an arrow the line gets too long. Instead of breaking into lines I'd rather use braces, what do you think?

lib/PuppeteerSharp/Page.cs
kblok: Same here

kblok:
Some feedback implemented here @Meir017

lib/PuppeteerSharp.Tests/Page/EvaluateTests.cs
Meir017: so i guess we should assert that the object returned is an empty object without properties? this way it looks like it has a null property called `a`.
I would go with something like `EvaluateAsync<Dictionary<string, object>>` and make sure it equals `new Dictionary<string, object>()`

kblok: Yes, check the updated code ;)

lib/PuppeteerSharp/ExecutionContext.cs
Meir017: my issue with this is that a function can be written in using different syntaxes:
```js
// 1.
function() { return 1 + 2; }
// 2.
() => 1 + 2;
// 3.
arg => 1 + 2;
// 4.
(arg1, arg2) => 1 + 2;
// not a function but contains =>
document.querySelectorAll('div').forEach(e=> console.log(e.innerText))
```

kblok: Good catch!

lib/PuppeteerSharp/ExecutionContext.cs
Meir017: well, that's because you sent a dynamic object in the first place and returned it... anyhow, it doesn't hurt

lib/PuppeteerSharp/Page.cs
Meir017: you could have the arrow on the next line, this way the method is a total of 2 lines instead of 4 lines

kblok:
Ready to review again @Meir017

lib/PuppeteerSharp.Tests/Page/EvaluateTests.cs
```diff
@@ -0,0 +1,127 @@
...
+        public async Task ShouldRejectPromiseWithExeption()
+        {
+            using (var page = await Browser.NewPageAsync())
+            {
+                var exception = await Assert.ThrowsAsync<EvaluationFailedException>(() =>
+                {
+                    return page.EvaluateFunctionAsync<object>("() => not.existing.object.property");
+                });
+
+                Assert.NotNull(exception);
```
Meir017: pretty sure `exception` cannot be null

lib/PuppeteerSharp.Tests/Page/EvaluateTests.cs
```diff
@@ -0,0 +1,127 @@
...
+                Assert.Equal(expected, result);
+            }
+        }
+
+        [Fact]
+        public async Task ShouldAcceptNullAsOneOfMultipleParameters()
+        {
+            using (var page = await Browser.NewPageAsync())
+            {
+                dynamic result = await page.EvaluateFunctionAsync<bool>("(a, b) => Object.is(a, null) && Object.is(b, 'foo')", null, "foo");
```
Meir017: `result` should be `bool`. or simply use `var`.

lib/PuppeteerSharp/Helper.cs
```diff
@@ -37,5 +37,24 @@ internal static object ValueFromRemoteObject(dynamic remoteObject)
                 Console.WriteLine(ex.ToString());
             }
         }
+
+        public static string GetExceptionMessage(EvaluateExceptionDetails exceptionDetails)
```
Meir017: change to `internal`?

lib/PuppeteerSharp/Page.cs
```diff
@@ -456,6 +457,18 @@ private Page(Session client, Target target, FrameTree frameTree, bool ignoreHTTP
             }
         }
 
+        public Task<dynamic> EvaluateExpressionAsync(string script) =>
```
Meir017: move the arrow to next line - https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Core/UrlHelperExtensions.cs#L366

lib/PuppeteerSharp.Tests/Page/EvaluateTests.cs
kblok: Why?

lib/PuppeteerSharp/Page.cs
kblok: Cool, I was looking for examples :+1:

lib/PuppeteerSharp.Tests/Page/EvaluateTests.cs
Meir017: the previous line will throw if the returned type is not an `EvaluationFailedException`

kblok: Gotcha!

lib/PuppeteerSharp.Tests/Page/EvaluateTests.cs
```diff
@@ -0,0 +1,127 @@
...
+        {
+            using (var page = await Browser.NewPageAsync())
+            {
+                var exception = await Assert.ThrowsAsync<EvaluationFailedException>(() =>
+                {
+                    return page.EvaluateFunctionAsync<object>("() => not.existing.object.property");
+                });
+
+                Assert.NotNull(exception);
+                Assert.True(exception.Message.Contains("not is not defined"));
```
Meir017: `Assert.Contains` would be better

lib/PuppeteerSharp.Tests/Page/EvaluateTests.cs
```diff
@@ -0,0 +1,127 @@
...
+            }
+        }
+
+        [Fact]
+        public async Task ShouldProperlyIgnoreUndefinedFields()
+        {
+            using (var page = await Browser.NewPageAsync())
+            {
+                var result = await page.EvaluateFunctionAsync<Dictionary<string, object>>("() => ({a: undefined})");
+                Assert.False(result.ContainsKey("a"));
```
Meir017: maybe `Assert.Empty(result)`?

lib/PuppeteerSharp.Tests/Page/EvaluateTests.cs
```diff
@@ -0,0 +1,127 @@
...
+            }
+        }
+
+        [Fact]
+        public async Task ShouldProperlySerializeNullFields()
+        {
+            using (var page = await Browser.NewPageAsync())
+            {
+                var result = await page.EvaluateFunctionAsync<Dictionary<string, object>>("() => ({a: null})");
+                Assert.True(result.ContainsKey("a"));
```
Meir017: this is good enough for now, maybe once https://github.com/xunit/xunit/issues/1620 is done we could improve this

lib/PuppeteerSharp/EvaluateExceptionDetails.cs
```diff
@@ -0,0 +1,14 @@
+﻿using Newtonsoft.Json;
+
+namespace PuppeteerSharp
+{
+    public class EvaluateExceptionDetails
+    {
+        [JsonProperty("exception")]
+        public EvaluateExceptionInfo Exception { get; set; }
```
Meir017: should the set be `internal`?

lib/PuppeteerSharp/EvaluateExceptionInfo.cs
```diff
@@ -0,0 +1,10 @@
+﻿using Newtonsoft.Json;
+
+namespace PuppeteerSharp
+{
+    public class EvaluateExceptionInfo
+    {
+        [JsonProperty("description")]
+        public string Description { get; set; }
```
Meir017: should the set be `internal`?

kblok:
Done @Meir017, keep pushing!

PR #96, opened 2018-03-15
added Page.title tests

### LINKED ISSUE #49: Implement Page.title tests
https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/test/test.js#L2945

Meir017:
closes #49


lib/PuppeteerSharp/ExecutionContext.cs
```diff
@@ -7,16 +7,18 @@ namespace PuppeteerSharp
...
         {
             _client = client;
             _contextId = contextPayload.Id;
+            FrameId = contextPayload.AuxData.FrameId;
+            IsDefault = contextPayload.AuxData.IsDefault;
             ObjectHandleFactory = objectHandleFactory;
         }
 
-        public Func<JSHandle, dynamic> ObjectHandleFactory { get; internal set; }
+        public Func<dynamic, JSHandle> ObjectHandleFactory { get; internal set; }
```
Meir017: took me a while to realize this...

lib/PuppeteerSharp/Frame.cs
```diff
@@ -92,6 +98,12 @@ internal Task<ElementHandle> AddScriptTag(dynamic options)
             throw new NotImplementedException();
         }
 
+        internal async Task<string> GetTitleAsync()
+        {
+            var result = await EvaluateAsync("document.title");
```
Meir017: using `() => document.title` didn't work since we are missing the logic - https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/lib/ExecutionContext.js#L53

kblok:
Could you delete and ignore Screenshots/test.png?
I thought I ignored that.

lib/PuppeteerSharp/ContextPayload.cs
```diff
@@ -6,8 +6,17 @@ public class ContextPayload
...
+            AuxData = context.auxData.ToObject<AuxData>();
         }
 
-        public string Id { get; internal set; }
+        public int Id { get; internal set; }
+
+        public AuxData AuxData { get; internal set; }
+    }
+
+    public class AuxData
```
kblok: 1. This could be an `struct`
2. I would give it another name, something like `ContextPayloadData`. and move it to its own file.

lib/PuppeteerSharp/Frame.cs
kblok: Evaluate is working? 😱

Meir017: just for expressions for now, for functions we need a bit more work

Meir017:
any more changes needed?

lib/PuppeteerSharp/ContextPayloadAuxData.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace PuppeteerSharp
+{
+
```
kblok: Let's remove this extra line

PR #101, opened 2018-03-21
use server in process instead of external process

Meir017:
improves #80

### DIFF
--- a/lib/PuppeteerSharp.TestServer/Program.cs
+++ b/lib/PuppeteerSharp.TestServer/Program.cs
@@ -1,29 +1,21 @@
-﻿using System;
-using System.IO;
-using System.Net;
+﻿using System.Net;
 using Microsoft.AspNetCore;
 using Microsoft.AspNetCore.Hosting;
 
 namespace PuppeteerSharp.TestServer
 {
-    class Program
+    public class Program
     {
-        static void Main(string[] args)
-        {
-            var host = WebHost.CreateDefaultBuilder(args)
-                .UseKestrel()
-                .UseStartup<Startup>()
-                .UseKestrel(options =>
+        public static IWebHostBuilder GetWebHostBuilder(params string[] args) => WebHost.CreateDefaultBuilder(args)
+            .UseKestrel()
+            .UseStartup<Startup>()
+            .UseKestrel(options =>
+            {
+                options.Listen(IPAddress.Loopback, 8907);
+                options.Listen(IPAddress.Loopback, 8908, listenOptions =>
                 {
-                    options.Listen(IPAddress.Loopback, 8907);
-                    options.Listen(IPAddress.Loopback, 8908, listenOptions =>
-                    {
-                        listenOptions.UseHttps("testCert.cer");
-                    });
-                })
-                .Build();
-
-            host.Run();
-        }
+                    listenOptions.UseHttps("testCert.cer");
+                });
+            });
     }
 }
--- a/lib/PuppeteerSharp.Tests/PuppeteerLoaderFixture.cs
+++ b/lib/PuppeteerSharp.Tests/PuppeteerLoaderFixture.cs
@@ -1,14 +1,13 @@
-﻿using System;
-using System.Diagnostics;
+﻿using Microsoft.AspNetCore.Hosting;
+using System;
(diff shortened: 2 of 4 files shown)


Meir017:
`System.DllNotFoundException : Unable to load DLL 'libuv': The specified module could not be found. (Exception from HRESULT: 0x8007007E)` sigh :(

kblok:
@Meir017 it's working locally but not on AppVeyor?

Meir017:
it's working for netcoreapp2.0 but not for net471

kblok:
ouch, we need to support net471 here.

Meir017:
it's a known issue - https://github.com/aspnet/KestrelHttpServer/issues/1652
I'm thinking of waiting for 2.1 which uses sockets instead of libuv by default

kblok:
ouch

Meir017:
is there a way to see what appveyor is executing? (`appveyor.yml`) it's not in the repo :(

lib/PuppeteerSharp.Tests/PuppeteerLoaderFixture.cs
```diff
@@ -45,64 +25,27 @@ public void Dispose()
...
-                taskWrapper.SetException(new Exception("Unable to start web server"));
-            };
+            await _host.StartAsync();
+        }
 
-            Timer timer = null;
-            //We have to declare timer before initializing it because if we don't do this 
-            //we can't dispose it in the action created in the constructor
-            timer = new Timer((state) =>
+        public static string FindParentFirectory(string directory)
```
Meir017: not sure where this method should be, maybe a static class `DirectoryHelper` or something like that?

Meir017:
* net471 test run was slightly faster ([`1.0426 Minutes`](https://ci.appveyor.com/project/kblok/puppeteer-sharp/build/1.0.207#L524)) than previous ([`1.1793 Minutes`](https://ci.appveyor.com/project/kblok/puppeteer-sharp/build/1.0.196#2723))
* netcoreapp2.0 test run was slightly slower ([`1.3293 Minutes`](https://ci.appveyor.com/project/kblok/puppeteer-sharp/build/1.0.207#369)) than previous ([`1.2213 Minutes`](https://ci.appveyor.com/project/kblok/puppeteer-sharp/build/1.0.196#1416))

lib/PuppeteerSharp.Tests/PuppeteerLoaderFixture.cs
kblok: I'm good having this guy here. Though I don't know what a Firectory is :joy:

kblok:
Is this working??? 😮

Meir017:
yup

lib/PuppeteerSharp.Tests/PuppeteerLoaderFixture.cs
```diff
@@ -45,64 +25,27 @@ public void Dispose()
...
-                taskWrapper.SetException(new Exception("Unable to start web server"));
-            };
+            await _host.StartAsync();
+        }
 
-            Timer timer = null;
-            //We have to declare timer before initializing it because if we don't do this 
-            //we can't dispose it in the action created in the constructor
-            timer = new Timer((state) =>
+        public static string FindParentDirectory(string directory)
```
kblok: My comment got lost due to the changes. I'd create a TestUtils.cs class with this. We might need it for other generic stuff.

lib/PuppeteerSharp.TestServer/Program.cs
```diff
@@ -1,29 +1,19 @@
...
-    class Program
+    public class Program
     {
-        static void Main(string[] args)
-        {
-            var host = WebHost.CreateDefaultBuilder(args)
-                .UseKestrel()
-                .UseStartup<Startup>()
-                .UseKestrel(options =>
+        public static IWebHostBuilder GetWebHostBuilder(params string[] args) => new WebHostBuilder()
```
kblok: I think we don't need a "Program" class anymore. Or we rename it to WebServer.cs or we can move it to the Startup class.

kblok:
After the program rename, this will be ready to good @Meir017

PR #102, opened 2018-03-22
Frame/FrameManager and Page redirect fixes

### LINKED ISSUE #100: Not all links produce an export
As stated yesterday not all links produce an export when this happens i get put into break mode and gives me an error that states the folowing:
* An unhandled exception of type 'System.NotImplementedException' occurred in mscorlib.dll
The method or operation is not implemented.
+ i have noticed for each link a new chromium instance gets opened but never closed is there a way to do this with code?

foreach (var list_item in sql_list)
            {
                string id = list_item.Item2.ToString();
                string extension = ".pdf";
                string exportname = id + extension;
                var chromiumRevision = 526987;

                Console.WriteLine("Downloading chromium");
                await Downloader.CreateDefault().DownloadRevisionAsync(chromiumRevision);

                Console.WriteLine("Navigating " + list_item.Item1);
                {
                    Headless = true
                }, chromiumRevision);

                await page.GoToAsync(list_item.Item1);

                Console.WriteLine("Generating PNG");
                await page.PdfAsync(Path.Combine(directory,exportname));

                Console.WriteLine("Export " + exportname + " completed");
                Console.WriteLine();

            }

kblok:
Several bug fixes on Frame and FrameManager
Fixed duplicated requests on Page

closes #100

### DIFF
--- a/lib/PuppeteerSharp.Tests/Issues/PdfIssuesTests.cs
+++ b/lib/PuppeteerSharp.Tests/Issues/PdfIssuesTests.cs
@@ -1,0 +1,32 @@
+﻿using System;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace PuppeteerSharp.Tests.Issues
+{
+    public class PdfIssuesTests : PuppeteerBaseTest
+    {
+        public PdfIssuesTests()
+        {
+            Downloader.CreateDefault().DownloadRevisionAsync(TestConstants.ChromiumRevision).GetAwaiter().GetResult();
+        }
+
+        [Fact]
+        public async Task Issue100()
+        {
+            using (var browser = PuppeteerSharp.Puppeteer.LaunchAsync(TestConstants.DefaultBrowserOptions(),
+                                                                     TestConstants.ChromiumRevision))
+            using (var page = await Browser.NewPageAsync())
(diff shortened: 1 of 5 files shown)


lib/PuppeteerSharp.Tests/Issues/PdfIssuesTests.cs
```diff
@@ -0,0 +1,32 @@
...
+{
+    public class PdfIssuesTests : PuppeteerBaseTest
+    {
+        public PdfIssuesTests()
+        {
+            Downloader.CreateDefault().DownloadRevisionAsync(TestConstants.ChromiumRevision).GetAwaiter().GetResult();
+        }
+
+        [Fact]
+        public async Task Issue100()
```
kblok: What do you think about this @Meir017?
Should we add every issue we find? Should those be enabled by default?

Meir017: maybe each file should cover a single issue - like this https://github.com/JamesNK/Newtonsoft.Json/tree/master/Src/Newtonsoft.Json.Tests/Issues ?

kblok:
Ready to review @Meir017

PR #105, opened 2018-03-24
implemented Page.setJavaScriptEnabled

### LINKED ISSUE #46: Implement Page.setJavascriptEnabled tests
https://github.com/GoogleChrome/puppeteer/blob/master/test/test.js#L2962

Meir017:
closes #46

### DIFF
--- a/lib/PuppeteerSharp.Tests/Page/SetJavaScriptEnabledTests.cs
+++ b/lib/PuppeteerSharp.Tests/Page/SetJavaScriptEnabledTests.cs
@@ -1,0 +1,25 @@
+﻿using System;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace PuppeteerSharp.Tests.Page
+{
+    [Collection("PuppeteerLoaderFixture collection")]
+    public class SetJavaScriptEnabledTests : PuppeteerBaseTest
+    {
+        [Fact]
+        public async Task ShouldWork()
+        {
+            var page = await Browser.NewPageAsync();
+            await page.SetJavaScriptEnabledAsync(false);
+            await page.GoToAsync("data:text/html, <script>var something = 'forbidden'</script>");
+
+            var exception = await Assert.ThrowsAnyAsync<Exception>(async () => await page.EvaluateFunctionAsync("something"));
+            Assert.Contains("something is not defined", exception.Message);
+
+            await page.SetJavaScriptEnabledAsync(true);
+            await page.GoToAsync("data:text/html, <script>var something = 'forbidden'</script>");
+            Assert.Equal("forbidden", await page.EvaluateExpressionAsync<string>("something"));
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


lib/PuppeteerSharp.Tests/Page/SetJavaScriptEnabledTests.cs
```diff
@@ -0,0 +1,25 @@
...
+    public class SetJavaScriptEnabledTests : PuppeteerBaseTest
+    {
+        [Fact]
+        public async Task ShouldWork()
+        {
+            var page = await Browser.NewPageAsync();
+            await page.SetJavaScriptEnabledAsync(false);
+            await page.GoToAsync("data:text/html, <script>var something = 'forbidden'</script>");
+
+            var exception = await Assert.ThrowsAnyAsync<Exception>(async () => await page.EvaluateFunctionAsync("something"));
```
Meir017: not sure why I had to execute `EvaluateFunctionAsync` here but it only worked this way (line22 uses `EvaluateExpressionAsync`)

kblok: I think we need to test `EvaluateExpressionAsync` here. If `EvaluateFunctionAsync` failed it's maybe because that's not a valid function call. But we definitely have to try `EvaluateExpressionAsync` here as Puppeteer does.

lib/PuppeteerSharp/ExecutionContext.cs
```diff
@@ -59,24 +59,13 @@ public ExecutionContext(Session client, ContextPayload contextPayload, Func<dyna
...
-                    {"contextId", _contextId},
-                    {"expression", script},
-                    {"returnByValue", false},
-                    {"awaitPromise", true}
-                });
-
-                return ObjectHandleFactory(remoteObject.result);
-            }
-            catch (Exception ex)
+            return await EvaluateHandleAsync("Runtime.evaluate", new Dictionary<string, object>()
```
kblok: So we're letting the EvaluateHandleAsync fail so we get a more accurate exception?

Meir017: it's the same behavior as the original:
* https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/lib/ExecutionContext.js#L57
* https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/lib/ExecutionContext.js#L69

lib/PuppeteerSharp.Tests/Page/SetJavaScriptEnabledTests.cs
```diff
@@ -0,0 +1,25 @@
...
+
+namespace PuppeteerSharp.Tests.Page
+{
+    [Collection("PuppeteerLoaderFixture collection")]
+    public class SetJavaScriptEnabledTests : PuppeteerBaseTest
+    {
+        [Fact]
+        public async Task ShouldWork()
+        {
+            var page = await Browser.NewPageAsync();
```
kblok: Can we use `using(var Page =...`?

Meir017: we don't really need to since `Browser.Dispose()` is called at the end

PR #107, opened 2018-03-24
implemented Page.cookies tests

### LINKED ISSUE #53: Implement Cookies tests
https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/test/test.js#L3148

Some tests depend on `Evaluate`

Meir017:
closes #53

### DIFF
--- a/lib/PuppeteerSharp.Tests/Page/CookiesTests.cs
+++ b/lib/PuppeteerSharp.Tests/Page/CookiesTests.cs
@@ -1,4 +1,5 @@
-﻿using System.Linq;
+﻿using System;
+using System.Linq;
 using System.Threading.Tasks;
 using Xunit;
 
@@ -116,7 +117,11 @@ namespace PuppeteerSharp.Tests.Page
         [Fact]
         public async Task ShouldNotSetACookieOnABlankPage()
         {
+            var page = await Browser.NewPageAsync();
+            await page.GoToAsync(TestConstants.AboutBlank);
 
+            var exception = await Assert.ThrowsAsync<MessageException>(async () => await page.SetCookieAsync(new CookieParam { Name = "example-cookie", Value = "best" }));
+            Assert.Equal("Protocol error (Network.deleteCookies): At least one of the url and domain needs to be specified ", exception.Message);
         }
 
         [Fact]
--- a/lib/PuppeteerSharp.Tests/Page/CookiesTests.cs
+++ b/lib/PuppeteerSharp.Tests/Page/CookiesTests.cs
@@ -7,220 +7,6 @@ namespace PuppeteerSharp.Tests.Page
     [Collection("PuppeteerLoaderFixture collection")]
     public class CookiesTests : PuppeteerBaseTest
     {
-        /*
-         describe('Cookies', function() {
-    afterEach(async({page, server}) => {
-      const cookies = await page.cookies(server.PREFIX + '/grid.html', server.CROSS_PROCESS_PREFIX);
-      for (const cookie of cookies)
-        await page.deleteCookie(cookie);
-    });
-    it('should set and get cookies', async({page, server}) => {
-      await page.goto(server.PREFIX + '/grid.html');
-      expect(await page.cookies()).toEqual([]);
-      await page.evaluate(() => {
-        document.cookie = 'username=John Doe';
-      });
-      expect(await page.cookies()).toEqual([{
-        name: 'username',
-        value: 'John Doe',
-        domain: 'localhost',
-        path: '/',
-        expires: -1,
-        size: 16,
-        httpOnly: false,
-        secure: false,
-        session: true }
-      ]);
-      await page.setCookie({
-        name: 'password',
-        value: '123456'
-      });
-      expect(await page.evaluate('document.cookie')).toBe('username=John Doe; password=123456');
-      const cookies = await page.cookies();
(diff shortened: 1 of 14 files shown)


lib/PuppeteerSharp/Session.cs
```diff
@@ -36,7 +36,7 @@ public Session(Connection connection, string targetId, string sessionId)
 
         public async Task<T> SendAsync<T>(string method, dynamic args = null)
         {
-            var content = await SendAsync(method, args, true);
+            var content = await SendAsync(method, true, args);
```
Meir017: first time this overload is used ;)

lib/PuppeteerSharp.Tests/Page/CookiesTests.cs
```diff
@@ -0,0 +1,217 @@
...
+            Assert.Equal("Protocol error (Network.deleteCookies): At least one of the url and domain needs to be specified ", exception.Message);
+        }
+
+        [Fact]
+        public async Task ShouldNotSetACookieWithBlankPageURL()
+        {
+            var page = await Browser.NewPageAsync();
+            await page.GoToAsync(TestConstants.ServerUrl + "/grid.html");
+
+            var exception = await Assert.ThrowsAnyAsync<Exception>(async () => await page.SetCookieAsync(new CookieParam
```
Meir017: this does not throw :( need to debug...

lib/PuppeteerSharp.Tests/Page/CookiesTests.cs
```diff
@@ -0,0 +1,217 @@
...
+            Assert.Equal(cookie.Session, true);
+        }
+
+        [Fact(Skip = "Line182 doesn't return")]
+        public async Task ShouldSetCookiesFromAFrame()
+        {
+            var page = await Browser.NewPageAsync();
+            await page.GoToAsync(TestConstants.ServerUrl + "/grid.html");
+            await page.SetCookieAsync(new CookieParam { Name = "localhost-cookie", Value = "best" });
+            await page.EvaluateFunctionAsync(@"src => {
```
Meir017: method does not return from server

kblok: Let me know if you need a hand man

Meir017: I think it's the first use of `EvaluateFunctionAsync` with arguments

kblok: This test is using arguments https://github.com/kblok/puppeteer-sharp/blob/master/lib/PuppeteerSharp.Tests/Page/EvaluateTests.cs#L67

Meir017: I have no idea why this hangs :(

kblok: I checked out this PR, pulled from master, and I got a green test!

lib/PuppeteerSharp/Page.cs
```diff
@@ -169,6 +170,10 @@ private Page(Session client, Target target, FrameTree frameTree, bool ignoreHTTP
                 {
                     cookie.Url = Url;
                 }
+                if (cookie.Url == "about:blank")
+                {
+                    throw new Exception($"Blank page can not have cookie \"{cookie.Name}\"");
```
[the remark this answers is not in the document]
kblok: I was thinking about the same thing in my previous PR. Definitely we shouldn't use an Exception, maybe we can create a PuppeterException for general exceptions.

On the other hand it could sound as an InvalidArgumentException.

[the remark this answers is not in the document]
kblok: Don't forget about this guy @Meir017 :)

lib/PuppeteerSharp/Page.cs
```diff
@@ -169,6 +170,10 @@ private Page(Session client, Target target, FrameTree frameTree, bool ignoreHTTP
                 {
                     cookie.Url = Url;
                 }
+                if (cookie.Url == "about:blank")
+                {
+                    throw new PuppeteerException($"Blank page can not have cookie \"{cookie.Name}\"");
+                }
```
kblok: According to [Puppeteer's test](https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/test/test.js#L3253). We should let Chromium fail and throw an exception from the connection https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/lib/Connection.js#L98

kblok: According to [Puppeteer's test](https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/test/test.js#L3253). We should let Chromium fail and throw an exception from the connection https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/lib/Connection.js#L98

Meir017: and that test fails throw the correct exception from the connection, the [next test](https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/test/test.js#L3269) throws an exception here - https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/lib/Page.js#L300

kblok: Awesome :)

Meir017:
@kblok is this ready?

PR #110, opened 2018-03-31
Implemented Page.goto tests

### LINKED ISSUE #16: Implement Page.goto tests
https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/test/test.js#L1010

Meir017:
closes #16


lib/PuppeteerSharp.Tests/Page/GotoTests.cs
```diff
@@ -0,0 +1,200 @@
...
+        }
+
+        [Fact(Skip = "complicated")]
+        public async Task ShouldWaitForNetworkIdleToSucceedNavigation()
+        {
+
+        }
+
+        [Fact]
+        public async Task ShouldNotLeakListenersDuringNavigation()
```
Meir017: is this test relevant for C#?

lib/PuppeteerSharp.Tests/Page/GotoTests.cs
```diff
@@ -0,0 +1,200 @@
...
+        {
+
+        }
+
+        [Fact]
+        public async Task ShouldNotLeakListenersDuringBadNavigation()
```
Meir017: is this test relevant for C#?

lib/PuppeteerSharp/NavigatorWatcher.cs
```diff
@@ -116,31 +120,24 @@ private bool CheckLifecycle(Frame frame, IEnumerable<string> expectedLifecycle)
...
-                        new ChromeProcessException($"Navigation Timeout Exceeded: '{_timeout}'ms exceeded"));
-                    _timer.Dispose();
-                    _timer = null;
-                }, null, _timeout, 0);
-            }
-
-            return wrapper.Task;
+            {
+                await Task.Delay(_timeout);
+                throw new ChromeProcessException($"Navigation Timeout Exceeded: {_timeout}ms exceeded");
```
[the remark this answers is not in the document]
Meir017: this is the same behavior as Puppeteer

lib/PuppeteerSharp.Tests/Page/GotoTests.cs
kblok: No. It's Node-specific.

lib/PuppeteerSharp.Tests/Page/GotoTests.cs
```diff
@@ -0,0 +1,216 @@
...
+            {
+                var response = await page.GoToAsync(TestConstants.AboutBlank);
+                Assert.Null(response);
+            }
+        }
+
+        [Theory]
+        [InlineData("domcontentloaded")]
+        [InlineData("networkidle0")]
+        [InlineData("networkidle2")]
```
kblok: What do you think about making a class in Puppeteer with these constants?
Something like
```cs
public class NativationEvents
{
    public const DOMContentLoaded = "domcontentloaded";
    ...
    ...
}
```

lib/PuppeteerSharp.Tests/Page/GotoTests.cs
```diff
@@ -0,0 +1,216 @@
...
+        public async Task ShouldFailWhenNavigatingToBadUrl()
+        {
+            using (var page = await Browser.NewPageAsync())
+            {
+                var exception = await Assert.ThrowsAnyAsync<Exception>(async () => await page.GoToAsync("asdfasdf"));
+                Assert.Contains("Cannot navigate to invalid URL", exception.Message);
+            }
+        }
+
+        [Fact(Skip = "message is ERR_CERT_COMMON_NAME_INVALID instead of ERR_CERT_AUTHORITY_INVALID")]
```
kblok: Let's use the error we're getting.

lib/PuppeteerSharp/NavigatorWatcher.cs
```diff
@@ -58,11 +61,12 @@ public NavigatorWatcher(FrameManager frameManager, Frame mainFrame, int timeout,
             }).ContinueWith((task) =>
             {
                 CleanUp();
+                return task.GetAwaiter().GetResult();
```
kblok: A return in a constructor?

Meir017: it's in the `ContinueWith` method ;)

lib/PuppeteerSharp/NavigatorWatcher.cs
```diff
@@ -116,31 +120,24 @@ private bool CheckLifecycle(Frame frame, IEnumerable<string> expectedLifecycle)
...
-            var wrapper = new TaskCompletionSource<bool>();
-
+        private async Task CreateTimeoutTask()
+        {
+            var wrapper = new TaskCompletionSource<bool>();
+
             if (_timeout == 0)
             {
-                wrapper.SetResult(true);
+                await Task.Delay(-1);
```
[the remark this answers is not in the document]
Meir017: that will make it complete immediately, this way it never completes

lib/PuppeteerSharp/NavigatorWatcher.cs
kblok: of course 😛 gotcha.

lib/PuppeteerSharp/NavigatorWatcher.cs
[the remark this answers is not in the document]
kblok: You are a genius.
How about Task leaks, shouldn't we send a cancellation token and cancel it when we don't need it anymore?

lib/PuppeteerSharp/NavigatorWatcher.cs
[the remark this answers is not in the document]
Meir017: just one thing - it should be `if (!ct.IsCancellationRequested)` since we want to fail only if it wasn't cancelled

lib/PuppeteerSharp.Tests/Page/GotoTests.cs
Meir017: An enum sounds better since it's more user friendly

lib/PuppeteerSharp.Tests/Page/GotoTests.cs
```diff
@@ -0,0 +1,205 @@
...
+                Assert.Contains("net::ERR_CONNECTION_REFUSED", exception.Message);
+            }
+        }
+
+        [Fact]
+        public async Task ShouldFailWhenExceedingMaximumNavigationTimeout()
+        {
+            using (var page = await Browser.NewPageAsync())
+            {
+                var exception = await Assert.ThrowsAnyAsync<Exception>(async () => await page.GoToAsync(TestConstants.MaximumNavigationTimeout, new NavigationOptions { Timeout = 1 }));
```
kblok: I try to avoid lines over 120 characters

lib/PuppeteerSharp.Tests/Page/GotoTests.cs
Meir017: this will remove the test `ShouldThrowIfNetworkidleIsPassedAsAnOption` 👍

lib/PuppeteerSharp.Tests/Page/GotoTests.cs
```diff
@@ -0,0 +1,157 @@
...
+        }
+
+        [Fact]
+        public async Task ShouldFailWhenNavigatingToBadUrl()
+        {
+            var exception = await Assert.ThrowsAnyAsync<Exception>(async () => await Page.GoToAsync("asdfasdf"));
+            Assert.Contains("Cannot navigate to invalid URL", exception.Message);
+        }
+
+        [Fact(Skip = "message is ERR_CERT_COMMON_NAME_INVALID instead of ERR_CERT_AUTHORITY_INVALID")]
```
Meir017: I think these errors are caused by the server not returning the correct error

kblok: Just leaving this here so we have some reference https://stackoverflow.com/questions/44711260/chorme-59-neterr-cert-common-name-invalid

Meir017:
the method [`waitForRequest`](https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/test/test.js#L1111) seems almost impossible without having a reference to the server object, I think we should wait for #101

kblok:
@Meir017 agreed, and it's being used in many places. If you reach a dead-end on #101 maybe we can play with file watchers. For instance:

* Create a control listening to `fetch-request-a.js`.
* When it gets hits it would create a file in `\log\fetch-request-a.js`
* It would return the real file

Then on our side, we would listen to file created in the log folder.

I know, it sounds over-complicated, it's just an idea.

Meir017:
it's working with net471 :)
just a few more issues with the screenshot paths...

Meir017:
could be easier once #116 is merged

lib/PuppeteerSharp.Tests/Page/GotoTests.cs
```diff
@@ -0,0 +1,228 @@
...
+            Page.Load += WaitPageLoad;
+            await pageLoaded.Task;
+
+            Assert.False(navigationFinished);
+
+            await initialFetchResourcesRequested;
+
+            Assert.False(navigationFinished);
+
+            while (responses.Count != 3) await Task.Delay(10); // todo: figure out why
```
Meir017: The test was reaching the foreach on line 174 with an empty list, this worked, not sure why (debugging passes since it's slower)
same on line 189

Meir017:
https://ci.appveyor.com/project/kblok/puppeteer-sharp/build/1.0.251#L335 not sure why the test run is aborted

PhenX:
I fixed it in #125 (https://github.com/kblok/puppeteer-sharp/pull/125/files#diff-c363ddac9f50e1502fca9e72725d7090R273) but in the end, it doesn't throw any Exception when run among other tests : https://ci.appveyor.com/project/kblok/puppeteer-sharp/build/1.0.250/tests

kblok:
Let me know if you need a hand here @Meir017.

Meir017:
I'm not sure why the test run was aborted :(

kblok:
I'll take a look tomorrow morning @Meir017

PhenX:
@kblok @Meir017 the error you have is because frameId is null and we try to use it as a key to get from the dictionary, I fixed it in #125 https://github.com/kblok/puppeteer-sharp/pull/125/files#diff-c363ddac9f50e1502fca9e72725d7090R273

lib/PuppeteerSharp/NetworkManager.cs
```diff
@@ -270,7 +270,12 @@ private void HandleRequestStart(string requestId, dynamic messageData)
 
         private void HandleRequestStart(string requestId, string interceptionId, string url, string resourceType, Payload requestPayload, string frameId)
         {
-            _frameManager.Frames.TryGetValue(frameId, out var frame);
+            Frame frame = null;
+
+            if (!string.IsNullOrEmpty(frameId))
+            {
+                _frameManager.Frames.TryGetValue(frameId, out frame);
+            }
```
kblok: I double checked in NodeJS. And @PhenX 's code is valid here.

kblok:
So, I found a few things @Meir017. This is as far as I could go today.

Browser needs to call `await target.InitializedTask;` before returning the page.
https://github.com/kblok/puppeteer-sharp/blob/master/lib/PuppeteerSharp/Browser.cs#L68

And most important, the main issue is in the GoToAsync race https://github.com/kblok/puppeteer-sharp/blob/master/lib/PuppeteerSharp/Page.cs#L301

In this test `watcher.NavigationTask.Exception` is winning the race, so `Navigate` doesn't have time to fail.

kblok:
@Meir017 this piece of code turned all tests green. 

```cs
var cts = new CancellationTokenSource();
var watcher = new NavigatorWatcher(_frameManager, mainFrame, timeout, options, cts);
var navigateTask = Navigate(_client, url, referrer);

await navigateTask;
var exception = navigateTask.Exception?.InnerException;
if (exception == null)
{
    var navigation = await watcher.NavigationTask;
    exception = navigation.Exception?.InnerException;
}

watcher.Cancel();
```

TL;DR: The race was messing with us, so no race.

Let's go through the code:

Once we create the task, those guys start running
```cs
var watcher = new NavigatorWatcher(_frameManager, mainFrame, timeout, options, cts);
var navigateTask = Navigate(_client, url, referrer); 
```

The navigate task finishes not when the page was loaded, but when the navigation message was received by chromium and the URL was validated.

```cs 
 await navigateTask;
```

It's no right to say that if the navigation task finished the timeout is no longer needed, we still need to timeout, so I remove the cancellation token call.

Once there we wait for the watcher to wait for the page.

```cs 
if (exception == null)
{
    var navigation = await watcher.NavigationTask;
    exception = navigation.Exception?.InnerException;
}
```

Meir017:
now the test `ShouldFailWhenExceedingDefaultMaximumNavigationTimeout` hangs :(
going to try something else

kblok:
ouch, I don't know why it didn't fail before, was this ignored?

Meir017:
I wasn't ignored... can you cancel the current one? it's over 35 minutes...

Meir017:
is it possible to set a max duration for the builds?

Meir017:
maybe a `testsettings` (or `runsettings` not sure what it's called today...) files? https://github.com/Microsoft/vstest-docs/blob/master/docs/configure.md

kblok:
Yeah, I was looking for some `dotnet test` argument

kblok:
I got this working with these changes. Let's see what you think.

First I remove the cancelation token in the NavigatorWatcher. The clock is still ticking for him:
```cs
await Task.Delay(_timeout);
throw new ChromeProcessException($"Navigation Timeout Exceeded: {_timeout}ms exceeded");
```

Now let's see the GoTo

```cs
await Task.WhenAny(
    watcher.NavigationTask,
    navigateTask);

//Notice that I'm checking the Exception of the Result which is also the task, the original task.
AggregateException exception = navigateTask.Exception ?? watcher.NavigationTask.Result.Exception;

if (exception == null)
{
    //If we got here it means that the WhenAny finished not because of an error
    //And that's good, the navigation finished, or the Watcher finished.
    //If the watcher finished first it means that we just need the Navigate to finished building the response
    //If the navigate finished we could still fail due to timeout. 
    await Task.WhenAll(
        watcher.NavigationTask,
        navigateTask);

    exception = navigateTask.Exception ?? watcher.NavigationTask.Result.Exception;
}

```

kblok:
Got a green build here #144 @Meir017.

A few thoughts:
I simplified the task return here, it wasn't the issue but it looked odd:
https://github.com/kblok/puppeteer-sharp/pull/144/files#diff-1bd1a04b2113bae7565f71076762f3f9R61

The real fix is here:
https://github.com/kblok/puppeteer-sharp/pull/144/files#diff-10077e53565f7b058a696aa64874696bR305

The problem we were having is with the `watcher.NavigationTask` Task, I don't know guy yet, but we were getting some `"Exception thrown: 'System.AggregateException' in System.Private.CoreLib.dll"` even trying to access the `Result` property, so something was getting stuck there.

If you find a way to make that piece of code a little bit prettier, go for it, but that is working :)

Meir017:
yup. maybe we could try running them against puppeteer's SimpleServer and check the error, it might be an issue with our server

Meir017:
https://github.com/webpack/webpack-dev-server/issues/854 might be related

Meir017:
it's definitely an issue with our server.
I tested it using the following code:
```ps
cd puppeteer # (nodejs version)
node .\test\server\run.js
```
Then create a new test class without any dependencies
```cs
    public class test
    {
        [Fact]
        public async Task ShouldFailWhenNavigatingToBadSSLAfterRedirects()
        {
            using (var browser = await PuppeteerSharp.Puppeteer.LaunchAsync(TestConstants.DefaultBrowserOptions(), TestConstants.ChromiumRevision))
            {

                var exception = await Assert.ThrowsAnyAsync<Exception>(async () => await page.GoToAsync(TestConstants.HttpsPrefix + "/redirect/2.html"));
                Assert.Contains("net::ERR_CERT_AUTHORITY_INVALID", exception.Message);
            }
        }
    }
```

test passes...

kblok:
@Meir017 How about using puppeteer's certs?

Meir017:
I can't seem to find a way to parse the private key :(
it starts with
`-----BEGIN ENCRYPTED PRIVATE KEY-----`
The only thing I'm able to parse is a `-----BEGIN RSA PRIVATE KEY-----`

kblok:
So those guys are committed in the repo https://github.com/GoogleChrome/puppeteer/tree/master/test/server

I had to do nothing to run Puppeteer's Simple Server. What would happen if we just copy those certs? Just asking...

Meir017:
there isn't a way to convert them into a `X509Certificate2` object

kblok:
@Meir017 how about merging this PR and keep digging into the SSL hell in a new one? This guy is fixing some important issues.

Meir017:
Good idea. We just need an issue to track the remaining work we have here

PR #119, opened 2018-04-05
Implemented a SimpleServer

### LINKED ISSUE #116: implement a SimpleServer like puppeteer
by creating a server like https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/test/server/SimpleServer.js we would be able to add routes/auths and more things more easily.

it will remove all of the test-specific routes

this depends on #101

Meir017:
closes #116


Meir017:
Just a single test failed

https://ci.appveyor.com/project/kblok/puppeteer-sharp/build/1.0.225#L356

only when using `netcoreapp2.0`

🤔

kblok:
@Meir017 the problem I think we are having here is that we run tests in parallels and we have a server in a static field, makes sense?

kblok:
Could your remove the controllers we're not going to use anymore? (like the auth one)

kblok:
I think I found the issue. I have a few ideas for you.

I added a few console.writelines here:
https://github.com/kblok/puppeteer-sharp/pull/119/files#diff-231086e53074993b9c85d09cd4295476R45

and here as:
https://github.com/kblok/puppeteer-sharp/pull/119/files#diff-554899e7b490512df5a6428184dc3e31R18

It turns out that the `Request` object is being changed between one moment and the other
screen shot 2018-04-08 at 2 22 10 pm

I was able to fix it by calling an `action` before the request is changed:

```cs
_requestSubscribers = new ConcurrentDictionary<string, Action<HttpRequest>>();
....
if (_requestSubscribers.TryGetValue(context.Request.Path, out var subscriber))
{
    subscriber(context.Request);
}
```

```cs
public async Task<T> WaitForRequest<T>(string path, Func<HttpRequest, T> selector)
{
    var taskCompletion = new TaskCompletionSource<T>();
    _requestSubscribers.Add(path, (httpRequest) =>
    {
        taskCompletion.SetResult(selector(httpRequest));
    });

    var request = await taskCompletion.Task;
    _requestSubscribers.Remove(path);

    return request;
}
```

kblok:
Nice!

lib/PuppeteerSharp.TestServer/SimpleServer.cs
```diff
@@ -0,0 +1,136 @@
...
+        public void SetRedirect(string from, string to)
+        {
+            SetRoute(from, context =>
+            {
+                context.Response.Redirect(to);
+                return Task.CompletedTask;
+            });
+        }
+
+        public async Task<T> WaitForRequest<T>(string path, Func<HttpRequest, T> selector)
```
kblok: Is the function a `selector`? That name sounds weird to me.

lib/PuppeteerSharp.Tests/Page/SetExtraHttpHeadersTests.cs
```diff
@@ -15,8 +15,10 @@ public class SetExtraHttpHeadersTests : PuppeteerPageBaseTest
                 ["Foo"] = "Bar"
             });
 
-            var response = await Page.GoToAsync($"{TestConstants.ServerUrl}/headertests/test");
-            Assert.Equal("Bar", await response.TextAsync());
+            var headerTask = Server.WaitForRequest("/empty.html", request => request.Headers["Foo"]);
+            await Task.WhenAll(Page.GoToAsync(TestConstants.EmptyPage), headerTask);
+
+            Assert.Equal("Bar", await headerTask);
```
kblok: I think `headerTask.Result` it's more clear here. The task was already awaited on 19

lib/PuppeteerSharp.TestServer/SimpleServer.cs
kblok: Good point

kblok: Good point

kblok:
This is ready to go right @Meir017?

Meir017:
yup

Issue #97, opened 2018-03-15
Use strongly typed objects instead of dynamic everywhere


Meir017:
This will probably be post v1.0 but just to keep track of this

kblok:
Yeah, I have a card for that when I was using Trello instead of Projects. Definitely, it's something we have to implement when we know the output or input structure.

PR #120, opened 2018-04-06
Implement Page.Events.Dialog

### LINKED ISSUE #18: Implement Page.Events.Dialog tests
https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/test/test.js#L1584

kblok:
closes #18

### DIFF
--- a/lib/PuppeteerSharp/DialogType.cs
+++ b/lib/PuppeteerSharp/DialogType.cs
@@ -3,6 +3,8 @@
     public enum DialogType
     {
         Alert,
-        Prompt
+        Prompt,
+        Confirm,
+        BeforeUnload
     }
 }
--- a/lib/PuppeteerSharp/DialogType.cs
+++ b/lib/PuppeteerSharp/DialogType.cs
@@ -1,0 +1,8 @@
+﻿namespace PuppeteerSharp
+{
+    public enum DialogType
+    {
+        Alert,
+        Prompt
+    }
+}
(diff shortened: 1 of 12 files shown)


lib/PuppeteerSharp/Messaging/PageJavascriptDialogOpeningMessageResponse.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using Newtonsoft.Json;
+
+namespace PuppeteerSharp.Messaging
```
kblok: @Meir017 I'm introducing this `Messaging` namespace so we can start working on #97, little by little

kblok:
Ready to review @Meir017

lib/PuppeteerSharp/DialogType.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace PuppeteerSharp
+{
+    public enum DialogType
```
Meir017: we should not have this and the class `Dialog.Type`. This should replace the constants

kblok: Agreed, How about now?

lib/PuppeteerSharp/Page.cs
```diff
@@ -714,8 +716,33 @@ private void HandleException(string exceptionDetails)
         {
         }
 
-        private void OnDialog(MessageEventArgs e)
+        private void OnDialog(PageJavascriptDialogOpeningMessageResponse message)
         {
+
+            DialogType dialogType;
+            if (message.Type == "alert")
```
Meir017: maybe create a dictionary to map the type to its string representation to clean this up?

kblok:
How about now @Meir017?

PR #121, opened 2018-04-07
Implement Page.Metrics

### LINKED ISSUE #87: Implement Page.Metrics
https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/test/test.js#L972

kblok:
closes #87


lib/PuppeteerSharp.Tests/Page/MetricsTests.cs
```diff
@@ -0,0 +1,59 @@
...
+                "ScriptDuration",
+                "TaskDuration",
+                "JSHeapUsedSize",
+                "JSHeapTotalSize",
+            };
+
+            foreach (var name in metrics.Keys)
+            {
+                Assert.Contains(name, metricsToCheck);
+                Assert.True(metrics[name] >= 0);
```
kblok: @Meir017 this looks like a dumb `Assert` to me. Puppeteer might be doing that to check if it has a number. We enforce that because we are using a real language. 
Would you leave it or remove it?

kblok:
Ready to review @Meir017

lib/PuppeteerSharp/Page.cs
```diff
@@ -101,6 +121,12 @@ private Page(Session client, Target target, FrameTree frameTree, bool ignoreHTTP
 
         #region Public Methods
 
+        public async Task<Dictionary<string, decimal>> MetricsAsync()
```
Meir017: could we use a strongly typed object instead of a `Dictionary<string, decimal>`?

kblok: This is different. It is a list. This could be a dynamic list of metrics, and I think that for the user it'd be super easy to do something like `metrics["myMetric"]`

Meir017: it's an object, not a list ;)
I just looked at https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/puppeteer/index.d.ts#L554

kblok: Well, that's not v1.0 :p

lib/PuppeteerSharp.Tests/Page/MetricsTests.cs
```diff
@@ -0,0 +1,59 @@
...
+            await Page.EvaluateExpressionAsync("console.timeStamp('test42')");
+            var result = await metricsTaskWrapper.Task;
+
+            Assert.Equal("test42", result.Title);
+            CheckMetrics(result.Metrics);
+        }
+
+        private void CheckMetrics(Dictionary<string, decimal> metrics)
+        {
+            var metricsToCheck = new List<string>{
```
Meir017: we could use the list that is declared in the `Page` class, by making it an `IEnumerable<string>`?

kblok:
Done @Meir017

lib/PuppeteerSharp.Tests/Page/MetricsTests.cs
Meir017: I guess I would leave it until we refactor the metrics to a strongly typed class

Issue #13, opened 2018-02-05
Implement Page.Events.Console


kblok:
https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/test/test.js#L918

PhenX:
I'm on it

PR #125, opened 2018-04-09
Implement OfflineMode tests

### LINKED ISSUE #7: Implement page.setOfflineMode tests
https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/test/test.js#L455

PhenX:
closes #7


PhenX:
Yes "it works locally" ©, but not on AppVeyor :(

kblok:
Yeah... I'm getting that error locally (on Mac) if I run all the tests though.

PhenX:
I'll give it a look. I think it might be due to the fact that the same Page instance is used between tests, but I'll have to confirm.

kblok:
I tried creating a new Page and even a new Browser in that test and now joy, but give it a shot.

Meir017:
each test gets a new Browser and a new Page, they are not reused between tests

PhenX:
@Meir017 that's not true if test is run among a collection (https://xunit.github.io/docs/shared-context#collection-fixture), which it was (and other tests are like this too).
I removed the collection attribute and it seems fine locally, but not on AppVeyor, maybe because it's too much isolated from other tests... I think we should include executable download/check in the initialize method ... or find something else.

Meir017:
hmm, I guess we should change it so each test gets a new instance of `Browser`.
I'm not totally sure what to do, not very familiar with xunit (I'm used to MSTest...).
Do you know what needs to change?

PhenX:
🤬

kblok:
@PhenX let's try to reduce the PR scope. We don't need to clean up all those usings right now.

lib/PuppeteerSharp.TestServer/PuppeteerSharp.TestServer.csproj
```diff
@@ -1,13 +1,11 @@
-<Project Sdk="Microsoft.NET.Sdk.Web">
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
```
kblok: Let's undo this change. It makes no difference

PhenX: Yeah, it's Rider that makes some formatting ...

PhenX: I think it's better now

PhenX: I can't see the difference on this line, maybe the BOM or different carriage return on that line ...

kblok:
Let me know if you need a hand here @PhenX

PhenX:
@kblok yes, I think I need help here ... I changed the way fixtures are implemented, to follow async pattern (with the `IAsyncLifetime` interface) because I thought it was this, and I checked that Browser and Page were always renewed, and that's the case, and as a last resot I updated Nuget packages.
Still no luck on AppVeyor, no Exception is thrown when in offline mode, while I always have one locally, wether I run only one or all of them ...

kblok:
No worries @PhenX. I'll take a look at this code tomorrow. Again, having tests coded is super helpful, this web sockets communication with chromium can be quite tricky.

In the meantime, if you want, you can take this guy #13. The implementation should be quite similar to the Dialog feature https://github.com/kblok/puppeteer-sharp/pull/120.

kblok:
I'm digging into this guy, it's related to the same error we're getting with @Meir017 here https://github.com/kblok/puppeteer-sharp/pull/110#issuecomment-380797643

We're not letting `Navigate` fail.

kblok:
How are we going with this guy @PhenX? I think if you pull from master this will improve.

kblok:
Continues on #150

PR #126, opened 2018-04-10
Fire and forget Target.closeTarget

### LINKED ISSUE #122: Page.Dispose() never completes
Using 0.3.2, if you try to dispose of a Page instance, it awaits forever. Try this, the last Console.WriteLine() call is never reached because it all hangs when the inner using statement calls Dispose on the page instance.

```
using System;
using System.IO;
using System.Threading.Tasks;
using PuppeteerSharp;

namespace PuppeteerSharpPdfDemo
{
    class MainClass
    {
        public static void Main(string[] args)
        {
            MainAsync().GetAwaiter().GetResult();
        }

        public static async Task MainAsync()
        {
            var options = new LaunchOptions
            {
                Headless = true
            };

            Console.WriteLine("Downloading chromium");
            await Downloader.CreateDefault().DownloadRevisionAsync(Downloader.DefaultRevision);

            Console.WriteLine("Navigating google");
            using (var browser = await Puppeteer.LaunchAsync(options, Downloader.DefaultRevision))
            {

                Console.WriteLine("Generating PDF");
                await page.PdfAsync(Path.Combine(Directory.GetCurrentDirectory(), "google.pdf"));

                Console.WriteLine("Export completed");
                //Console.ReadLine();
            }
            Console.WriteLine("The End"); // this line never reached :(
        }
    }
}

```

kblok:
In some scenarios `GetAwaiter().Result` might produce some deadlocks, there are many [post about this](https://blog.stephencleary.com/2012/07/dont-block-on-async-code.html).
I was never a fan of GetAwaiters because there was something in my mind saying "hey this might break", but we were forced to use `GetAwaiter` because we don't have a DisposableAsync.

So we fell into some deadlocks in console apps. 
When we lock the execution here:
https://github.com/kblok/puppeteer-sharp/blob/master/lib/PuppeteerSharp/Page.cs#L783

There we are preventing the `ReceiveAsync` method from getting the result the same `CloseAsync`, through `target.closeTarget` is waiting for.

https://github.com/kblok/puppeteer-sharp/blob/master/lib/PuppeteerSharp/Connection.cs#L134

I'm fixing this by calling that close as a trigger and forget. I don't think that all `SendAsync` that are not expecting a value should be a trigger and forget call, because is good to know if the command was successfully executed. But a close is a close, we want to get out of there as soon as possible.

closes #122

### DIFF
--- a/lib/PuppeteerSharp/Connection.cs
+++ b/lib/PuppeteerSharp/Connection.cs
@@ -52,7 +52,7 @@ namespace PuppeteerSharp
 
         #region Public Methods
 
-        public async Task<dynamic> SendAsync(string method, dynamic args = null)
+        public async Task<dynamic> SendAsync(string method, dynamic args = null, bool waitForResponse = true)
         {
             var id = ++_lastId;
             var message = JsonConvert.SerializeObject(new Dictionary<string, object>(){
@@ -74,7 +74,14 @@ namespace PuppeteerSharp
                 _closeMessageSent = true;
             }
 
-            return await _responses[id].Task;
+            if (waitForResponse)
+            {
+                return await _responses[id].Task;
+            }
+            else
+            {
+                return Task.CompletedTask;
+            }
         }
 
         private void QueueId(int id)
--- a/lib/PuppeteerSharp/Connection.cs
+++ b/lib/PuppeteerSharp/Connection.cs
@@ -52,7 +52,7 @@ namespace PuppeteerSharp
 
         #region Public Methods
 
-        public async Task<dynamic> SendAsync(string method, dynamic args = null, bool waitForResponse = true)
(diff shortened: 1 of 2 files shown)


kblok:
@Meir017

Meir017:
couldn't we just change the method `CloseAsync` to not wait for the response instead of changing the `Connection` class?

kblok:
The only thing we'd win would be reducing the scope and avoid adding a parameter only for one call right?
Or do you see another risk on this code @Meir017?

Meir017:
It seems a bit weird for an async method to return a task that does nothing, I think it's a bit misleading

kblok:
@Meir017 CloseAsync is still returning a Task so you can await it if you want, for instance, if you don't use a using.

PR #129, opened 2018-04-11
Make navigation options all public!

cdroulers:
Add at least one test that uses the options!

### DIFF
--- a/lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
+++ b/lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
@@ -38,6 +38,19 @@ namespace PuppeteerSharp.Tests.Puppeteer
             }
         }
 
+        [Fact]
+        public async Task ShouldWorkInRealLifeWithOptions()
+        {
+            var options = TestConstants.DefaultBrowserOptions();
+
+            using (var browser = await PuppeteerSharp.Puppeteer.LaunchAsync(options, TestConstants.ChromiumRevision))
+            using (var page = await browser.NewPageAsync())
+            {
+                var response = await page.GoToAsync("https://www.google.com", new NavigationOptions() { Timeout = 5000, WaitUntil = new[] { "networkidle0" } });
+                Assert.Equal(response.Status.ToString(), "OK");
+            }
+        }
+
         [Fact(Skip = "test is not part of v1.0.0")]
         public async Task NetworkRedirectsShouldReportSecurityDetails()
         {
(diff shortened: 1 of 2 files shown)


cdroulers:
Slightly irrelevant, but couldn't `testCert.cer` be commited along the code since it's a self-signed test certificate? It would make contributing slightly easier!

kblok:
I could try to upload my certificate, but as it's a self-signed I don't know if it would work. That's why I created a tutorial instead https://github.com/kblok/puppeteer-sharp/blob/master/CONTRIBUTING.md#getting-setup :p

kblok:
Could you create an issue, asking for the certificate?

lib/PuppeteerSharp/NavigationOptions.cs
```diff
@@ -2,7 +2,8 @@
 {
     public class NavigationOptions
     {
-        public int? Timeout { get; internal set; }
-        public string[] WaitUntil { get; internal set; }
+        public int? Timeout { get; set; }
+
```
kblok: Let's remove the extra line

lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
```diff
@@ -38,6 +38,19 @@ public class PuppeteerLaunchTests : PuppeteerBaseTest
...
 
+        [Fact]
+        public async Task ShouldWorkInRealLifeWithOptions()
+        {
+            var options = TestConstants.DefaultBrowserOptions();
+
+            using (var browser = await PuppeteerSharp.Puppeteer.LaunchAsync(options, TestConstants.ChromiumRevision))
+            using (var page = await browser.NewPageAsync())
+            {
+                var response = await page.GoToAsync("https://www.google.com", new NavigationOptions() { Timeout = 5000, WaitUntil = new[] { "networkidle0" } });
```
kblok: Can we make this multi-line?

```cs 
var response = await page.GoToAsync("https://www.google.com", new NavigationOptions
{ 
    Timeout = 5000, 
    WaitUntil = new[] { "networkidle0" }
});
```

cdroulers: Stylecop recommends putting each parameter on its own line in that case, is that ok?

```csharp
                var response = await page.GoToAsync(
                    "https://www.google.com",
                    new NavigationOptions()
                    {
                        Timeout = 5000,
                        WaitUntil = new[] { "networkidle0" }
                    });
```

lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
```diff
@@ -38,6 +38,19 @@ public class PuppeteerLaunchTests : PuppeteerBaseTest
...
+                Assert.Equal(response.Status.ToString(), "OK");
```
Meir017: use `HttpStatusCode.OK` instead

cdroulers: Will change all in this file!

kblok:
I'd like your approval @Meir017

kblok:
Congratulations @cdroulers!

cdroulers:
@kblok As for the certificate, I figured out how to create one anyway and it worked, I just realized that it probably wouldn't out of the box because the certificate needs to be trusted, otherwise we'd get HTTPS errors. So instructions are fine, really!

PR #134, opened 2018-04-12
Implement Console Events and Page.close tests

### LINKED ISSUE #5: Implement Page.close tests
https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/test/test.js#L321

### LINKED ISSUE #13: Implement Page.Events.Console

PhenX:
closes #13 
closes #5

### DIFF
--- a/lib/PuppeteerSharp.Tests/Issues/Issue0100.cs
+++ b/lib/PuppeteerSharp.Tests/Issues/Issue0100.cs
@@ -7,17 +7,6 @@ namespace PuppeteerSharp.Tests.Issues
     [Collection("PuppeteerLoaderFixture collection")]
     public class Issue0100 : PuppeteerBaseTest
     {
-        [Fact(Skip = "It's locking our tests, I'll try to run it again when get have GoToAsync read")]
-        public async Task PdfWundergroundShouldWork()
-        {
-            using (var page = await Browser.NewPageAsync())
-            {
-                var response = await page.GoToAsync("https://www.wunderground.com/weather/be/antwerp");
-                var pdf = await page.PdfStreamAsync();
-                Assert.NotNull(pdf);
-            }
-        }
-
         [Fact]
         public async Task PdfDarkskyShouldWork()
         {


lib/PuppeteerSharp/ConsoleMessage.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        public ConsoleMessage(string type, IList<JSHandle> args)
+        {
+            Type = GetConsoleType(type);
+            Args = args;
+        }
+
+        public string Text()
+        {
+            return string.Join(" ", Args);
```
Meir017: use arrow function

lib/PuppeteerSharp/Messaging/PageConsoleResponse.cs
```diff
@@ -0,0 +1,16 @@
+﻿using Newtonsoft.Json;
+
+namespace PuppeteerSharp.Messaging
+{
+    public class PageConsoleResponse
+    {
+        [JsonProperty("type")]
+        public string Type { get; set; }
```
Meir017: `Newtonsoft.Json` knows how to deserialize enums (`StringEnumConverter`) so we can use `ConsoleType` here and remove the conversion in the 
`ConsoleMessage` class

PhenX: @Meir017 I got "inspiration" from Dialog.cs made by @kblok , which uses a mapping method but you are right. @kblok what do you think ?

kblok: If you can also improve Dialog go for it :)

Meir017: It only works when deserializing since:
```cs
JsonConvert.SerializeObject(DialogType.BeforeUnload)
// output: "\"beforeUnload\"" 
```
and not `beforeunload`

Meir017: A possible solution is to use `[EnumMember(Value = "beforeunload")]` for enum values where the default cameCase doesn't work, does this make sense?

PhenX: I made the changes, but didn't include "[JsonConverter(typeof(StringEnumConverter), true)]", do you confirm it's necessary ? It works find without it to deserialize, do we need to serialize these values too ?

kblok: If we don't really need it, don't add it.

lib/PuppeteerSharp/Page.cs
```diff
@@ -761,8 +763,37 @@ private void OnDialog(PageJavascriptDialogOpeningResponse message)
...
+                foreach (var arg in message.Args)
+                {
+                    await Helper.ReleaseObject(_client, arg);
+                }
+                
+                return;
+            }
+            
+            var handles = message.Args
+                .Select(_ => (JSHandle)_frameManager.CreateJsHandle(message.ExecutionContextId, _))
```
Meir017: do we need to cast here?

lib/PuppeteerSharp/Page.cs
```diff
@@ -761,8 +763,37 @@ private void OnDialog(PageJavascriptDialogOpeningResponse message)
             Dialog?.Invoke(this, new DialogEventArgs(dialog));
         }
 
-        private void OnConsoleAPI(MessageEventArgs e)
+        private async Task OnConsoleAPI(PageConsoleResponse message)
         {
+            // TODO Do we need to translate this part ?
```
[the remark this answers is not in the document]
PhenX: I think I have my answer :)

lib/PuppeteerSharp/Page.cs
PhenX: Yes, I don't really understand why, but it doesn't recognize the return type of CreateJsHandle : 

    Unable to cast object of type 'System.Collections.Generic.List<dynamic>' to type 'System.Collections.Generic.IList<PuppeteerSharp.JSHandle>'

lib/PuppeteerSharp/ConsoleType.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace PuppeteerSharp
+{
+    public enum ConsoleType
```
Meir017: I this should have more options - https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/puppeteer/index.d.ts#L133

lib/PuppeteerSharp/Page.cs
Meir017: it's because `message.Args` is dynamic

lib/PuppeteerSharp/Messaging/PageConsoleResponse.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace PuppeteerSharp.Messaging
+{
+    public class PageConsoleResponse
+    {
+        [JsonProperty("type")]
+        public ConsoleType Type { get; set; }
+        
+        [JsonProperty("args")]
+        public dynamic[] Args { get; set; }
```
Meir017: should be `JSHandle[]`

lib/PuppeteerSharp/Page.cs
PhenX: Yes,  that's normal, it's the original data, not yet transformed into a JSHandle

Meir017: too bad.
ok

kblok:
Yeah I'm ok with that @PhenX @Meir017, change the description so the project can track it

PhenX:
@Meir017 @kblok thanks I was starting to panic :)

lib/PuppeteerSharp.Tests/Page/Events/ConsoleTests.cs
```diff
@@ -0,0 +1,99 @@
...
+            await Page.EvaluateExpressionAsync("console.log('hello', 5, {foo: 'bar'})");
+
+            Assert.Equal("hello 5 JSHandle@object", message.Text());
+            Assert.Equal(ConsoleType.Log, message.Type);
+
+            Assert.Equal("hello", (await message.Args[0].JsonValue()).ToString());
+            Assert.Equal("5", (await message.Args[1].JsonValue()).ToString());
+
+            // TODO Json Serialization is not exactly the same as in Puppeteer
+            Assert.Equal("{\r\n  \"foo\": \"bar\"\r\n}", (await message.Args[2].JsonValue()).ToString());
```
kblok: Here we should get a json, object, and in line 30 we should get a 5 as an int.

I think that [ValueFromRemoteObject](https://github.com/kblok/puppeteer-sharp/blob/master/lib/PuppeteerSharp/Helper.cs#L22) should try to parse that as a JSON value. I should succeed parsing strings, numbers, and objects.

lib/PuppeteerSharp/FrameManager.cs
```diff
@@ -104,23 +104,28 @@ private void OnExecutionContextDestroyed(int executionContextId)
...
-                }
-                if (remoteObject.subtype == "node")
-                {
-                    return new ElementHandle(storedContext, _client, remoteObject, _page);
-                }
-                return new JSHandle(storedContext, _client, remoteObject);
-            });
+        private void OnExecutionContextCreated(ContextPayload contextPayload)
+        {
+
```
kblok: Let's remove this empty line

lib/PuppeteerSharp/Page.cs
kblok: First time I see `_` in use, fancy!

What would be the benefit of using `_` instead of defaulting that argument to null in the method?

lib/PuppeteerSharp.Tests/Page/Events/ConsoleTests.cs
PhenX: You are right, I'll change it to use remoteObject.type (https://chromedevtools.github.io/devtools-protocol/tot/Runtime#type-RemoteObject) with a switch to handle all the cases, what do you think ?

PhenX:
The "_" is not null, it corresponds to the items from the Args array. I use underscore for Queryables and Enumerables by habit :)

PhenX:
Hum, I'm not sure to understand, do we agree that this : 

```c#
    var handles = message.Args
        .Select(_ => (JSHandle)_frameManager.CreateJsHandle(message.ExecutionContextId, _))
```
is the same as this ?

```c#
    var handles = message.Args
        .Select(item => (JSHandle)_frameManager.CreateJsHandle(message.ExecutionContextId, item))
```
or even this ?
```c#
    var handles = new List<JSHandle>();
    foreach (var arg in message.Args)
    {
        handles.Add(_frameManager.CreateJsHandle(message.ExecutionContextId, arg));
    }
```

Meir017:
I think that once one of the argument is of type `dynamic` then all of the expression is dynamic but I'm not sure.
The casting will be removed once everything is strongly typed

PhenX:
@Meir017 yes, I think that's exactly this, I see it with the remoteObject object, I need to convert it explicitly  to string in a string variable to get something other than dynamic.

lib/PuppeteerSharp.Tests/Page/Events/ConsoleTests.cs
kblok: Yeah, sounds good

Meir017: maybe make `JsonValue` return `Task<T>` instead of `Task<object>`?

kblok: Thinking about this a little bit, the doc says:
>Remote object value in case of primitive values or JSON values (if it was requested).

So if we fall there it's because it's serializable, you can just use JSON.Parse in one line

PhenX: I didn't find a way to make it work for the third assert (with `{foo: "bar"}`) without using generics, and for the second (with "5"), there are no integers in JS, only numbers, so I have to force it to float. You'll see all this in my next commit.

kblok:
How are we going with this guy @PhenX ?

PhenX:
What do you think about my latest changes on jsonvalue?

lib/PuppeteerSharp.Tests/Page/Events/ConsoleTests.cs
```diff
@@ -22,15 +23,15 @@ void EventHandler(object sender, ConsoleEventArgs e)
...
             Assert.Equal(ConsoleType.Log, message.Type);
 
-            Assert.Equal("hello", (await message.Args[0].JsonValue()).ToString());
-            Assert.Equal("5", (await message.Args[1].JsonValue()).ToString());
-
-            // TODO Json Serialization is not exactly the same as in Puppeteer
-            Assert.Equal("{\r\n  \"foo\": \"bar\"\r\n}", (await message.Args[2].JsonValue()).ToString());
+            Assert.Equal("hello", await message.Args[0].JsonValue());
+            Assert.Equal(5, await message.Args[1].JsonValue<float>());
+            Assert.Equal(obj,await message.Args[2].JsonValue<Dictionary<string, object>>());
```
kblok: How about testing this?
```cs
Assert.Equal("bar", await message.Args[2].JsonValue<dynamic>().foo);
```

lib/PuppeteerSharp/Helper.cs
```diff
@@ -19,7 +20,38 @@ internal static object ValueFromRemoteObject(dynamic remoteObject)
...
+            }
+
+            // https://chromedevtools.github.io/devtools-protocol/tot/Runtime#type-RemoteObject
+            string objectValue = remoteObject.value.ToString();
+            string objectType = remoteObject.type.ToString();
+            
+            switch (objectType)
+            {
+                case "object":
+                    return JsonConvert.DeserializeObject<T>(objectValue);
```
kblok: I think JsonConvert will be able to serialize a "number" a "boolean" and a "bigint", what do you think?

kblok:
Looking good!
Could you try pulling from master and see if it builds?

lib/PuppeteerSharp.Tests/Page/Events/ConsoleTests.cs
PhenX: Yes it could be in addition to the current test

lib/PuppeteerSharp/Helper.cs
PhenX: Yes it's pretty sure it can

PhenX:
I'll try and do that tomorrow. When do you think you'll release 0.4?

kblok:
I want to release it during the weekend, we are almost there

lib/PuppeteerSharp.Tests/Page/CloseTests.cs
```diff
@@ -0,0 +1,22 @@
...
+    [Collection("PuppeteerLoaderFixture collection")]
+    public class CloseTests : PuppeteerBaseTest
+    {
+        [Fact]
+        public async Task ShouldRejectAllPromisesWhenPageIsClosed()
+        {
+            var newPage = await Browser.NewPageAsync();
+            var neverResolves = newPage.EvaluateFunctionAsync("() => new Promise(r => {})");
+            
+            await newPage.CloseAsync();
```
kblok: One thing I noticed on Puppeteer is that they are not awaiting the close
https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/test/test.js#L325

lib/PuppeteerSharp/ConsoleMessage.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System.Collections.Generic;
+
+namespace PuppeteerSharp
+{
+    public class ConsoleMessage
+    {
+        public ConsoleType Type { get; }
+        public IList<JSHandle> Args { get; }
+
+        public ConsoleMessage(ConsoleType type, IList<JSHandle> args)
```
kblok: ConsoleMessage has a text property which is different from the args

https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/lib/Page.js#L447

PhenX: Yes, at the beginning it was present, I didn't see the purpose ... I replaced it by this in ConsoleMessage which does the same 

```cs
    public string Text() => string.Join(" ", Args);
```

If really needed I'll change it

kblok: Yup, we have to honor Puppeteer

PhenX:
```
(•_•)
( •_•)>⌐■-■
(⌐■_■)
```

kblok:
Looks good to me, what do you think @Meir017 ?

kblok:
@PhenX you have officially completed version 0.4 :)

PhenX:
:D

PR #143, opened 2018-04-14
Tests timeout

kblok:
Set to 8 minutes


kblok:
@Meir017, Once this get to master I'll setup AppVeyor to use it

PR #145, opened 2018-04-16
Process exit improvements

### LINKED ISSUE #128: Exception when Chrome fails to launch
If Chrome somehow fails to launch, the puppeteer fails to read the error message. Instead, I get the following exception:

```
Unhandled Exception: System.InvalidOperationException: Cannot mix synchronous and asynchronous operation on process stream.
at System.Diagnostics.Process.get_StandardError()
at PuppeteerSharp.Launcher.<>c__DisplayClass19_0.<WaitForEndpoint>b__0(Object sender, EventArgs e)
at System.Diagnostics.Process.OnExited()
at System.Diagnostics.Process.RaiseOnExited()
at System.Diagnostics.Process.CompletionCallback(Object context, Boolean wasSignaled)
at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
```

which occurs at this line: https://github.com/kblok/puppeteer-sharp/blob/master/lib/PuppeteerSharp/Launcher.cs#L241

I think it is because this line https://github.com/kblok/puppeteer-sharp/blob/master/lib/PuppeteerSharp/Launcher.cs#L287 is asynchronous, while the former is synchronous.

kblok:
* Check if the process is running before killing it.
* Grab the output from the output variable instead of trying to read the stream again.

Closes #128

### DIFF
--- a/lib/PuppeteerSharp/Launcher.cs
+++ b/lib/PuppeteerSharp/Launcher.cs
@@ -153,10 +153,10 @@ namespace PuppeteerSharp
                 var keepAliveInterval = options.KeepAliveInterval;
 
                 _connection = await Connection.Create(browserWSEndpoint, connectionDelay, keepAliveInterval);
+                _processLoaded = true;
 
                 if (options.LogProcess)
                 {
-                    _processLoaded = true;
                     Console.WriteLine($"PROCESS COUNT: {Interlocked.Increment(ref _processCount)}");
                 }
 
@@ -169,7 +169,7 @@ namespace PuppeteerSharp
             }
 
         }
-        
+
         public async Task TryDeleteUserDataDir(int times = 10, TimeSpan? delay = null)
         {
             if (!IsChromeClosed)
@@ -226,7 +226,7 @@ namespace PuppeteerSharp
         #endregion
 
         #region Private methods
-        
+
         private Task<string> WaitForEndpoint(Process chromeProcess, int timeout, bool dumpio)
         {
             var taskWrapper = new TaskCompletionSource<string>();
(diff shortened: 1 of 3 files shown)


lib/PuppeteerSharp/Launcher.cs
```diff
@@ -156,6 +156,7 @@ public Launcher()
 
                 if (options.LogProcess)
                 {
+                    _processLoaded = true;
```
Meir017: I think this should not be inside the `if` statement

lib/PuppeteerSharp/Launcher.cs
```diff
@@ -236,10 +237,14 @@ private Task<string> WaitForEndpoint(Process chromeProcess, int timeout, bool du
...
+                if (_options.LogProcess && !_processLoaded)
+                {
+                    Console.WriteLine($"PROCESS COUNT: {Interlocked.Increment(ref _processCount)}");
+                }
+
                 CleanUp();
 
-                var error = chromeProcess.StandardError.ReadToEnd();
-                taskWrapper.SetException(new ChromeProcessException($"Failed to launch chrome! {error}"));
+                taskWrapper.SetException(new ChromeProcessException($"Failed to launch chrome! {output}"));
```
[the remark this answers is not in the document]
kblok: `output` is being concatenated here https://github.com/kblok/puppeteer-sharp/blob/320fcc2876a54b9daac8c16160e818f41b3c7660/lib/PuppeteerSharp/Launcher.cs#L254

We first get the output and then the process exits.

lib/PuppeteerSharp/Launcher.cs
kblok: true.

kblok: Done

PR #147, opened 2018-04-17
Implemented Page.waitFor tests

### LINKED ISSUE #141: Implement Page.waitFor
https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/test/test.js#L887

Meir017:
closes #141

### DIFF
--- a/lib/PuppeteerSharp/Frame.cs
+++ b/lib/PuppeteerSharp/Frame.cs
@@ -191,8 +191,8 @@ namespace PuppeteerSharp
 
         internal Task WaitForTimeoutAsync(int milliseconds) => Task.Delay(milliseconds);
 
-        internal Task WaitForFunctionAsync(string function, WaitForFunctionOptions options, params object[] args)
-            => new WaitTask(this, function, options.Polling, options.Timeout, args).Task;
+        internal Task WaitForFunctionAsync(string script, WaitForFunctionOptions options, params object[] args)
+            => new WaitTask(this, script, options.Polling, options.Timeout, args).Task;
 
         internal Task WaitForSelectorAsync(string selector, WaitForSelectorOptions options)
         {
@@ -213,7 +213,7 @@ namespace PuppeteerSharp
                 return !!(rect.top || rect.bottom || rect.width || rect.height);
               }
             }";
-            var polling = options.Visible || options.Hidden ? "raf" : "mutation";
+            var polling = options.Visible || options.Hidden ? WaitForFunctionPollingOption.Raf : WaitForFunctionPollingOption.Mutation;
             return WaitForFunctionAsync(predicate, new WaitForFunctionOptions
             {
                 Timeout = options.Timeout,
--- a/lib/PuppeteerSharp/Frame.cs
+++ b/lib/PuppeteerSharp/Frame.cs
@@ -12,10 +12,11 @@ namespace PuppeteerSharp
         private Frame _parentFrame;
         private string _defaultContextId = "<not-initialized>";
         private object _context = null;
-        private string _url = string.Empty;
-        private List<WaitTask> _waitTasks;
+        private string _url = string.Empty;        
         private bool _detached;
         private TaskCompletionSource<ElementHandle> _documentCompletionSource;
+
+        internal List<WaitTask> WaitTasks;
 
         public Frame(Session client, Page page, Frame parentFrame, string frameId)
         {
@@ -31,7 +32,7 @@ namespace PuppeteerSharp
 
             SetDefaultContext(null);
 
-            _waitTasks = new List<WaitTask>();
+            WaitTasks = new List<WaitTask>();
             LifecycleEvents = new List<string>();
         }
 
@@ -163,7 +164,7 @@ namespace PuppeteerSharp
             {
                 ContextResolveTaskWrapper.SetResult(context);
 
-                foreach (var waitTask in _waitTasks)
+                foreach (var waitTask in WaitTasks)
                 {
                     waitTask.Rerun();
                 }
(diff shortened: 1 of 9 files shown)


lib/PuppeteerSharp/WaitTask.cs
```diff
@@ -1,17 +1,192 @@
...
+            }
+
+            if (_terminated || runCount != _runCount)
+            {
+                if (success != null) await success.Dispose();
+                return;
+            }
+            if (exception != null && await _frame.EvaluateFunctionAsync<bool>("s => !s", success))
+            {
+                if (success != null) await success.Dispose();
```
Meir017: we can't `await success?.Dispose();` since https://github.com/dotnet/csharplang/issues/35

lib/PuppeteerSharp/WaitTask.cs
```diff
@@ -1,17 +1,192 @@
...
+      else
+        setTimeout(onTimeout, pollInterval);
+    }
+  }
+}";
+
+        internal WaitTask(Frame frame, string predicateBody, string polling, int timeout, object[] args)
+        {
+            _frame = frame;
+            _predicateBody = $"return ( {predicateBody} )(...args)"; //predicateBody;
```
Meir017: not sure we need a conditional operator here since `predicateBody` is always a string

Meir017:
ci failed:
test: `PuppeteerSharp.Tests.Puppeteer.PuppeteerLaunchTests.ShouldWorkInRealLifeWithOptions`
PuppeteerSharp.NavigationException : Navigation Timeout Exceeded: 5000ms exceeded
---- PuppeteerSharp.ChromeProcessException : Navigation Timeout Exceeded: 5000ms exceeded

perhaps we should use a local page instead of real-life?

lib/PuppeteerSharp/WaitTask.cs
kblok: We can't validate that the predicate is valid, but maybe we can throw an `InvalidArgumentException` if null or empty.

kblok:
@Meir017 let's wait and see if it's a real pita.

Meir017:
CI failed for the same reason again :(

kblok:
Only this PR is failing due to timeout, are you sure we aren't breaking anything?

Meir017:
that test is pretty long:
- https://ci.appveyor.com/project/kblok/puppeteer-sharp/build/1.0.336/tests - ShouldWorkInRealLifeWithOptions 4 sec 17 ms
- https://ci.appveyor.com/project/kblok/puppeteer-sharp/build/1.0.330/tests - ShouldWorkInRealLifeWithOptions 4 sec 803 ms
- https://ci.appveyor.com/project/kblok/puppeteer-sharp/build/1.0.328/tests - ShouldWorkInRealLifeWithOptions 5 sec 489 ms

kblok:
We don't need `should throw when unknown type` because our functions are explicit right?

lib/PuppeteerSharp/Page.cs
```diff
@@ -535,6 +535,15 @@ public Task<T> EvaluateFunctionAsync<T>(string script, params object[] args)
             return navigationTask.Result;
         }
 
+        public Task WaitForTimeoutAsync(int milliseconds)
+            => MainFrame.WaitForTimeoutAsync(milliseconds);
+
+        public Task WaitForFunctionAsync(string script, WaitForFunctionOptions options = null, params object[] args)
```
kblok: Can we add this override?
`public Task WaitForFunctionAsync(string script, params object[] args)`

or the `params object[]` gets in the middle?

lib/PuppeteerSharp/WaitTask.cs
```diff
@@ -1,17 +1,197 @@
...
+        internal async void Rerun()
+        {
+            var runCount = ++_runCount;
+            JSHandle success = null;
+            Exception exception = null;
+
+            var context = await _frame.GetExecutionContextAsync();
+            try
+            {
+                success = await context.EvaluateFunctionHandleAsync(WaitForPredicatePageFunction, new object[] { _predicateBody, _polling, _timeout }.Concat(_args).ToArray());
```
kblok: Let's break this line

lib/PuppeteerSharp/WaitTask.cs
```diff
@@ -1,17 +1,197 @@
...
+                if (success != null) await success.Dispose();
+                return;
+            }
+            if (exception != null && await _frame.EvaluateFunctionAsync<bool>("s => !s", success))
+            {
+                if (success != null) await success.Dispose();
+                return;
+            }
+
+            if (exception?.Message.Contains("Execution context was destroyed") == true)
```
kblok: No need to `== true`

lib/PuppeteerSharp/WaitTask.cs
```diff
@@ -1,17 +1,197 @@
...
+            {
+                return;
+            }
+
+            if (exception?.Message.Contains("Cannot find context with specified id") == true)
```
kblok: No need to `== true`

lib/PuppeteerSharp/WaitTask.cs
Meir017: since we use the null conditional operator we have to do it this way

lib/PuppeteerSharp/WaitTask.cs
Meir017: since we use the null conditional operator we have to do it this way

lib/PuppeteerSharp/Frame.cs
```diff
@@ -12,11 +12,12 @@ public class Frame
         private Frame _parentFrame;
         private string _defaultContextId = "<not-initialized>";
         private object _context = null;
-        private string _url = string.Empty;
-        private List<WaitTask> _waitTasks;
+        private string _url = string.Empty;        
         private bool _detached;
         private TaskCompletionSource<ElementHandle> _documentCompletionSource;
 
+        internal List<WaitTask> WaitTasks;
```
kblok: This should be expressed as a getter property right?
```
internal List<WaitTask> WaitTask { get; }
```

Meir017: yup

PR #157, opened 2018-04-19
Implemented Frame.waitForFunction tests

### LINKED ISSUE #139: Implement Frame.waitforFunction
https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/test/test.js#L648

Meir017:
closes #139 

the following tests aren't relevant for C#:
- should accept a string (always a string...)
- should throw on bad polling value (we use an enum)

### DIFF
--- a/lib/PuppeteerSharp/WaitTask.cs
+++ b/lib/PuppeteerSharp/WaitTask.cs
@@ -10,6 +10,7 @@ namespace PuppeteerSharp
         private readonly Frame _frame;
         private readonly string _predicateBody;
         private readonly WaitForFunctionPollingOption _polling;
+        private readonly int? _pollingInterval;
         private readonly int _timeout;
         private readonly object[] _args;
         private readonly Task _timeoutTimer;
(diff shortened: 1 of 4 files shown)


Meir017:
in puppeteer `options.polling ` in the function `waitForFunction` can be a `number` or `'mutation'` or `'raf'`
not sure yet what the best way would be for us

lib/PuppeteerSharp/WaitTask.cs
```diff
@@ -107,24 +108,29 @@ internal class WaitTask
...
 }";
 
-        internal WaitTask(Frame frame, string predicateBody, WaitForFunctionPollingOption polling, int timeout, object[] args)
+        internal WaitTask(Frame frame, string predicateBody, WaitForFunctionPollingOption polling, int? pollingInterval, int timeout, object[] args)
         {
             if (string.IsNullOrEmpty(predicateBody))
             {
                 throw new ArgumentNullException(nameof(predicateBody));
             }
+            if(pollingInterval <= 10)
```
kblok: Missing space between `if` and `(`

kblok:
I like your solution, Polling and PollingInterval

Meir017:
the test `PdfDarkskyShouldWork` takes between 3 and 30+ seconds...

PR #158, opened 2018-04-20
Implemented Target.createCDPSession tests

### LINKED ISSUE #56: Implement Target.sessions tests
https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/test/test.js#L3477

Meir017:
closes #56 

added functionality required for tests:

- Session.DetachAsync
- Coverage
  - Coverage.StartJSCoverageAsync
  - Coverage.StopJSCoverageAsync - without returning reports since they aren't needed yet.

### DIFF
--- a/lib/PuppeteerSharp.Tests/PuppeteerBaseTest.cs
+++ b/lib/PuppeteerSharp.Tests/PuppeteerBaseTest.cs
@@ -36,5 +36,24 @@ namespace PuppeteerSharp.Tests
         }
 
         public void Dispose() => DisposeAsync().GetAwaiter().GetResult();
+
+        protected static Task<dynamic> WaitForEvents(Session emitter, string eventName, int eventCount = 1)
+        {
+            var completion = new TaskCompletionSource<dynamic>();
+            void handler(object sender, MessageEventArgs e)
+            {
+                if (e.MessageID != eventName) return;
+
+                --eventCount;
+                if (eventCount > 0) return;
+
+                emitter.MessageReceived -= handler;
+                completion.SetResult(e.MessageData);
+            };
+
+            emitter.MessageReceived += handler;
+
+            return completion.Task;
+        }
     }
 }
--- a/lib/PuppeteerSharp/JSCoverage.cs
+++ b/lib/PuppeteerSharp/JSCoverage.cs
@@ -71,7 +71,7 @@ namespace PuppeteerSharp
             switch (e.MessageID)
             {
                 case "Debugger.scriptParsed":
-                    OnScriptParsed(e.MessageData.ToObject<ScriptParsedResponse>());
+                    OnScriptParsed(e.MessageData.ToObject<DebuggerScriptParsedResponse>());
                     break;
                 case "Runtime.executionContextsCleared":
                     OnExecutionContextsCleared();
@@ -79,9 +79,9 @@ namespace PuppeteerSharp
             }
         }
 
-        private async void OnScriptParsed(ScriptParsedResponse scriptParseResponse)
+        private async void OnScriptParsed(DebuggerScriptParsedResponse scriptParseResponse)
         {
-            if (scriptParseResponse.Url != null)
+            if (scriptParseResponse.Url == null)
             {
                 return;
             }
--- a/lib/PuppeteerSharp/JSCoverage.cs
+++ b/lib/PuppeteerSharp/JSCoverage.cs
@@ -62,7 +62,7 @@ namespace PuppeteerSharp
                _client.SendAsync("Debugger.disable")
            );
             _client.MessageReceived -= client_MessageReceived;
-            
+
             // TODO: return coverage;            
         }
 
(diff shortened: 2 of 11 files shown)


lib/PuppeteerSharp/JSCoverage.cs
```diff
@@ -0,0 +1,110 @@
...
+            var profileResponseTask = _client.SendAsync("Profiler.takePreciseCoverage");
+            await Task.WhenAll(
+               profileResponseTask,
+               _client.SendAsync("Profiler.stopPreciseCoverage"),
+               _client.SendAsync("Profiler.disable"),
+               _client.SendAsync("Debugger.disable")
+           );
+            _client.MessageReceived -= client_MessageReceived;
+            
+            // TODO: return coverage;
```
Meir017: the result is not needed yet

lib/PuppeteerSharp.Tests/PuppeteerBaseTest.cs
```diff
@@ -36,5 +36,24 @@ public PuppeteerBaseTest()
         }
 
         public void Dispose() => DisposeAsync().GetAwaiter().GetResult();
+
+        protected static Task<dynamic> WaitForEvents(Session emitter, string eventName, int eventCount = 1)
+        {
+            var completion = new TaskCompletionSource<dynamic>();
```
kblok: Cool. you attached to the generic Message. I was wondering how to implement this when I implemented the WaitForError. Good Job

lib/PuppeteerSharp/JSCoverage.cs
```diff
@@ -0,0 +1,110 @@
...
+
+        private void client_MessageReceived(object sender, MessageEventArgs e)
+        {
+            switch (e.MessageID)
+            {
+                case "Debugger.scriptParsed":
+                    OnScriptParsed();
+                    break;
+                case "Runtime.executionContextsCleared":
+                    OnExecutionContextsCleared(e.MessageData);
```
kblok: We can create a `RuntimeExecutionContextsClearedResponse` class here. Like we're doing here https://github.com/Meir017/puppeteer-sharp/blob/c1653182074f1ecd6be8318958ee1fad539807c5/lib/PuppeteerSharp/Page.cs#L782

lib/PuppeteerSharp/JSCoverage.cs
```diff
@@ -0,0 +1,112 @@
...
+            if (!_resetOnNavigation)
+            {
+                return;
+            }
+
+            _scriptURLs.Clear();
+            _scriptSources.Clear();
+        }
+
+        private async void OnExecutionContextsCleared(RuntimeExecutionContextsClearedResponse @event)
```
kblok: We could call `@event` differently (I was expecting that when we change to `Response`) so we avoid that `@`

lib/PuppeteerSharp/JSCoverage.cs
```diff
@@ -0,0 +1,112 @@
...
+            {
+                case "Debugger.scriptParsed":
+                    OnScriptParsed();
+                    break;
+                case "Runtime.executionContextsCleared":
+                    OnExecutionContextsCleared(e.MessageData.ToObject<RuntimeExecutionContextsClearedResponse>());
+                    break;
+            }
+        }
+
```
kblok: It seems that this code is mixed up.
`OnScriptParsed` has code related to `OnExecutionContextsCleared`: 🤔 

https://github.com/GoogleChrome/puppeteer/blob/master/lib/Coverage.js#L102

Meir017: oops. good catch

Meir017:
@kblok ready

Meir017:
the test `PuppeteerSharp.Tests.Puppeteer.PuppeteerLaunchTests.UserDataDirArgument` failed 😢 

System.Net.WebSockets.WebSocketException : The 'System.Net.WebSockets.InternalClientWebSocket' instance cannot be used for communication because it has been transitioned into the 'Aborted' state.
---- System.Threading.Tasks.TaskCanceledException : A task was canceled.

kblok:
That tends to be an erratic error, rebuilding

lib/PuppeteerSharp/Messaging/ScriptParsedResponse.cs
```diff
@@ -0,0 +1,13 @@
+﻿using Newtonsoft.Json;
+
+namespace PuppeteerSharp.Messaging
+{
+    internal class ScriptParsedResponse
```
kblok: Let's call it DebuggerScriptParsedResponse

lib/PuppeteerSharp/JSCoverage.cs
```diff
@@ -0,0 +1,112 @@
...
+                    break;
+                case "Runtime.executionContextsCleared":
+                    OnExecutionContextsCleared();
+                    break;
+            }
+        }
+
+        private async void OnScriptParsed(ScriptParsedResponse scriptParseResponse)
+        {
+            if (scriptParseResponse.Url != null)
```
kblok: In Puppeteer it is `if(scriptParseResponse == null)`

```js 
if (!event.url)
```

Meir017: I always get confused when they have a `if(!expression)` 😄

PR #159, opened 2018-04-20
Frame fixes

### LINKED ISSUE #149: Change ShouldBeAbleToReconnectToADisconnectedBrowser implementation
The test is using `/frames/nested-frames.html` instead of `/frames/frame.html`

kblok:
closes #149

### DIFF
--- a/lib/PuppeteerSharp.Tests/FrameUtils.cs
+++ b/lib/PuppeteerSharp.Tests/FrameUtils.cs
@@ -22,7 +22,7 @@ namespace PuppeteerSharp.Tests
             var result = indentation + Regex.Replace(frame.Url, @":\d{4}", ":<PORT>");
             foreach (var child in frame.ChildFrames)
             {
-                result += Environment.NewLine + DumpFrames(child, "    " + indentation);
+                result += "\n" + DumpFrames(child, "    " + indentation);
             }
 
             return result;
--- a/lib/PuppeteerSharp/Page.cs
+++ b/lib/PuppeteerSharp/Page.cs
@@ -16,7 +16,7 @@ namespace PuppeteerSharp
     [DebuggerDisplay("Page {Url}")]
     public class Page : IDisposable
     {
-        public int DefaultNavigationTimeout { get; set; } = 30000;
+        public int DefaultNavigationTimeout { get; set; } = 30000000;
 
         private readonly Session _client;
         private readonly bool _ignoreHTTPSErrors;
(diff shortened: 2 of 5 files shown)


lib/PuppeteerSharp/Page.cs
```diff
@@ -16,7 +16,7 @@ namespace PuppeteerSharp
     [DebuggerDisplay("Page {Url}")]
     public class Page : IDisposable
     {
-        public int DefaultNavigationTimeout { get; set; } = 30000;
+        public int DefaultNavigationTimeout { get; set; } = 30000000;
```
Meir017: 30,000 seconds?

kblok: oops :joy:

That was messing with my debugging. Good catch.

kblok:
Ready to review @Meir017

lib/PuppeteerSharp.Tests/FrameUtils.cs
```diff
@@ -22,7 +22,7 @@ public static string DumpFrames(PuppeteerSharp.Frame frame, string indentation =
             var result = indentation + Regex.Replace(frame.Url, @":\d{4}", ":<PORT>");
             foreach (var child in frame.ChildFrames)
             {
-                result += Environment.NewLine + DumpFrames(child, "    " + indentation);
+                result += "\n" + DumpFrames(child, "    " + indentation);
```
cdroulers: Late to the party, but this makes the tests fail on Windows! Was NewLine not working on Mac or Linux? It should give "\n" on either platforms!

kblok: The CI is a Windows Machine

Meir017: maybe we should add CI on a linux machine as well?

cdroulers: Looks like AppVeyor sets the line feed when checking out files to "\n" which is why it works on the CI.

https://help.appveyor.com/discussions/problems/1119-allow-changing-git-autocrlf-setting

I'll check out a working copy with the proper setting.

kblok: Yeah. I've just got that in my local VM. We should fix this but in a more pragmatic way. I don't care if the browser returns a \n or a \r\n is not what we need to test.

I hate spending time making test happy.

PR #162, opened 2018-04-21
Respect xunit assert analyzers

### LINKED ISSUE #161: Respect xunit analyzers
we have close to 200 asserts that trigger `warning xUnit2000: The literal or constant value xxx should be the first argument in the call to 'Assert.Equal(expected, actual)'`

Meir017:
closes #161


kblok:
Beautiful

PR #164, opened 2018-04-21
Implemented Target tests

### LINKED ISSUE #54: Implement Targets tests
https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/test/test.js#L3360

Meir017:
closes #54

### DIFF
--- a/lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
+++ b/lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
@@ -84,7 +84,7 @@ namespace PuppeteerSharp.Tests.Puppeteer
                                                                             TestConstants.ChromiumRevision))
             using (var page = await browser.NewPageAsync())
             {
-                var neverResolves = page.EvaluateFunctionHandle("() => new Promise(r => {})");
+                var neverResolves = page.EvaluateFunctionHandleAsync("() => new Promise(r => {})");
                 await browser.CloseAsync();
 
                 await neverResolves;
--- a/lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
+++ b/lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
@@ -21,7 +21,7 @@ namespace PuppeteerSharp.Tests.Puppeteer
             using (var page = await browser.NewPageAsync())
             {
                 var response = await page.GoToAsync(TestConstants.HttpsPrefix + "/empty.html");
-                Assert.Equal(response.Status, HttpStatusCode.OK);
+                Assert.Equal(HttpStatusCode.OK, response.Status);
             }
         }
 
@@ -34,7 +34,7 @@ namespace PuppeteerSharp.Tests.Puppeteer
             using (var page = await browser.NewPageAsync())
             {
                 var response = await page.GoToAsync("https://www.google.com");
-                Assert.Equal(response.Status, HttpStatusCode.OK);
+                Assert.Equal(HttpStatusCode.OK, response.Status);
             }
         }
 
@@ -53,7 +53,7 @@ namespace PuppeteerSharp.Tests.Puppeteer
                         Timeout = 10000,
                         WaitUntil = new[] { WaitUntilNavigation.Networkidle0 }
                     });
-                Assert.Equal(response.Status, HttpStatusCode.OK);
+                Assert.Equal(HttpStatusCode.OK, response.Status);
             }
         }
 
@@ -159,7 +159,7 @@ namespace PuppeteerSharp.Tests.Puppeteer
             using (var page = await browser.NewPageAsync())
             {
                 var response = await page.GoToAsync(TestConstants.EmptyPage);
-                Assert.Equal(response.Status, HttpStatusCode.OK);
+                Assert.Equal(HttpStatusCode.OK, response.Status);
 
                 await browser.CloseAsync();
 
@@ -177,7 +177,7 @@ namespace PuppeteerSharp.Tests.Puppeteer
             using (var page = await browser.NewPageAsync())
             {
                 var response = await page.GoToAsync(TestConstants.EmptyPage);
-                Assert.Equal(response.Status, HttpStatusCode.OK);
+                Assert.Equal(HttpStatusCode.OK, response.Status);
             }
 
             Assert.True(launcher.IsChromeClosed);
--- a/lib/PuppeteerSharp.Tests/Target/Tests.cs
+++ b/lib/PuppeteerSharp.Tests/Target/Tests.cs
@@ -175,7 +175,7 @@ namespace PuppeteerSharp.Tests.Target
             serverResponse.Redirect("/injectedstyle.css");
             serverResponseEnd.SetResult(true);
             // Wait for the new page to load.            
-            // TODO: await waitForEvents(newPage, 'load');
+            await WaitForEvents(newPage._client, "Page.loadEventFired");
             // Cleanup.
             await newPage.CloseAsync();
         }
--- a/lib/PuppeteerSharp.Tests/Target/Tests.cs
+++ b/lib/PuppeteerSharp.Tests/Target/Tests.cs
@@ -1,0 +1,34 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace PuppeteerSharp.Tests.Target
+{
+    [Collection("PuppeteerLoaderFixture collection")]
+    public class Tests : PuppeteerPageBaseTest
+    {
+        [Fact]
+        public async Task BrowserTargetsShouldReturnAllOfTheTargets()
+        {
+            // The pages will be the testing page and the original newtab page
+            var targets = Browser.Targets();
+            Assert.Contains(targets, target => target.Type == "page"
+                && target.Url == TestConstants.AboutBlank);
+            Assert.Contains(targets, target => target.Type == "other"
+                && target.Url == string.Empty);
+        }
+
+        [Fact]
+        public async Task BrowserPagesShouldReturnAllOfThePages()
+        {
+            // The pages will be the testing page and the original newtab page
+            var allPages = (await Browser.Pages()).ToArray();
+            Assert.Equal(2, allPages.Length);
+            Assert.Contains(Page, allPages);
+            Assert.NotSame(allPages[0], allPages[1]);
+        }
+    }
+}
--- a/lib/PuppeteerSharp.Tests/Target/Tests.cs
+++ b/lib/PuppeteerSharp.Tests/Target/Tests.cs
@@ -1,4 +1,6 @@
-﻿using System.Linq;
+﻿using Microsoft.AspNetCore.Http;
+using System;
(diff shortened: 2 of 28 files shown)


lib/PuppeteerSharp.Tests/Target/Tests.cs
```diff
@@ -0,0 +1,183 @@
+﻿using Microsoft.AspNetCore.Http;
+using System;
+using System.Linq;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace PuppeteerSharp.Tests.Target
+{
+    [Collection("PuppeteerLoaderFixture collection")]
+    public class Tests : PuppeteerPageBaseTest
```
Meir017: not sure what to call this class...
in puppeteer it's just
```js
describe('Target', ...
```

lib/PuppeteerSharp/Page.cs
```diff
@@ -19,11 +19,12 @@
 namespace PuppeteerSharp
 {
     [DebuggerDisplay("Page {Url}")]
-    public class Page : IDisposable
+    public class Page : IDisposable, IEquatable<Page>
     {
         public int DefaultNavigationTimeout { get; set; } = 30000;
 
-        private readonly Session _client;
+        internal readonly Session _client;
```
Meir017: for the test ShouldNotCrashWhileRedirectingIfOriginalRequestWasMissed

lib/PuppeteerSharp/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,3 @@
+﻿using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("PuppeteerSharp.Tests")]
```
Meir017: for the test ShouldNotCrashWhileRedirectingIfOriginalRequestWasMissed

lib/PuppeteerSharp/Target.cs
```diff
@@ -36,29 +39,32 @@ public Target(Browser browser, TargetInfo targetInfo)
...
             {
+                _pageTask = await _browser.Connection.CreateSession(_targetInfo.TargetId)
+                    .ContinueWith(clientTask
+                    => PuppeteerSharp.Page.CreateAsync(clientTask.Result, this, _browser.IgnoreHTTPSErrors, _browser.AppMode, _browser.ScreenshotTaskQueue));
                 var client = await _browser.Connection.CreateSession(_targetInfo.TargetId);
-                return await PuppeteerSharp.Page.CreateAsync(client, this, _browser.IgnoreHTTPSErrors, _browser.AppMode, _browser.ScreenshotTaskQueue);
             }
 
-            return null;
+            return await (_pageTask ?? Task.FromResult<Page>(null));
```
Meir017: we can't `await` on a null value

lib/PuppeteerSharp/Browser.cs
```diff
@@ -70,14 +70,24 @@ public string WebSocketEndpoint
             return await target.Page();
         }
 
+        /// <summary>
+        /// Returns An Array of all active targets
+        /// </summary>
+        /// <returns>An Array of all active targets</returns>
+        public IEnumerable<Target> Targets() => _targets.Values.Where(target => target.IsInitialized);
```
Meir017: would it make sense to return an array instead? not sure we want this to be lazy

lib/PuppeteerSharp/Browser.cs
```diff
@@ -70,14 +70,24 @@ public string WebSocketEndpoint
...
+
+        /// <summary>
+        /// Returns a Task which resolves to an array of all open pages.
+        /// </summary>
+        /// <returns>Task which resolves to an array of all open pages.</returns>
         public async Task<IEnumerable<Page>> Pages()
```
Meir017: would it make sense to return an array instead? do we want this to be lazy?

lib/PuppeteerSharp/Browser.cs
```diff
@@ -132,7 +142,8 @@ private void ChangeTargetInfo(MessageEventArgs args)
                 throw new InvalidTargetException("Target should exists before ChangeTargetInfo");
             }
 
-            var target = _targets[args.MessageData.targetInfo.targetId.Value];
+            string targetId = args.MessageData.targetInfo.targetId.Value;
+            var target = _targets[targetId];
```
Meir017: this way the IDE knows that target is of type `Target` and show that the method `TargetInfoChanged` has a reference 😏

Meir017:
@kblok ready

lib/PuppeteerSharp.Tests/Target/Tests.cs
kblok: I'd call this TargetTests

lib/PuppeteerSharp/Browser.cs
kblok: Absolutely, let's do it

lib/PuppeteerSharp/Target.cs
```diff
@@ -36,29 +39,32 @@ public Target(Browser browser, TargetInfo targetInfo)
 
         public async Task<Page> Page()
```
kblok: This should be `PageAsync`. Let's add xml docs

lib/PuppeteerSharp.Tests/Target/Tests.cs
```diff
@@ -0,0 +1,183 @@
...
+        public async Task ShouldReportWhenANewPageIsCreatedAndClosed()
+        {
+            var otherPageTaskCompletion = new TaskCompletionSource<PuppeteerSharp.Page>();
+            async void TargetCreatedEventHandler(object sender, TargetChangedArgs e)
+            {
+                otherPageTaskCompletion.SetResult(await e.Target.Page());
+                Browser.TargetCreated -= TargetCreatedEventHandler;
+            }
+            Browser.TargetCreated += TargetCreatedEventHandler;
+            await Page.EvaluateFunctionHandle("url => window.open(url)", TestConstants.CrossProcessUrl);
```
kblok: I know this is not part of this PR, but this should be `EvaluateFunctionHandleAsync`

lib/PuppeteerSharp/Browser.cs
kblok: `PagesAsync`

lib/PuppeteerSharp.Tests/Target/Tests.cs
```diff
@@ -0,0 +1,183 @@
...
+        [Fact]
+        public async Task ShouldReportWhenATargetUrlChanges()
+        {
+            await Page.GoToAsync(TestConstants.EmptyPage);
+
+            var changedTargetTaskCompletion = new TaskCompletionSource<PuppeteerSharp.Target>();
+            void ChangedTargetEventHandler(object sender, TargetChangedArgs e)
+            {
+                changedTargetTaskCompletion.SetResult(e.Target);
+                Browser.TargetChanged -= ChangedTargetEventHandler;
```
kblok: This is because in NodeJS is `once` right?

lib/PuppeteerSharp/Browser.cs
kblok: Let's do it

lib/PuppeteerSharp/Browser.cs
kblok: And if we do ` _targets[args.MessageData.targetInfo.targetId.Value.ToString()]` ?

lib/PuppeteerSharp/Page.cs
kblok: If it's internal let's make this an internal getter property.

lib/PuppeteerSharp/Page.cs
```diff
@@ -874,5 +875,8 @@ private void OnDialog(PageJavascriptDialogOpeningResponse message)
         #region IDisposable
         public void Dispose() => CloseAsync();
         #endregion
+        #region IEquatable
+        public bool Equals(Page other) => MainFrame.Id == other.MainFrame.Id;
```
kblok: Are we sure about this equality? Where are we using this?

lib/PuppeteerSharp/Target.cs
```diff
@@ -68,5 +74,6 @@ public void TargetInfoChanged(TargetInfo targetInfo)
         /// <returns>A task that returns a <see cref="Session"/></returns>
         public Task<Session> CreateCDPSession() => _browser.Connection.CreateSession(TargetId);
 
+        public bool Equals(Target other) => TargetId.Equals(other.TargetId);
```
kblok: Same here regarding the equals, are we sure about this? where are we using it?

lib/PuppeteerSharp.Tests/Target/Tests.cs
Meir017: exactly

lib/PuppeteerSharp/Browser.cs
Meir017: since the `dynamic` keyword is part of the expression the result is also a dynamic object. The `ToString` would only be evaluated at runtime...

lib/PuppeteerSharp/Target.cs
Meir017: for `Assert.Equal` on target objects (ShouldReportWhenAServiceWorkerIsCreatedAndDestroyed)

lib/PuppeteerSharp/Page.cs
Meir017: i guess it wasn't needed, forgot why I added it 😆

lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
```diff
@@ -84,7 +84,7 @@ public class PuppeteerLaunchTests : PuppeteerBaseTest
                                                                             TestConstants.ChromiumRevision))
             using (var page = await browser.NewPageAsync())
             {
-                var neverResolves = page.EvaluateFunctionHandle("() => new Promise(r => {})");
+                var neverResolves = page.EvaluateFunctionHandleAsync("() => new Promise(r => {})");
```
kblok: No reason to this change right?

lib/PuppeteerSharp/Page.cs
```diff
@@ -18,7 +18,8 @@ public class Page : IDisposable
     {
         public int DefaultNavigationTimeout { get; set; } = 30000;
 
-        private readonly Session _client;
+        internal Session Client { get; }
```
kblok: Let's move this to the Properties region

lib/PuppeteerSharp/Target.cs
kblok: @Meir017 👆

lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerLaunchTests.cs
kblok: Friendly reminder @Meir017  :)

Meir017: the method `EvaluateFunctionHandle` was renamed to `EvaluateFunctionHandleAsync`

kblok: _facepalm_ you're right

PR #167, opened 2018-04-23
Removed dynamic evaluate functions

### LINKED ISSUE #166: ERROR: PuppeteerSharp.MessageException : Protocol error (Runtime.callFunctionOn): Object reference chain is too long
when using `EvaluateExpressionAsync` or `EvaluateFunctionAsync` to return a `dynamic` object of something big the following exception is thrown:
PuppeteerSharp.MessageException : Protocol error (Runtime.callFunctionOn): Object reference chain is too long.

to reproduce this:
```cs
public class test
{
    [Fact]
    public async Task EvaluateBigObject()
    {
        using (var browser = await PuppeteerSharp.Puppeteer.LaunchAsync(TestConstants.DefaultBrowserOptions(), TestConstants.ChromiumRevision))
        {
            var window = await page.EvaluateExpressionAsync("window"); // this throws
        }
    }
}
```

I suggest we delete the overloads that return `dynamic`.

whenever we don't care about the result we should use `EvaluateExpressionHandleAsync` or `EvaluateFunctionHandleAsync`

Meir017:
closes #166


lib/PuppeteerSharp/Page.cs
```diff
@@ -156,13 +156,13 @@ private Page(Session client, Target target, FrameTree frameTree, bool ignoreHTTP
         public async Task<IEnumerable<ElementHandle>> GetElementsAsync(string selector)
             => await MainFrame.GetElementsAsync(selector);
 
-        public async Task<JSHandle> EvaluateExpressionHandle(string script)
+        public async Task<JSHandle> EvaluateExpressionHandleAsync(string script)
```
kblok: Now you'll have to add XML Docs to the changed methods :joy:

lib/PuppeteerSharp.Tests/Frame/EvaluateTests.cs
```diff
@@ -17,8 +17,8 @@ public class EvaluateTests : PuppeteerPageBaseTest
             var frame1 = Page.Frames.ElementAt(0);
             var frame2 = Page.Frames.ElementAt(1);
 
-            await frame1.EvaluateExpressionAsync("window.FOO = 'foo'");
-            await frame2.EvaluateExpressionAsync("window.FOO = 'bar'");
+            await frame1.EvaluateExpressionHandleAsync("window.FOO = 'foo'");
+            await frame2.EvaluateExpressionHandleAsync("window.FOO = 'bar'");
```
kblok: One thing I noticed across this PR (I won't mention this many times only here). Is that we're switching from `EvaluateExpression` to `EvaluateExpressionHandle` and that's not equivalent because the first one is disposing the handle as soon as possible.

kblok: Just a thought. How about if we get rid of `EvaluateExpressionHandleAsync` and we treat `EvaluateExpressionAsync<JSHandle>` in a special way?

Meir017: puppeteer has both a `evaluate` and `evaluateHandle` functions, and I prefer not to add **magic** to the function

kblok: @Meir017  👆 the first comment

Meir017:
I think I found a solution - https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/lib/ExecutionContext.js#L42
```js
const result = await handle.jsonValue().catch(error => undefined);
```
we should change the line
https://github.com/kblok/puppeteer-sharp/blob/master/lib/PuppeteerSharp/ExecutionContext.cs#L33
```cs
var result = await handle.JsonValue<T>();
```
and make it catch the exception and return `null`

kblok:
Make sense, we should also add the `EvaluateBigObject` test here.

Meir017:
```
PuppeteerSharp.Tests.Puppeteer.PuppeteerLaunchTests.UserDataDirOption [FAIL]
System.Net.WebSockets.WebSocketException : The 'System.Net.WebSockets.InternalClientWebSocket' instance cannot be used for communication because it has been transitioned into the 'Aborted' state.
```

kblok:
Conflicted @Meir017

PR #169, opened 2018-04-23
Implemented Network Events tests

### LINKED ISSUE #39: Implement Network Events page
https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/test/test.js#L2519

Meir017:
closes #39


lib/PuppeteerSharp/MultiMap.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace PuppeteerSharp
+{
+    internal class MultiMap<TKey, TValue>
```
kblok: :boom:

lib/PuppeteerSharp.Tests/Network/NetworkEventTests.cs
```diff
@@ -0,0 +1,152 @@
...
+            await Page.SetRequestInterceptionAsync(true);
+            Page.RequestCreated += async (sender, e) =>
+            {
+                if (e.Request.Url.EndsWith("css"))
+                    await e.Request.Abort();
+                else
+                    await e.Request.Continue();
+            };
+            var failedRequests = new List<Request>();
+            Page.RequestFailed += (sender, e) => failedRequests.Add(e.Request);
```
Meir017: RequestFailed is not being raised 😭 in `NetworkManager.OnLoadingFailed` the `_requestIdToRequest` does not contain the requestId

Meir017: @kblok this is the last thing

lib/PuppeteerSharp.Tests/Network/NetworkEventTests.cs
```diff
@@ -0,0 +1,177 @@
...
+        }
+
+        [Fact]
+        public async Task PageEventsResponseShouldProvideBody()
+        {
+            Response response = null;
+            Page.ResponseCreated += (sender, e) => response = e.Response;
+            await Page.GoToAsync(TestConstants.ServerUrl + "/simple.json");
+            Assert.NotNull(response);
+            Assert.Equal($"{{\"foo\": \"bar\"}}{Environment.NewLine}", await response.TextAsync());
```
Meir017: fails:
```ini
Assert.Equal() Failure
                        ↓ (pos 14)
Expected: {"foo": "bar"}\r\n
Actual:   {"foo": "bar"}\n
                        ↑ (pos 14)
```
not sure how to work around this, works locally (windows 10)

kblok: I had the same issue in other test. I ended up using `\n` instead of NewLine. I'll work on Mac

Meir017: on windows it would fail:
```cs
Assert.Equal("{\"foo\": \"bar\"}\n", await response.TextAsync());
```

```ini
Result Message:	
Assert.Equal() Failure
                        ↓ (pos 14)
Expected: {"foo": "bar"}\n
Actual:   {"foo": "bar"}\r\n
                        ↑ (pos 14)


```

kblok: I'm trying to give you a good solution without trying to sound lazy :p  

Could you try instead of hardkoding the json text. open the simple.json file?
So the test would be about "You are sending me the same content as the file I'm requesting".

lib/PuppeteerSharp.Tests/Network/NetworkEventTests.cs
```diff
@@ -0,0 +1,179 @@
...
+        [Fact]
+        public async Task PageEventsRequest()
+        {
+            var requests = new List<Request>();
+            Page.RequestCreated += (sender, e) => requests.Add(e.Request);
+            await Page.GoToAsync(TestConstants.EmptyPage);
+            Assert.Single(requests);
+            Assert.Equal(TestConstants.EmptyPage, requests[0].Url);
+            Assert.Equal(ResourceType.Document, requests[0].ResourceType);
+            Assert.Equal("GET", requests[0].Method);
```
kblok: I wouldn't like to introduce a dependency to MVC to have the Verb Enum (https://msdn.microsoft.com/en-us/library/system.web.mvc.httpverbs(v=vs.118).aspx)

But it would be great to have an enum for this.

lib/PuppeteerSharp/Request.cs
```diff
@@ -68,42 +62,62 @@ public class Request : Payload
...
+                var requestData = new Dictionary<string, object> { ["interceptionId"] = InterceptionId };
+                if (overrides?.Url != null) requestData["url"] = overrides.Url;
+                if (overrides?.Method != null) requestData["method"] = overrides.Method;
+                if (overrides?.PostData != null) requestData["postData"] = overrides.PostData;
+                if (overrides?.Headers != null) requestData["headers"] = overrides.Headers;
+
+                await _client.SendAsync("Network.continueInterceptedRequest", requestData);
             }
-            catch (Exception)
+            catch (Exception ex)
```
kblok: Don't you get an `unused ex variable` warning here?

lib/PuppeteerSharp/Request.cs
```diff
@@ -166,31 +180,42 @@ await _client.SendAsync("Network.continueInterceptedRequest", new Dictionary<str
...
             {
-                await _client.SendAsync("Network.continueInterceptedReques", new Dictionary<string, object>
+                await _client.SendAsync("Network.continueInterceptedRequest", new Dictionary<string, object>
                 {
                     {"interceptionId", InterceptionId},
                     {"errorReason", errorReason}
                 });
             }
-            catch (Exception)
+            catch (Exception ex)
```
kblok: Same here

lib/PuppeteerSharp/Response.cs
```diff
@@ -42,7 +42,7 @@ public Response(Session client, Request request, HttpStatusCode status, Dictiona
 
         #region Public Methods
 
-        public Task<string> Buffer()
+        public Task<string> BufferAsync()
```
kblok: XML Docs :)

lib/PuppeteerSharp/Response.cs
```diff
@@ -69,15 +69,10 @@ public Task<string> Buffer()
             return ContentTaskWrapper.Task;
         }
 
+        public Task<string> TextAsync() => BufferAsync();
```
kblok: XML Docs :)

lib/PuppeteerSharp/Response.cs
```diff
@@ -69,15 +69,10 @@ public Task<string> Buffer()
...
 
-        public async Task<string> TextAsync() => await Buffer();
-
-        public async Task<JObject> Json()
-        {
-            var text = await TextAsync();
-            return JObject.Parse(text);
-        }
-
+        public async Task<JObject> JsonAsync() => JObject.Parse(await TextAsync());
```
kblok: XML Docs :)

lib/PuppeteerSharp/Response.cs
```diff
@@ -69,15 +69,10 @@ public Task<string> Buffer()
...
+        public async Task<T> JsonAsync<T>() => (await JsonAsync()).ToObject<T>();
```
kblok: XML Docs :)

lib/PuppeteerSharp.Tests/Network/NetworkEventTests.cs
Meir017: could we use `System.Net.Http.HttpMethod` since we already use `System.Net.HttpStatusCode`?

Meir017: hmmm, after changing it to `System.Net.Http.HttpMethod` the json serializer throw an exception on it since it's not a string.
I ended up creating a simple JsonConverter for the `Method` property

lib/PuppeteerSharp.Tests/Network/NetworkEventTests.cs
```diff
@@ -0,0 +1,179 @@
...
+        [Fact]
+        public async Task PageEventsRequest()
+        {
+            var requests = new List<Request>();
+            Page.RequestCreated += (sender, e) => requests.Add(e.Request);
+            await Page.GoToAsync(TestConstants.EmptyPage);
+            Assert.Single(requests);
+            Assert.Equal(TestConstants.EmptyPage, requests[0].Url);
+            Assert.Equal(ResourceType.Document, requests[0].ResourceType);
+            Assert.Equal(HttpMethod.Get , requests[0].Method);
```
kblok: Once you see the space between `GET` and `,` you can't unsee it :joy:

kblok:
Looking good @Meir017. What's the status?

Meir017:
can't figure out why the test `PageEventsRequestFailed` is failing

kblok:
One letter @Meir017  https://github.com/kblok/puppeteer-sharp/commit/b0bf547f834091899b8e2baf70b3c6bf59cc4bc4 :)

Meir017:
I would never have thought of looking there... 😌

kblok:
@Meir017 Finding this wasn't fun at all https://github.com/kblok/puppeteer-sharp/pull/181/commits/1f31803b1b3b941964b8766187a286934b540c52

This shouldn't be WIP anymore right?

PR #170, opened 2018-04-24
Implement Tracing

### LINKED ISSUE #52: Implement Tracing tests
https://github.com/GoogleChrome/puppeteer/blob/master/test/test.js#L3239

kblok:
closes #52

### DIFF
--- a/lib/PuppeteerSharp.Tests/PuppeteerBaseTest.cs
+++ b/lib/PuppeteerSharp.Tests/PuppeteerBaseTest.cs
@@ -35,7 +35,7 @@ namespace PuppeteerSharp.Tests
             InitializeAsync().GetAwaiter().GetResult();
         }
 
-        public void Dispose() => DisposeAsync().GetAwaiter().GetResult();
+        public virtual void Dispose() => DisposeAsync().GetAwaiter().GetResult();
 
         protected static Task<dynamic> WaitForEvents(Session emitter, string eventName, int eventCount = 1)
         {
--- a/lib/PuppeteerSharp.Tests/Tracing/TracingTests.cs
+++ b/lib/PuppeteerSharp.Tests/Tracing/TracingTests.cs
@@ -1,0 +1,66 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace PuppeteerSharp.Tests.Tracing
+{
+    [Collection("PuppeteerLoaderFixture collection")]
+    public class TracingTests : PuppeteerPageBaseTest
+    {
+        private string _file;
+
+        public TracingTests()
+        {
+            _file = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
+        }
+
+        public override void Dispose()
+        {
+            base.Dispose();
+
+            int attempts = 0;
+            const int attemptTimes = 5;
+
+            while (true)
+            {
+                try
+                {
+                    attempts++;
+                    if (File.Exists(_file))
+                    {
+                        File.Delete(_file);
+                    }
+                    break;
+                }
+                catch (UnauthorizedAccessException)
+                {
+                    if (attempts == attemptTimes)
+                    {
+                        break;
+                    }
+
+                    Task.Delay(1000).GetAwaiter().GetResult();
+                }
+            }
+        }
+
+        [Fact]
+        public async Task ShouldWork()
+        {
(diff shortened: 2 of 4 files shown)


kblok:
Ready to review @Meir017

lib/PuppeteerSharp.Tests/Tracing/TracingTests.cs
```diff
@@ -0,0 +1,111 @@
...
+        {
+            _file = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
+        }
+
+        public override void Dispose()
+        {
+            base.Dispose();
+
+            int attempts = 0;
+            const int attemptTimes = 5;
```
Meir017: maybe rename to `maxAttempts`

lib/PuppeteerSharp.Tests/Tracing/TracingTests.cs
```diff
@@ -0,0 +1,111 @@
...
+                    break;
+                }
+                catch (UnauthorizedAccessException)
+                {
+                    if (attempts == attemptTimes)
+                    {
+                        break;
+                    }
+
+                    Task.Delay(1000).GetAwaiter().GetResult();
```
Meir017: This feels a bit weird, you should override `DisposeAsync` and in it call `await base.DisposeAsync()`

lib/PuppeteerSharp.Tests/PuppeteerBaseTest.cs
```diff
@@ -35,7 +35,7 @@ public PuppeteerBaseTest()
             InitializeAsync().GetAwaiter().GetResult();
         }
 
-        public void Dispose() => DisposeAsync().GetAwaiter().GetResult();
+        public virtual void Dispose() => DisposeAsync().GetAwaiter().GetResult();
```
Meir017: this method shouldn't be virtual, for that we have `DisposeAsync`

lib/PuppeteerSharp/Session.cs
```diff
@@ -133,6 +136,13 @@ internal void OnMessage(string message)
                     }
                 }
             }
+            else if (obj.method == "Tracing.tracingComplete")
+            {
+                TracingComplete?.Invoke(this, new TracingCompleteEventArgs
+                {
+                    Stream = objAsJObject["params"]["stream"].ToString()
```
Meir017: I think you can use `objAsJObject.Value<string>("params.stream")`

lib/PuppeteerSharp/Tracing.cs
```diff
@@ -1,12 +1,117 @@
...
+        /// <summary>
+        /// Starts tracing.
+        /// </summary>
+        /// <returns>Start task</returns>
+        /// <param name="options">Tracing options</param>
+        public async Task StartAsync(TracingOptions options)
+        {
+            if (_recording)
+            {
+                Console.WriteLine("Cannot start recording trace while already recording trace.");
```
Meir017: I'm pretty sure that `console.assert` throws an error and not just logs, so I think we should throw an `InvalidOperationException`

lib/PuppeteerSharp/TracingCompleteEventArgs.cs
```diff
@@ -0,0 +1,8 @@
+﻿using System.IO;
+namespace PuppeteerSharp
+{
+    public class TracingCompleteEventArgs
```
Meir017: should implement `EventArgs`

lib/PuppeteerSharp/Tracing.cs
```diff
@@ -1,12 +1,117 @@
...
+                categories.Add("disabled-by-default-devtools.screenshot");
+            }
+
+            _path = options.Path;
+            _recording = true;
+
+            await _client.SendAsync("Tracing.start", new
+            {
+                transferMode = "ReturnAsStream",
+                categories = string.Join(", ", categories.ToArray())
```
Meir017: remove the `.ToArray()` it's not needed

lib/PuppeteerSharp/Tracing.cs
kblok: That's event clear to me, but yeah, let's treat that as a validation.

lib/PuppeteerSharp/Session.cs
kblok: I wasn't able to do that. This is cleaner though `objAsJObject["params"].Value<string>("stream")`

kblok:
Ready to review @Meir017

PR #178, opened 2018-04-27
Don't await browse.close task

### LINKED ISSUE #176: Connected browsers hangs if created on using blocks
```cs 
var options = new ConnectOptions()
{
    BrowserWSEndpoint = $"wss://chrome.browserless.io?token={apikey}"
};

using(var browser = await Puppeteer.ConnectAsync(options))
using(var page = await browser.NewPageAsync())
{
    await page.GoToAsync(url);
    var stream = await page.ScreenshotStreamAsync(new ScreenshotOptions
    {
        FullPage = true
    });
}
```

It never gets out of the using block.
This was implement in an Azure Function.

kblok:
We have a similar issue we had on #126. The GetAwaiter was hanging on some platforms (like a console app). For instance, the PuppeteerConnectTests.ShouldBeAbleToReconnectToADisconnectedBrowser (https://github.com/kblok/puppeteer-sharp/blob/9bebc1f8a305af365b99d18eedb135e68139104c/lib/PuppeteerSharp.Tests/Puppeteer/PuppeteerConnectTests.cs#L32) hangs if run on a console app.

The solution is quite similar to #126. As we don't need to wait for `Browse.close`.  We just trigger and forget it.

`_closeCallBack` might have a process kill (when we create a process) we would need to await, or a `Browser.close` (when we call connect) we can just trigger and forget.

closes #176

### DIFF
--- a/lib/PuppeteerSharp/Browser.cs
+++ b/lib/PuppeteerSharp/Browser.cs
@@ -108,7 +108,13 @@ namespace PuppeteerSharp
 
             IsClosed = true;
 
-            await _closeCallBack();
+            var closeTask = _closeCallBack();
+
+            if (closeTask != null)
+            {
+                await closeTask;
+            }
+
             Disconnect();
             Closed?.Invoke(this, new EventArgs());
         }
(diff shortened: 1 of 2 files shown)


kblok:
Ready to review @Meir017

PR #180, opened 2018-04-28
Implemented input tests

### LINKED ISSUE #34: Implement input tests
https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/test/test.js#L2011

Meir017:
closes #34


kblok:
Don't forget to add the `Async` suffix in async methods (Mouse.Down, Mouse.Up, etc etc)

lib/PuppeteerSharp/ClickOptions.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace PuppeteerSharp
+{
+    public class ClickOptions
+    {
+        public int? Delay { get; set; }
+        public int ClickCount { get; set; } = 1;
+        public string Button { get; set; } = "left";
```
kblok: If we can have an enum of buttons it would be great.

kblok:
Also don't forget about XML Docs :)

lib/PuppeteerSharp/Input/DownOptions.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace PuppeteerSharp.Input
+{
+
```
kblok: Nit extra line

lib/PuppeteerSharp/Input/keyDefinitions.cs
```diff
@@ -0,0 +1,1597 @@
+using System.Collections.Generic;
```
kblok: Filename should be `KeyDefinitions` capital K

lib/PuppeteerSharp/Page.cs
```diff
@@ -141,12 +141,12 @@ private Page(Session client, Target target, FrameTree frameTree, bool ignoreHTTP
         public async Task TapAsync(string selector)
         {
             var handle = await GetElementAsync(selector);
-
-            if (handle != null)
+            if (handle == null)
             {
-                await handle.TapAsync();
-                await handle.DisposeAsync();
+                throw new PuppeteerException($"No node found for selector: {selector}");
```
kblok: How about throwing a SelectorException?

Meir017:
@kblok ready except for 2 failing tests:
- PuppeteerSharp.Tests.Input.InputTests.ShouldSpecifyRepeatProperty
- uppeteerSharp.Tests.Input.InputTests.ShouldClickTheButtonAfterNavigation

lib/PuppeteerSharp.Tests/Input/InputTests.cs
```diff
@@ -0,0 +1,539 @@
...
+            Assert.False(await Page.EvaluateExpressionAsync<bool>("result.check"));
+        }
+
+        [Fact]
+        public async Task ShouldClickOnCheckboxLabelAndToggle()
+        {
+            await Page.GoToAsync(TestConstants.ServerUrl + "/input/checkbox.html");
+            Assert.Null(await Page.EvaluateExpressionAsync("result.check"));
+            await Page.ClickAsync("label[for=\"agree\"]");
+            Assert.Equal(new[] {
```
kblok: Missing
`expect(await page.evaluate(() => result.check)).toBe(true);`

lib/PuppeteerSharp.Tests/Input/InputTests.cs
```diff
@@ -0,0 +1,539 @@
...
+            }", input));
+        }
+
+        [Fact]
+        public async Task ShouldMoveWithTheArrowKeys()
+        {
+            await Page.GoToAsync(TestConstants.ServerUrl + "/input/textarea.html");
+            await Page.TypeAsync("textarea", "Hello World!");
+            Assert.Equal("Hello World!", await Page.EvaluateExpressionAsync<string>("document.querySelector('textarea').value"));
+            for (var i = 0; i < "World!".Length; i++)
```
kblok: Let's always use brackets (lets review this across the file)

lib/PuppeteerSharp/Input/KeyDefinitions.cs
```diff
@@ -0,0 +1,1595 @@
...
+                Key = "8",
+                Code = "Digit8"
+            },
+            ["9"] = new KeyDefinition
+            {
+                KeyCode = 57,
+                Key = "9",
+                Code = "Digit9"
+            },
+            ["Power"] = new KeyDefinition
```
kblok: I'm wondering how would the user know what are our available keys...
Maybe it would be great if we could use an enum or a string.
We can create another issue if you think it would get out of scope.

lib/PuppeteerSharp/Input/Keyboard.cs
```diff
@@ -203,5 +166,50 @@ private int modifierBit(string key)
...
+                Key = string.Empty,
+                KeyCode = 0,
+                Code = string.Empty,
+                Text = string.Empty,
+                Location = 0
+            };
+
+            var definition = KeyDefinitions.Get(keyString);
+
+            if (!string.IsNullOrEmpty(definition.Key))
```
kblok: Review brackets

lib/PuppeteerSharp/Input/KeyDefinitions.cs
Meir017: maybe we could add something like this - https://github.com/SeleniumHQ/selenium/blob/master/dotnet/src/webdriver/Keys.cs?

kblok: Yes that could work

Meir017: hmm I noticed that there isn't a clear way to describe some of the properties.
- we can't use the key of each `KeyDefinition` for it's property's name since some of the are invalid names (digits, empty...)
- `KeyDefinition.Code` isn't unique

Meir017: would a simple class that has an entry for each key in https://github.com/kblok/puppeteer-sharp/pull/180/files#diff-7e6afb5424df8f9e8bee32a50727f245L12 be good enough?
I was thinking of creating an overload for that class
```cs
public static class KeyboardExtensions
{
  public static Task TypeAsync(this Keyboard keyboard, Key key, TypeOptions options = null)
    => keyboard.TypeAsync(key.ToString(), options);
}

public class Key
{
  public string Value { get; }
  private Key(string value) => Value = value;

  public static readonly Keys Digit0 = new Keys("0");
  public static readonly Keys Digit1 = new Keys("1");
  public static readonly Keys Digit2 = new Keys("2");
  public static readonly Keys Digit3 = new Keys("3");

  .
  .
  .

  public override string ToString() => Value;
}
```

kblok: Agreed on the Key. I don't see why it should be an extension method rather than an overload though.

lib/PuppeteerSharp/Page.cs
```diff
@@ -138,15 +141,22 @@ private Page(Session client, Target target, FrameTree frameTree, bool ignoreHTTP
             return BuildMetricsObject(response.Metrics);
         }
 
+        /// <summary>
+        /// Fetches an element with <paramref name="selector"/>, scrolls it into view if needed, and then uses <see cref="Touchscreen"/> to tap in the center of the element.
+        /// </summary>
+        /// <param name="selector">A selector to search for element to tap. If there are multiple elements satisfying the selector, the first will be clicked.</param>
+        /// <param name="options"></param>
```
kblok: There is no options param

lib/PuppeteerSharp.Tests/Input/InputTests.cs
```diff
@@ -0,0 +1,554 @@
...
+            {
+                Page.Keyboard.PressAsync("ArrowLeft");
```
kblok: I'm getting several warnings because these guys are not being awaited. 
How about assigning the task to a dummy variable?

Meir017: Would it be okay to define a task field on the class and use it everywhere?

kblok: Absolutely

kblok:
@Meir017 take look at this
https://github.com/kblok/puppeteer-sharp/commit/c9b5f2db83fc0b1a64e774892daea5ec15c4ea88
and this
https://github.com/kblok/puppeteer-sharp/commit/3478ba497ab40c83939c078c4ace53e9279e2b1d

Meir017:
I think another flaky test caused the fail - https://ci.appveyor.com/project/kblok/puppeteer-sharp/build/1.0.563#L630

lib/PuppeteerSharp/Input/Key.cs
```diff
@@ -0,0 +1,112 @@
+﻿namespace PuppeteerSharp.Input
+{
+    public class Key
```
kblok: I don't find the usage of this class @Meir017.

Meir017: it's a utility class - https://github.com/kblok/puppeteer-sharp/pull/180#discussion_r185612378

PR #186, opened 2018-05-02
Implement PageError

### LINKED ISSUE #182: Implement Page.Events.PageError
https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/test/test.js#L1613

In #117 we implemented another test.

kblok:
Closes #182


kblok:
Ready to review @Meir017

lib/PuppeteerSharp/PageErrorException.cs
```diff
@@ -0,0 +1,22 @@
+﻿using Newtonsoft.Json;
+
+namespace PuppeteerSharp
+{
+    public class PageErrorException
```
Meir017: this name is confusing since it doesn't inherit from `Exception`

lib/PuppeteerSharp/Page.cs
```diff
@@ -848,8 +852,9 @@ private void EmitMetrics(PerformanceMetricsResponse metrics)
             }
         }
 
-        private void HandleException(dynamic exceptionDetails)
+        private void HandleException(PageError pageError)
         {
+            PageError?.Invoke(this, new PageErrorEventArgs(pageError));
```
Meir017: I noticed that puppeteer calls `helper.getExceptionMessage(exceptionDetails)` and then raises the event with a simple exception that only contains a message. not the complete `exceptionDetails`

kblok: Agreed, I went too far there.
I'll implement a RuntimeExceptionThrownResponse with only the data we need there, but I would still need a class called "Exception"...

kblok: `ExceptionData`?

Meir017: maybe just a `Message` property?

Meir017: can't we just stick it on `PageErrorEventArgs`?

PR #199, opened 2018-05-06
Implement Page.SelectAsync

### LINKED ISSUE #51: Implement Page.Select tests
https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/test/test.js#L3044

kblok:
closes #51

### DIFF
--- a/lib/PuppeteerSharp/Frame.cs
+++ b/lib/PuppeteerSharp/Frame.cs
@@ -128,7 +128,7 @@ namespace PuppeteerSharp
             var value = await document.GetElementAsync(selector);
             return value;
         }
-        
+
         internal Task<object> EvalMany(string selector, Func<object> pageFunction, object[] args)
         {
             throw new NotImplementedException();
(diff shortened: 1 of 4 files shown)


kblok:
Ready to review @Meir017

Meir017:
looks like you're missing the test
* `should deselect all options when passed no values for a select without multiple`

lib/PuppeteerSharp/Frame.cs
```diff
@@ -315,6 +316,22 @@ internal Task<JSHandle> WaitForFunctionAsync(string script, WaitForFunctionOptio
             return handle.AsElement();
         }
 
+        internal async Task<string[]> SelectAsync(string selector, object[] args)
```
Meir017: make `params`?

lib/PuppeteerSharp/Frame.cs
```diff
@@ -315,6 +316,22 @@ internal Task<JSHandle> WaitForFunctionAsync(string script, WaitForFunctionOptio
...
+        {
+            return await QuerySelectorAsync(selector).EvaluateFunctionAsync<string[]>(@"(element, values) => {
```
Meir017: arrow function? (not completely sure about this one)

lib/PuppeteerSharp/Frame.cs
Meir017: redundant `async` and `await` keywords. just return ` QuerySelectorAsync(...`

Meir017: in puppeteer the `args` is called `values` and if of type `string[]` not `object[]`

kblok: I didn't know that `params string[]` was valid, that's cool.

Meir017: params works with `T[]` 😄

kblok:
Done @Meir017

PR #229, opened 2018-05-15
Implement missing Launch Tests

### LINKED ISSUE #76: Implement puppeteer.launch tests using the evaluate function

kblok:
closes #76

### DIFF
--- a/lib/PuppeteerSharp.Tests/PuppeteerTests/PuppeteerLaunchTests.cs
+++ b/lib/PuppeteerSharp.Tests/PuppeteerTests/PuppeteerLaunchTests.cs
@@ -85,12 +85,12 @@ namespace PuppeteerSharp.Tests.PuppeteerTests
                 TestConstants.ChromiumRevision))
             using (var page = await browser.NewPageAsync())
             {
-				var neverResolves = page.EvaluateFunctionHandleAsync("() => new Promise(r => {})");
-				await browser.CloseAsync();
-				var exception = await Assert.ThrowsAsync<TargetClosedException>(() => neverResolves);
+                var neverResolves = page.EvaluateFunctionHandleAsync("() => new Promise(r => {})");
+                await browser.CloseAsync();
+                var exception = await Assert.ThrowsAsync<TargetClosedException>(() => neverResolves);
                 Assert.Contains("Protocol error", exception.Message);
-			}
-		}
+            }
+        }
 
         [Fact]
         public async Task ShouldRejectIfExecutablePathIsInvalid()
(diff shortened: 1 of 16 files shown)


lib/PuppeteerSharp.Tests/PuppeteerTests/PuppeteerLaunchTests.cs
```diff
@@ -77,22 +77,20 @@ public class PuppeteerLaunchTests : PuppeteerBaseTest
...
                 TestConstants.DefaultBrowserOptions(),
                 TestConstants.ChromiumRevision))
             using (var page = await browser.NewPageAsync())
             {
-                var neverResolves = page.EvaluateFunctionHandleAsync("() => new Promise(r => {})");
-                await browser.CloseAsync();
-
-                await neverResolves;
-                var exception = await Assert.ThrowsAsync<Exception>(() => neverResolves);
+				var neverResolves = page.EvaluateFunctionHandleAsync("() => new Promise(r => {})");
```
kblok: @Meir017 just if you want you to jump in and give me a hand in here.
`neverResolves` is not being canceled when we run all the tests altogether. But it is being resolved when I debug it. That task gets stuck in `WaitingForActivation`. It's diving me crazy.

kblok: The `callback.TaskWrapper.SetException` on Session.OnClosed is being called but that exception never gets to the test. It's like someone is cutting the wire in the middle.

kblok:
Ready to review @Meir017. I just needed a rubber duck, this was a tough one.

lib/PuppeteerSharp.Tests/PuppeteerTests/PuppeteerLaunchTests.cs
```diff
@@ -143,6 +141,85 @@ public class PuppeteerLaunchTests : PuppeteerBaseTest
             }
         }
 
+        [Fact]
+        public async Task UserDataDirOptionShouldRestoreState()
+        {
+            var launcher = new Launcher();
```
Meir017: this class inherits from `PuppeteerBaseTest` that creates a `Browser` for each test. maybe it should inherit from a more basic class that doesn't do anything since the created `Browser` isn't used?
I was thinking of adding a `PuppeteerBrowserBaseTest` that will work the same way `PuppeteerBaseTest` works now, and then refactor `PuppeteerBaseTest` to be more basic

lib/PuppeteerSharp/Connection.cs
```diff
@@ -59,18 +61,22 @@ public Connection(string url, int delay, ClientWebSocket ws)
...
             var buffer = new ArraySegment<Byte>(encoded, 0, encoded.Length);
-            QueueId(id);
-
             await _socketQueue.Enqueue(() => WebSocket.SendAsync(buffer, WebSocketMessageType.Text, true, default(CancellationToken)));
 
-            return await _responses[id].Task;
-        }
+            if (method == CloseMessage)
+            {
+                _stopReading = true;
```
Meir017: maybe call `StopReading();` here?

kblok:
Done @Meir017

PR #233, opened 2018-05-21
AppVeyor environments

kblok:


kblok:
@Meir017 though this doesn't buy us parallel builds it's cleaner on the AppVeyor side.

PR #234, opened 2018-05-22
Implement JSCoverage

### LINKED ISSUE #57: Implement JSCoverage tests
https://github.com/GoogleChrome/puppeteer/blob/master/test/test.js#L3650

kblok:
Closes #57

### DIFF
--- a/lib/PuppeteerSharp/CoverageEntry.cs
+++ b/lib/PuppeteerSharp/CoverageEntry.cs
@@ -2,6 +2,9 @@
 
 namespace PuppeteerSharp
 {
+    /// <summary>
+    /// Coverage report for all non-anonymous scripts.
+    /// </summary>
     public class CoverageEntry
     {
         /// <summary>
--- a/lib/PuppeteerSharp/CoverageEntry.cs
+++ b/lib/PuppeteerSharp/CoverageEntry.cs
@@ -4,8 +4,20 @@ namespace PuppeteerSharp
 {
     public class CoverageEntry
     {
+        /// <summary>
+        /// Script URL
+        /// </summary>
+        /// <value>Script URL.</value>
         public string Url { get; set; }
+        /// <summary>
+        /// Script ranges that were executed. Ranges are sorted and non-overlapping.
+        /// </summary>
+        /// <value>Ranges.</value>
         public CoverageEntryRange[] Ranges { get; set; }
-        public string Text { get; internal set; }
+        /// <summary>
+        /// Script content
+        /// </summary>
+        /// <value>Script content.</value>
+        public string Text { get; set; }
     }
 }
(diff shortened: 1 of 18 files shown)


kblok:
Ready to review @Meir017

lib/PuppeteerSharp/CoverageEntry.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Collections.Generic;
+
+namespace PuppeteerSharp
+{
+    public class CoverageEntry
```
Meir017: docs?

lib/PuppeteerSharp/JSCoverage.cs
```diff
@@ -1,7 +1,9 @@
 ﻿using PuppeteerSharp.Messaging;
 using System;
+using System.Linq;
 using System.Collections.Generic;
 using System.Threading.Tasks;
+using static PuppeteerSharp.Messaging.ProfilerTakePreciseCoverageResponse;
```
Meir017: is this needed?

lib/PuppeteerSharp/JSCoverage.cs
```diff
@@ -63,7 +65,91 @@ public JSCoverage(Session client)
...
+            {
+                _scriptURLs.TryGetValue(entry.ScriptId, out var url);
+                _scriptSources.TryGetValue(entry.ScriptId, out var text);
+
+                if (text == null || url == null)
+                {
+                    continue;
+                }
+
+                var flattenRanges = new List<ProfilerTakePreciseCoverageResponseRange>();
```
Meir017: we could use a `SelectMany` here instead of `foreach` & `AddRange`
```cs
var flattenRanges = entry.Functions
    .SelectMany(func => func.Ranges)
    .ToList();
```

lib/PuppeteerSharp/JSCoverage.cs
```diff
@@ -63,7 +65,91 @@ public JSCoverage(Session client)
            );
             _client.MessageReceived -= client_MessageReceived;
 
-            // TODO: return coverage;            
+            var coverage = new List<CoverageEntry>();
+            foreach (var entry in profileResponseTask.Result.Result)
+            {
+                _scriptURLs.TryGetValue(entry.ScriptId, out var url);
```
Meir017: we could change this to
```cs
if(!_scriptURLs.TryGetValue(entry.ScriptId, out var url) || !_scriptSources.TryGetValue(entry.ScriptId, out var text))
{
    continue;
}
```

lib/PuppeteerSharp/JSCoverage.cs
kblok: Yes because `ProfilerTakePreciseCoverageResponseRange` is a sub-class. I thought that making a new file for `ProfilerTakePreciseCoverageResponseRange` was "too much". What do you think?

kblok:
How about now @Meir017 ?

PR #258, opened 2018-05-24
page docs

### LINKED ISSUE #236: Page Class XML Documentation

Meir017:
closes #236

### DIFF
--- a/lib/PuppeteerSharp/Page.cs
+++ b/lib/PuppeteerSharp/Page.cs
@@ -68,9 +68,9 @@ namespace PuppeteerSharp
             _frameManager.FrameDetached += (sender, e) => FrameDetached?.Invoke(this, e);
             _frameManager.FrameNavigated += (sender, e) => FrameNavigated?.Invoke(this, e);
 
-            _networkManager.RequestCreated += (sender, e) => Request?.Invoke(this, e);
+            _networkManager.Request += (sender, e) => Request?.Invoke(this, e);
             _networkManager.RequestFailed += (sender, e) => RequestFailed?.Invoke(this, e);
-            _networkManager.ResponseCreated += (sender, e) => Response?.Invoke(this, e);
+            _networkManager.Response += (sender, e) => Response?.Invoke(this, e);
             _networkManager.RequestFinished += (sender, e) => RequestFinished?.Invoke(this, e);
 
             Client.MessageReceived += client_MessageReceived;
@@ -618,7 +618,7 @@ namespace PuppeteerSharp
                 }
             };
 
-            _networkManager.RequestCreated += createRequestEventListener;
+            _networkManager.Request += createRequestEventListener;
 
             var mainFrame = _frameManager.MainFrame;
             var timeout = options?.Timeout ?? DefaultNavigationTimeout;
@@ -651,7 +651,7 @@ namespace PuppeteerSharp
             }
 
             watcher.Cancel();
-            _networkManager.RequestCreated -= createRequestEventListener;
+            _networkManager.Request -= createRequestEventListener;
 
             if (exception != null)
             {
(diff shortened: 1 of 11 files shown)


lib/PuppeteerSharp/Page.cs
```diff
@@ -63,28 +76,97 @@ private Page(Session client, Target target, FrameTree frameTree, bool ignoreHTTP
...
+        /// <summary>
+        /// Emitted when a <see cref="Response"/> is received.
+        /// </summary>
         public event EventHandler<ResponseCreatedEventArgs> ResponseCreated;
+
+        /// <summary>
+        /// Emitted when a page issues a request. The <see cref="Request"/> object is read-only.
+        /// In order to intercept and mutate requests, see <see cref="SetRequestInterceptionAsync(bool)"/>
+        /// </summary>
         public event EventHandler<RequestEventArgs> RequestCreated;
```
Meir017: I think that this should be rename to `Request` to match puppeteer

lib/PuppeteerSharp/Page.cs
```diff
@@ -63,28 +76,97 @@ private Page(Session client, Target target, FrameTree frameTree, bool ignoreHTTP
...
+
+        /// <summary>
+        /// Emitted when a frame is navigated to a new url.
+        /// </summary>
         public event EventHandler<FrameEventArgs> FrameNavigated;
 
+        /// <summary>
+        /// Emitted when a <see cref="Response"/> is received.
+        /// </summary>
         public event EventHandler<ResponseCreatedEventArgs> ResponseCreated;
```
Meir017: I think that this should be rename to `Response` to match puppeteer

lib/PuppeteerSharp/Page.cs
```diff
@@ -63,28 +76,97 @@ private Page(Session client, Target target, FrameTree frameTree, bool ignoreHTTP
...
+
+        /// <summary>
+        /// Emitted when the page crashes
+        /// </summary>
         public event EventHandler<ErrorEventArgs> Error;
+
+        /// <summary>
+        /// Emitted when the JavaScript code makes a call to <c>console.timeStamp</c>. For the list of metrics see <see cref="Page.MetricsAsync"/>.
+        /// </summary>
         public event EventHandler<MetricEventArgs> MetricsReceived;
```
Meir017: I think that this should be renamed to `Metrics` to match puppeteer

lib/PuppeteerSharp/Page.cs
kblok: Though I don't like the idea, I think we had this conversation about being strong matching puppeteer. Let's do it

lib/PuppeteerSharp/Page.cs
kblok: Same

lib/PuppeteerSharp/Page.cs
kblok: Same

lib/PuppeteerSharp/Page.cs
```diff
@@ -55,54 +68,159 @@ private Page(Session client, Target target, FrameTree frameTree, bool ignoreHTTP
...
 
             Client.MessageReceived += client_MessageReceived;
         }
 
+        internal Session Client { get; }
+
         #region Public Properties
+
+        /// <summary>
+        /// Emitted when the JavaScript <c>load</c> <see cref="https://developer.mozilla.org/en-US/docs/Web/Events/load"/> event is dispatched.
```
kblok: In C# I think the proper word should be `raised` rather than `emitted`

lib/PuppeteerSharp/Page.cs
```diff
@@ -126,6 +244,13 @@ private Page(Session client, Target target, FrameTree frameTree, bool ignoreHTTP
 
         #region Public Methods
 
+        /// <summary>
+        /// Returns metrics
+        /// </summary>
+        /// <returns>metrics</returns>
```
kblok: I think it would be good to follow the wording "Task which resolves to" a list of metrics" (across the class)

PR #261, opened 2018-05-28
New ScreenshotDataAsync and PdfDataAsync methods

### LINKED ISSUE #248: PerformScreenshot forces data to be copied
The internal method `PerformScreenshot` returns `Task<Stream>`. The `Stream` is in fact a `MemoryStream`, constructed from a `byte[]`.

If we want to get the data as a `byte[]` we have to read it out of the stream into another copy, adding to the load on the LOH.

The `MemoryStream` constructor stores the `byte[]` directly, and it does has a `GetBuffer` method, but the constructor overload used in this case sets an internal flag so that `GetBuffer` throws an exception.

Workaround: use reflection to sniff for the `byte[]`!

It would be great if there were overloads that return the `byte[]` without wrapping it.

kblok:
Closes #248


kblok:
Ready to review @Meir017

PR #286, opened 2018-05-30
Fix demo build status badge in README

kenny-evitt:
I'm not sure these changes are correct, but as-is the two AppVeyor 'build status' badges in the README are pointing to the same (AppVeyor) project. Based on the markup in the README file itself, I'm guessing one of the badges was meant to reference the build status of the demo project.

But, even with these changes, there's no visual distinction between the two build status badges. If you have an idea how that should be affected, I'll take a stab at it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
 # Puppeteer Sharp
 
 [![NuGet](https://img.shields.io/nuget/v/PuppeteerSharp.svg?style=flat-square&label=nuget&colorB=green)][NugetUrl]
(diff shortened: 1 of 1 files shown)


kblok:
Good catch @kenny-evitt and thank you for stepping in :)

Check this out https://www.appveyor.com/blog/2014/11/26/customizing-text-of-build-status-badges/

We could call one "master" and the other "demo"

kenny-evitt:
@kblok Thanks for the project!

I added the custom text to the badges just now.

kblok:
Your welcome @kenny-evitt! Thank you for jumping in!

kenny-evitt:
@kblok Need anything else for this before merging?

PR #290, opened 2018-05-31
Fix PDF README example

kenny-evitt:
This remove an extraneous `)` in the PDF example in the README.


kblok:
Thank you @kenny-evitt

PR #292, opened 2018-06-01
CDPSession docs

### LINKED ISSUE #266: Session Class XML Documentation

kblok:
closes #266


kblok:
Take another look @Meir017 I reduce the scope of many methods and constructors.

I think that another good exercise in these docs task is to check what we need to leave public and what shouldn’t be.

kblok:
Done @Meir017

PR #300, opened 2018-06-02
JSHandle docs

### LINKED ISSUE #240: JSHandle Class XML Documentation

kblok:
closes #240

### DIFF
--- a/lib/PuppeteerSharp.Tests/PuppeteerSharp.Tests.csproj
+++ b/lib/PuppeteerSharp.Tests/PuppeteerSharp.Tests.csproj
@@ -2,6 +2,7 @@
   <PropertyGroup>
     <TargetFrameworks>netcoreapp2.0;net471</TargetFrameworks>
     <IsPackable>false</IsPackable>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.CSharp" Version="4.5.0-preview1-26216-02" />
@@ -17,6 +18,10 @@
   </ItemGroup>
   <PropertyGroup Condition="'$(TargetFramework)'=='net471'">
     <RuntimeIdentifier>win7-x64</RuntimeIdentifier>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
   </PropertyGroup>
   <ItemGroup>
     <ProjectReference Include="..\PuppeteerSharp.TestServer\PuppeteerSharp.TestServer.csproj" />
--- a/lib/PuppeteerSharp.Tests/PuppeteerSharp.Tests.csproj
+++ b/lib/PuppeteerSharp.Tests/PuppeteerSharp.Tests.csproj
@@ -2,7 +2,6 @@
   <PropertyGroup>
     <TargetFrameworks>netcoreapp2.0;net471</TargetFrameworks>
     <IsPackable>false</IsPackable>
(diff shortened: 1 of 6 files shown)


lib/PuppeteerSharp/JSHandle.cs
```diff
@@ -7,23 +7,46 @@
...
+            Client = client;
+            Logger = Client.Connection.LoggerFactory.CreateLogger(this.GetType());
             RemoteObject = remoteObject;
         }
 
+        /// <summary>
+        /// Gets the execution context.
+        /// </summary>
+        /// <value>The execution context.</value>
         public ExecutionContext ExecutionContext => _context;
```
Meir017: this can be a get only property

kblok: It is a get only property 🤔

Meir017: we don't need the private field...

kblok: Ah yeah 😛

kblok:
Done @Meir017. Now let's wait 2 days till all builds run 😝

lib/PuppeteerSharp/JSHandle.cs
```diff
@@ -7,23 +7,44 @@
...
+            Logger = Client.Connection.LoggerFactory.CreateLogger(this.GetType());
             RemoteObject = remoteObject;
         }
 
-        public ExecutionContext ExecutionContext => _context;
+        /// <summary>
+        /// Gets the execution context.
+        /// </summary>
+        /// <value>The execution context.</value>
+        public ExecutionContext ExecutionContext { get; private set; }
```
Meir017: I don't think we need the `private set` you can set in the ctor even without it

kblok: cool, I didn't know about that.

kblok: Done @Meir017

lib/PuppeteerSharp.Tests/PuppeteerSharp.Tests.csproj
```diff
@@ -18,6 +19,10 @@
   <PropertyGroup Condition="'$(TargetFramework)'=='net471'">
     <RuntimeIdentifier>win7-x64</RuntimeIdentifier>
   </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
```
Meir017: what are these for?

lib/PuppeteerSharp/JSHandle.cs
```diff
@@ -7,23 +7,44 @@
...
+        /// <summary>
+        /// Gets or sets a value indicating whether this <see cref="PuppeteerSharp.JSHandle"/> is disposed.
+        /// </summary>
+        /// <value><c>true</c> if disposed; otherwise, <c>false</c>.</value>
         public bool Disposed { get; set; }
+        /// <summary>
+        /// Gets or sets the remote object.
+        /// </summary>
+        /// <value>The remote object.</value>
         public dynamic RemoteObject { get; internal set; }
```
Meir017: do we need the `internal set` here?

lib/PuppeteerSharp.Tests/PuppeteerSharp.Tests.csproj
kblok: Damned. I have to revert this. My VS is hanging when I try to debug test. I'm trying to see what's going on. I accidentally committed this.

kblok: Reverted.

lib/PuppeteerSharp/JSHandle.cs
kblok: No, we don't. I removed that and I also change the Disposed setter to private.

PR #308, opened 2018-06-05
Fix Chromium downloader on 32bits apps runnning on a x64 platform

### LINKED ISSUE #294: PuppeteerSharpPdfDemo throws exception
using (var browser = await Puppeteer.LaunchAsync(options, Downloader.DefaultRevision))
throws exception 
An attempt was made to transition a task to a final state when it had already completed

kblok:
`IntPtr.Size == 8` checks if the application is in 64bits. We need to check whether the host OS is 64bits or not.

closes #294


kblok:
@Meir017

PR #328, opened 2018-06-15
Add editorconfig

kblok:
Reference: https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference


kblok:
@Meir017 I invite you to go to the reference site and check this config.  Let me know what you think.

Meir017:
maybe have the file in the root of the repo and then it will also work for the demo projects?

kblok:
Done @Meir017

PR #335, opened 2018-06-18
Closes issue #333 by handling paths with spaces in them.

### LINKED ISSUE #333: Chrome process launch fails when Windows user name contains 'special' characters.
My name on some Windows installations is embedded into the user directory so that it comes out as 

`C:\Users\Ryan O'Neill\AppData\....`

There are two problems here, the space and the apostrophe. I imagine other characters will cause this issue too. When launching Chrome, the process ignores all of the characters past the space and shows this error (via a popup dialogue, visible when headless is turned off);

> Failed to Create Data Directory
> Chromium cannot read or write to its data directory:
> C:\Users\Ryan

This can be rectified in Launcher.cs by amending;
`                    chromeArguments.Add($"{UserDataDirArgument}={options.UserDataDir}");
`
to
`                    chromeArguments.Add($"{UserDataDirArgument}=\"{options.UserDataDir}\"");
`

To reproduce this error, create a Windows user with a space in their name. A lot of the unit tests then fail.

I am happy to provide a pull-request, but thought I would check first.

RyanONeill1970:
closes #333

### DIFF
--- a/lib/PuppeteerSharp.Tests/PuppeteerTests/PuppeteerLaunchTests.cs
+++ b/lib/PuppeteerSharp.Tests/PuppeteerTests/PuppeteerLaunchTests.cs
@@ -137,7 +137,7 @@ namespace PuppeteerSharp.Tests.PuppeteerTests
             var launcher = new Launcher(TestConstants.LoggerFactory);
             var userDataDir = Launcher.GetTemporaryDirectory();
             var options = TestConstants.DefaultBrowserOptions();
-            options.Args = options.Args.Concat(new[] { $"--user-data-dir={userDataDir}" }).ToArray();
+            options.Args = options.Args.Concat(new[] { $"--user-data-dir=\"{userDataDir}\"" }).ToArray();
 
             using (var browser = await launcher.LaunchAsync(options, TestConstants.ChromiumRevision))
             {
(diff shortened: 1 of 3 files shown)


RyanONeill1970:
I didn't add a test for downloading a full Chromium to a path with a space in it as it seemed due to the download time, but I can add it if you wish.

There are other unrelated tests which are failing on a \r and \r\n comparison in their results.

kblok:
@RyanONeill1970 if tests are green I'm good.

Meir017:
can we use extensions methods like https://github.com/cake-build/cake/blob/develop/src/Cake.Core/Extensions/StringExtensions.cs

RyanONeill1970:
No problem, in the Helpers folder under StringExtensions.cs?

I had never seen an unquote/quote method before I wrote these, funny how close to the Cake code they are.

kblok:
Yeah, go for it @RyanONeill1970

Meir017:
make sense 👍

RyanONeill1970:
Happy to tweak more if it's not right.

Meir017:
can we use the same implementation that cake uses?

RyanONeill1970:
Sure, just bring those two methods in though or the whole file?

Meir017:
I think just those methods

kblok:
Thank you @RyanONeill1970 for your contribution. I hope this is the first of many to come :)

RyanONeill1970:
Thanks, I hope so too. Keep up the good work, it's a really useful tool. I may have spotted some more issues, will run tests and let you know.

kblok:
That's great, we have great unit test coverage, but we need more real-life usage.

PR #369, opened 2018-06-29
Wait for the page to close in the browser in page.close

### LINKED ISSUE #349: Wait for the page to close in the browser in page.close
https://github.com/kblok/puppeteer/pull/2/commits/d75fdb7098732cb9e5888729f77b90b40caf96f1

kblok:
closes #349

### DIFF
--- a/lib/PuppeteerSharp/Page.cs
+++ b/lib/PuppeteerSharp/Page.cs
@@ -1750,11 +1750,7 @@ namespace PuppeteerSharp
 
             string SerializeArgument(object arg)
             {
-                if (arg == null)
-                {
-                    return "undefined";
-                }
-                return JsonConvert.SerializeObject(arg);
+                return arg == null ? "undefined" : JsonConvert.SerializeObject(arg);
             }
         }
         #endregion
--- a/lib/PuppeteerSharp/Page.cs
+++ b/lib/PuppeteerSharp/Page.cs
@@ -981,17 +981,17 @@ namespace PuppeteerSharp
         /// Closes the page.
         /// </summary>
         /// <returns>Task.</returns>
-        public Task CloseAsync()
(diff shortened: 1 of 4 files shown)


lib/PuppeteerSharp/Page.cs
```diff
@@ -981,17 +981,17 @@ public Task EmulateMediaAsync(MediaType media)
         /// Closes the page.
         /// </summary>
         /// <returns>Task.</returns>
-        public Task CloseAsync()
+        public async Task CloseAsync()
         {
             if (!(Client?.Connection?.IsClosed ?? true))
             {
-                return Client.Connection.SendAsync("Target.closeTarget", new
+                await Client.Connection.SendAsync("Target.closeTarget", new
```
Meir017: can't we just use return for both tasks? (and in line 993)

lib/PuppeteerSharp/Page.cs
```diff
@@ -1750,7 +1750,10 @@ private static string EvaluationString(string fun, params object[] args)
 
             string SerializeArgument(object arg)
             {
-                if (arg == null) return "undefined";
+                if (arg == null)
```
Meir017: maybe use a conditional expression here?
```cs
return arg == null ? "undefind" : JsonConvert.SerializeObject(arg);
```

lib/PuppeteerSharp/Page.cs
kblok: I don't think so. Both are in the same if block.
```cs 
await Client.Connection.SendAsync();
await Target.CloseTask;
```

lib/PuppeteerSharp/Page.cs
kblok: Done

PR #372, opened 2018-06-30
add type=module to addScriptTag

### LINKED ISSUE #359: Add type=module to addScriptTag
https://github.com/kblok/puppeteer/pull/2/commits/625c7ebdda2ef0473ee8d593dd30a1d34d3ee54a

Meir017:
closes #359

### DIFF
--- a/lib/PuppeteerSharp/Frame.cs
+++ b/lib/PuppeteerSharp/Frame.cs
@@ -404,7 +404,7 @@ namespace PuppeteerSharp
         /// <returns>Task which resolves to the added tag when the script's onload fires or when the script content was injected into frame</returns>
         /// <seealso cref="Page.AddScriptTagAsync(AddTagOptions)"/>
         /// <seealso cref="Page.AddScriptTagAsync(string)"/>
-        public Task<ElementHandle> AddScriptTag(AddTagOptions options)
+        public async Task<ElementHandle> AddScriptTag(AddTagOptions options)
         {
             const string addScriptUrl = @"async function addScriptUrl(url, type) {
               const script = document.createElement('script');
@@ -439,7 +439,7 @@ namespace PuppeteerSharp
                 var url = options.Url;
                 try
                 {
-                    return AddScriptTagPrivate(addScriptUrl, url, options.Type);
+                    return await AddScriptTagPrivate(addScriptUrl, url, options.Type);
                 }
                 catch (PuppeteerException)
                 {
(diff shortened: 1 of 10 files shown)


Meir017:
these tests are failing
* ShouldWorkWithAPath
* ShouldIncludeSourcemapWhenPathIsProvided
* ShouldWorkWithContent

looks like an issue with the new `type` property on the `AddTagOptions` class

lib/PuppeteerSharp/Frame.cs
```diff
@@ -406,31 +406,40 @@ public Task<string[]> SelectAsync(string selector, params string[] values)
...
               document.head.appendChild(script);
               return script;
             }";
 
+            async Task<ElementHandle> AddScriptTagPrivate(string script, string urlOrContent, string type)
+            {
+                var context = await GetExecutionContextAsync();
+                return string.IsNullOrEmpty(type)
+                        ? (await context.EvaluateFunctionHandleAsync(script, urlOrContent)) as ElementHandle
+                        : (await context.EvaluateFunctionHandleAsync(script, urlOrContent, type)) as ElementHandle;
```
kblok: Can't we do this?

```cs 
return await (string.IsNullOrEmpty(type)
    ? context.EvaluateFunctionHandleAsync(script, urlOrContent)
    : context.EvaluateFunctionHandleAsync(script, urlOrContent, type)) as ElementHandle;
```

lib/PuppeteerSharp/Frame.cs
```diff
@@ -406,31 +406,40 @@ public Task<string[]> SelectAsync(string selector, params string[] values)
...
+            const string addScriptContent = @"function addScriptContent(content, type = 'text/javascript') {
               const script = document.createElement('script');
-              script.type = 'text/javascript';
+              script.type = type;
               script.text = content;
               document.head.appendChild(script);
               return script;
             }";
 
+            async Task<ElementHandle> AddScriptTagPrivate(string script, string urlOrContent, string type)
```
kblok: Let's add `private` here

kblok: I don't like much the `Private` suffix here. This could be a private override right?

Meir017: It's a local method

kblok: 🤓

PR #373, opened 2018-06-30
Target refactor

### LINKED ISSUE #355: Refactor Browser.js into seperate files
https://github.com/kblok/puppeteer/pull/2/commits/ffe5b63dbade012744f6a91a6f9a89b3615d5246

This is not much about the refactor itself but about the changes they made on many methods and the target constructor

kblok:
closes #355

### DIFF
--- a/lib/PuppeteerSharp/Helpers/FlexibleStringEnumConverter.cs
+++ b/lib/PuppeteerSharp/Helpers/FlexibleStringEnumConverter.cs
@@ -1,0 +1,28 @@
+﻿using System;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
+
+namespace PuppeteerSharp.Helpers
+{
+    internal class FlexibleStringEnumConverter : StringEnumConverter
+    {
+        private Enum _fallbackValue;
+
+        public FlexibleStringEnumConverter(Enum fallbackValue)
+        {
+            _fallbackValue = fallbackValue;
+        }
+
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
+            try
+            {
+                return base.ReadJson(reader, objectType, existingValue, serializer);
+            }
+            catch
+            {
+                return _fallbackValue;
+            }
+        }
+    }
+}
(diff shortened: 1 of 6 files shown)


lib/PuppeteerSharp/Target.cs
```diff
@@ -41,7 +48,10 @@ internal Target(Browser browser, TargetInfo targetInfo)
         /// Gets the type. It will be <see cref="TargetInfo.Type"/> if it's "page" or "service_worker". Otherwise it will be "other"
         /// </summary>
         /// <value>The type.</value>
-        public string Type => _targetInfo.Type == "page" || _targetInfo.Type == "service_worker" || _targetInfo.Type == "browser" ? _targetInfo.Type : "other";
+        public string Type => _targetInfo.Type == "page" || _targetInfo.Type == "service_worker" || _targetInfo.Type == "browser" ?
```
Meir017: is it too late to introduce an enum for this?

lib/PuppeteerSharp/Target.cs
```diff
@@ -10,24 +12,29 @@ namespace PuppeteerSharp
     public class Target
     {
         #region Private members
-        private Browser _browser;
         private TargetInfo _targetInfo;
+        private string _targetId;
```
Meir017: these should be readonly

lib/PuppeteerSharp/Target.cs
kblok: It's never too late :)

kblok:
How about now @Meir017 ?

lib/PuppeteerSharp/Helpers/FlexibleStringEnumConverter.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
+
+namespace PuppeteerSharp.Helpers
+{
+    internal class FlexibleStringEnumConverter : StringEnumConverter
```
Meir017: I'm guessing this file exists for the `other` case

kblok: Exactly. I tried with `StringEnumConverter` but it failed when it got the type  `background_page`, and I was unable to find the complete list of types.

PR #385, opened 2018-07-10
Introduce url at error message at page.goto

### LINKED ISSUE #358: Introduce url at error message at page.goto
https://github.com/kblok/puppeteer/pull/2/commits/52d3bc05dea670e087afced923e288e4633c09b2

kblok:
Closes #358

### DIFF
--- a/lib/PuppeteerSharp/Page.cs
+++ b/lib/PuppeteerSharp/Page.cs
@@ -649,13 +649,13 @@ namespace PuppeteerSharp
             var referrer = _networkManager.ExtraHTTPHeaders?.GetValueOrDefault("referer");
             var requests = new Dictionary<string, Request>();
 
-            EventHandler<RequestEventArgs> createRequestEventListener = (object sender, RequestEventArgs e) =>
+            void createRequestEventListener(object sender, RequestEventArgs e)
             {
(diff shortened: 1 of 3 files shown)


lib/PuppeteerSharp/Page.cs
```diff
@@ -1740,7 +1740,7 @@ await Task.WhenAll(Frames.Select(frame => frame.EvaluateExpressionAsync(expressi
 
             if (response.errorText != null)
             {
-                throw new NavigationException(response.errorText.ToString());
+                throw new NavigationException($"{response.errorText.ToString()} at {url}");
```
Meir017: We could add an `Url` property on the `NavigationException` class

kblok: Would you leave the url on the `Message` as well?

Meir017: I think so, similar to how the [ArgumentException](https://github.com/dotnet/corefx/blob/master/src/Common/src/CoreLib/System/ArgumentException.cs#L77) works

kblok:
Done @Meir017

Meir017:
@kblok conflicts 😢

kblok:
Done @Meir017

PR #436, opened 2018-07-20
Screenshot improvements

### LINKED ISSUE #378: System.ArgumentException: 'options.Quality is unsupported for the jpeg screenshots'
Using the following snippet I get an exception when trying to take a screenshot:

```
await page.ScreenshotAsync(@"c:\tmp\MyScreenshot.jpeg", new ScreenshotOptions
{
	Type = "jpeg",
	Quality = 90
});
```

I'm assuming `Quality` should be an allowed option since the only other format (`png`) is lossless.

I can submit a PR for this if that's helpful.

### LINKED ISSUE #379: Explicitly setting file type in ScreenshotAsync should allow custom file extension
I get an exception `System.ArgumentException: 'Unknown options.type jpg'`
when doing this:

```
await page.ScreenshotAsync(@"c:\tmp\MyScreenshot.jpg", new ScreenshotOptions
{
	Type = "jpeg"
});
```

I think custom file extensions should be allowed when explicitly setting the `Type`. Alternately, `jpg` could be allowed as an alias for Type `jpeg`.

kblok:
closes #378 
closes #379  

We're introducing a breaking change here.

### DIFF
--- a/lib/PuppeteerSharp/ScreenshotOptions.cs
+++ b/lib/PuppeteerSharp/ScreenshotOptions.cs
@@ -16,7 +16,7 @@ namespace PuppeteerSharp
             ["jpe"] = ScreenshotType.Jpeg,
             ["jpeg"] = ScreenshotType.Jpeg,
             ["jpg"] = ScreenshotType.Jpeg,
-            ["png"] = ScreenshotType.Jpeg,
+            ["png"] = ScreenshotType.Png,
         };
 
         /// <summary>
--- a/lib/PuppeteerSharp/ScreenshotOptions.cs
+++ b/lib/PuppeteerSharp/ScreenshotOptions.cs
@@ -1,4 +1,6 @@
 ﻿using System;
+using System.Collections.Generic;
+using System.IO;
 using Newtonsoft.Json;
 using PuppeteerSharp.Media;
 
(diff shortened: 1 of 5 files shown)


lib/PuppeteerSharp/ScreenshotOptions.cs
```diff
@@ -32,12 +42,19 @@ public class ScreenshotOptions
...
-        public string Type { get; set; }
+        public ScreenshotType? Type { get; set; }
         /// <summary>
         /// The quality of the image, between 0-100. Not applicable to png images.
         /// </summary>
         /// <value>The quality.</value>
         [JsonProperty("quality")]
         public decimal? Quality { get; set; }
+
+        internal static ScreenshotType? GetScreenshotTypeFromFile(string file)
```
Meir017: is similar to https://github.com/GoogleChrome/puppeteer/blob/master/lib/Page.js#L794?

kblok: Yes, it matches. The AspNet.Core library has the `FileExtensionContentTypeProvider`, which resolves mime types. But I didn't want to add a dependency only to check 4 file extensions.

So that dictionary it's based on a mime type list, but it has the ScreenshotType as a result.

Meir017: do we need a test for this?

kblok: I'd make sense

kblok: Done @Meir017

PR #509, opened 2018-07-29
feat(BrowserContext): introduce Browser Contexts

### LINKED ISSUE #453: feat(BrowserContext): introduce Browser Contexts
https://github.com/kblok/puppeteer/pull/5/commits/3b03ff65c7b369debfcaa67a8673fba105a2d20d

Meir017:
closes #453

### DIFF
--- a/lib/PuppeteerSharp.Tests/BrowserContextTests/BrowserContextTests.cs
+++ b/lib/PuppeteerSharp.Tests/BrowserContextTests/BrowserContextTests.cs
@@ -21,8 +21,7 @@ namespace PuppeteerSharp.Tests.BrowserContextTests
             Assert.Single(Browser.BrowserContexts());
             var defaultContext = Browser.BrowserContexts()[0];
             Assert.False(defaultContext.IsIncognito);
-            var exception = await Assert.ThrowsAsync<PuppeteerException>(
-                () => defaultContext.CloseAsync());
+            var exception = await Assert.ThrowsAsync<PuppeteerException>(defaultContext.CloseAsync);
             Assert.Contains("cannot be closed", exception.Message);
         }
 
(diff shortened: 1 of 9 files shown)


Meir017:
@kblok I'm a bit stuck here 😢

kblok:
Good catch @cristiz1992

kblok:
Check this out @Meir017 https://github.com/kblok/puppeteer-sharp/commit/1257790ff07c74ff5f976f41b345415182a99bea

I bet the `should close all belonging targets once closing context` was wrong. I grabbed the assert values from master https://github.com/GoogleChrome/puppeteer/blob/master/test/browsercontext.spec.js#L44

lib/PuppeteerSharp.Tests/BrowserContextTests/BrowserContextTests.cs
```diff
@@ -0,0 +1,148 @@
...
+        }
+
+        [Fact]
+        public async Task ShouldHaveDefaultContext()
+        {
+            Assert.Single(Browser.BrowserContexts());
+            var defaultContext = Browser.BrowserContexts()[0];
+            Assert.False(defaultContext.IsIncognito);
+            var exception = await Assert.ThrowsAsync<PuppeteerException>(
+                () => defaultContext.CloseAsync());
```
kblok: This can me expressed as:
`var exception = await Assert.ThrowsAsync<PuppeteerException>(defaultContext.CloseAsync);`

lib/PuppeteerSharp/Browser.cs
```diff
@@ -139,24 +147,30 @@ public class Browser : IDisposable
...
 
+        public async Task<BrowserContext> CreateIncognitoBrowserContextAsync()
+        {
+            var response = await Connection.SendAsync<CreateBrowserContextResponse>("Target.createBrowserContext", new { });
+            var context = new BrowserContext(this, response.BrowserContextId);
+            _contexts[response.BrowserContextId] = context;
+            return context;
+        }
+
+        public BrowserContext[] BrowserContexts()
```
kblok: I'm wondering whether this should be a method or a readonly property

Meir017: It can't be a readonly property because the method `CreateIncognitoBrowserContextAsync` adds a new context to the browser

kblok: I mean something like this
```
public BrowserContext[] BrowserContexts
{
    get
	{
		var allContexts = new BrowserContext[_contexts.Count + 1];
        allContexts[0] = _defaultContext;
        _contexts.Values.CopyTo(allContexts, 0);
        return allContexts;
	}
}
```

Meir017: I made it a method to be like the `Targets` method

kblok:
If the tests are green and you add the missing XML comments (I wrote those here https://github.com/kblok/puppeteer-sharp/commit/1257790ff07c74ff5f976f41b345415182a99bea) I'm good

PR #511, opened 2018-07-29
feat(targets): add target.opener()

### LINKED ISSUE #468: feat(targets): add target.opener()
https://github.com/kblok/puppeteer/pull/5/commits/f6356683cdcadbae84e46bb61508a263e40d64b5

Meir017:
closes #468

### DIFF
--- a/lib/PuppeteerSharp/Target.cs
+++ b/lib/PuppeteerSharp/Target.cs
@@ -57,6 +57,25 @@ namespace PuppeteerSharp
         public string TargetId => _targetInfo.TargetId;
 
         /// <summary>
+        /// Get the target that opened this target
+        /// </summary>
+        /// <remarks>
+        /// Top-level targets return <c>null</c>.
+        /// </remarks>
+        public Target Opener
+        {
+            get
+            {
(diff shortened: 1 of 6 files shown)


lib/PuppeteerSharp/Target.cs
```diff
@@ -56,6 +56,25 @@ internal Target(TargetInfo targetInfo, Func<Task<CDPSession>> sessionFactory, Br
...
+        public Target Opener
+        {
+            get
+            {
+                var opener = _targetInfo.OpenerId;
+                if (opener == null)
+                {
+                    return null;
+                }
+                return Browser.TargetsMap.GetValueOrDefault(opener);
```
kblok: Can’t we solve this property in one line by sending a possible null to GetValueOrDefault?

Meir017: `GetValueOrDefault` uses `IDictionary<TKey, TValue>.TryGetValue` and will throw an `ArgumentNullException` exception if the argument is null

Meir017:
@kblok Done!

PR #521, opened 2018-08-03
implemented Pass through "background_page" types

### LINKED ISSUE #454: feat(Target): Pass through "background_page" types
https://github.com/kblok/puppeteer/pull/5/commits/23d25a6e1231bde9cc8d0bdc2145fbc1e4b008c6

Meir017:
closes #454

### DIFF
--- a/lib/PuppeteerSharp.Tests/TestConstants.cs
+++ b/lib/PuppeteerSharp.Tests/TestConstants.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Diagnostics;
+using System.IO;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Xunit;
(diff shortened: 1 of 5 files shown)


lib/PuppeteerSharp.Tests/TestConstants.cs
```diff
@@ -39,6 +41,17 @@ public static class TestConstants
             LogProcess = true
         };
 
+        public static LaunchOptions BrowserWithExtensionOptions() => new LaunchOptions
```
kblok: This can be a property

Meir017: It's similar to the `DefaultBrowserOptions` method

kblok: Fair enough :joy:

PR #525, opened 2018-08-03
Added .ConfigureAwait(false) to each await

xMarkos:
- to prevent deadlocks, every await must be paired with .ConfigureAwait(false)
- this is especially important when puppeteer is called from context which must wait for the task synchronously, e.g. using .GetAwaiter.GetResult() or .Result property

- for more info about this topic check https://msdn.microsoft.com/en-us/magazine/jj991977.aspx?f=255&MSPPError=-2147217396
- note: it is important that all devs understand why this is needed and that future code always uses ConfigureAwait(false)


kblok:
Great job @xMarkos!
I always considered ConfigureAwait an ugly boilerplate. But we needed this.

PR #526, opened 2018-08-03
Improved access to files to use true asynchronous methods

xMarkos:
- if file is not opened with parameter "useAsync: true", all calls to async method are in fact just wrappers to the synchronous methods, which might hurt performance
- you can read more info on this topic in article https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/async/using-async-for-file-access


kblok:
Great job @xMarkos!
I always considered ConfigureAwait an ugly boilerplate. But we needed this.

kblok:
Conflicted @xMarkos

lib/PuppeteerSharp/ElementHandle.cs
```diff
@@ -61,7 +61,7 @@ public async Task ScreenshotAsync(string file, ScreenshotOptions options)
 
             var data = await ScreenshotDataAsync(options).ConfigureAwait(false);
 
-            using (var fs = new FileStream(file, FileMode.Create, FileAccess.Write))
+            using (var fs = new FileStream(file, FileMode.Create, FileAccess.Write, FileShare.Read, 4096, true))
```
kblok: Can we move all the `4096` to a constant?
It’s a shame that it’s an internal constant I the FileStream class.
I can buy an internal constant in Frame or Page

xMarkos:
@kblok I have moved the logic to a new helper class "AsyncFile" so the default parameters are hidden there as well. Check the new commit whether it suits you or you think it needs more work.

kblok:
I love it man, great job.
Can we call it `AsyncFileHelper`?
I’m good after that.

xMarkos:
@kblok Done

PR #529, opened 2018-08-06
does not pass user options to browser constructor

### LINKED ISSUE #455: refactor: do not pass user options to browser constructor
https://github.com/kblok/puppeteer/pull/5/commits/51645932b7dc3c3f3d958deb667c22b3519bcc91

Meir017:
closes #455

### DIFF
--- a/lib/PuppeteerSharp/Browser.cs
+++ b/lib/PuppeteerSharp/Browser.cs
@@ -45,10 +45,10 @@ namespace PuppeteerSharp
         /// <param name="closeCallBack">An async function called before closing</param>
         public Browser(
             Connection connection,
-         bool ignoreHTTPSErrors,
-         bool setDefaultViewport,
-         Process process,
-         Func<Task> closeCallBack)
+            bool ignoreHTTPSErrors,
+            bool setDefaultViewport,
+            Process process,
+            Func<Task> closeCallBack)
         {
             Process = process;
             Connection = connection;
@@ -308,7 +308,7 @@ namespace PuppeteerSharp
         internal static async Task<Browser> CreateAsync(
             Connection connection,
             bool ignoreHTTPSErrors,
-         bool appMode,
+            bool appMode,
             Process process,
             Func<Task> closeCallBack)
         {
--- a/lib/PuppeteerSharp/Browser.cs
+++ b/lib/PuppeteerSharp/Browser.cs
@@ -53,7 +53,7 @@ namespace PuppeteerSharp
             Process = process;
             Connection = connection;
             IgnoreHTTPSErrors = ignoreHTTPSErrors;
-            _setDefaultViewport = setDefaultViewport;
+            DetDefaultViewport = setDefaultViewport;
             TargetsMap = new Dictionary<string, Target>();
             ScreenshotTaskQueue = new TaskQueue();
 
@@ -66,8 +66,7 @@ namespace PuppeteerSharp
 
         #region Private members
         internal readonly Dictionary<string, Target> TargetsMap;
-        internal readonly bool _setDefaultViewport;
-
+        
         private readonly Func<Task> _closeCallBack;
         private readonly ILogger<Browser> _logger;
         #endregion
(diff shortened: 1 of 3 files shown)


lib/PuppeteerSharp/Browser.cs
```diff
@@ -39,15 +39,21 @@ public class Browser : IDisposable
...
+            bool setDefaultViewport,
+            Process process,
+            Func<Task> closeCallBack)
         {
             Process = process;
             Connection = connection;
-            IgnoreHTTPSErrors = options.IgnoreHTTPSErrors;
-            AppMode = options.AppMode;
+            IgnoreHTTPSErrors = ignoreHTTPSErrors;
+            AppMode = setDefaultViewport;
```
kblok: In order to get closer to the pptr implementation, I’d remove the AppMode property and replace it with the SetDefaultViewport one.
This would also change how the target creates a page 
https://github.com/kblok/puppeteer-sharp/blob/6a16150d4bb6edee07ad1d5e7c39fcfb6e300715/lib/PuppeteerSharp/Target.cs#L96

lib/PuppeteerSharp/Browser.cs
```diff
@@ -66,6 +66,8 @@ public class Browser : IDisposable
 
         #region Private members
         internal readonly Dictionary<string, Target> TargetsMap;
+        internal readonly bool _setDefaultViewport;
```
kblok: How about an internal readonly property?
I don’t like much the idea of accessing `_fields` from other classes.

lib/PuppeteerSharp/Launcher.cs
```diff
@@ -125,7 +125,7 @@ public async Task<Browser> LaunchAsync(LaunchOptions options)
                     _logger.LogInformation("Process Count: {ProcessCount}", Interlocked.Increment(ref _processCount));
                 }
 
-                var browser = await Browser.CreateAsync(_connection, options, _chromeProcess, GracefullyCloseChrome).ConfigureAwait(false);
+                var browser = await Browser.CreateAsync(_connection, options.IgnoreHTTPSErrors, options.AppMode, _chromeProcess, GracefullyCloseChrome).ConfigureAwait(false);
```
kblok: According to pptr SetDefaultViewport is `!appMode` 
https://github.com/kblok/puppeteer/commit/51645932b7dc3c3f3d958deb667c22b3519bcc91#diff-cd756d74c4f724a8d0ebe5f47fff175aR163

lib/PuppeteerSharp/Target.cs
```diff
@@ -92,8 +92,8 @@ public Task<Page> PageAsync()
 
         private async Task<Page> CreatePageAsync()
         {
-            var session = await _sessionFactory().ConfigureAwait(false);
-            return await Page.CreateAsync(session, this, Browser.IgnoreHTTPSErrors, !Browser.AppMode, Browser.ScreenshotTaskQueue).ConfigureAwait(false);
+            var session = await _sessionFactory().ConfigureAwait(false);
+            return await Page.CreateAsync(session, this, Browser.IgnoreHTTPSErrors, !Browser.SetDefaultViewport, Browser.ScreenshotTaskQueue).ConfigureAwait(false);
```
kblok: It should be be `Browser.SetDefaultViewport` (remove the not) here 
https://github.com/kblok/puppeteer/blob/51645932b7dc3c3f3d958deb667c22b3519bcc91/lib/Target.js#L43

lib/PuppeteerSharp/Launcher.cs
```diff
@@ -156,7 +156,7 @@ public async Task<Browser> ConnectAsync(ConnectOptions options)
 
                 _connection = await Connection.Create(options.BrowserWSEndpoint, connectionDelay, keepAliveInterval, _loggerFactory).ConfigureAwait(false);
 
-                return await Browser.CreateAsync(_connection, options, null, () =>
+                return await Browser.CreateAsync(_connection, options.IgnoreHTTPSErrors, options.AppMode, null, () =>
```
kblok: Sorry for not picking everything at once. But here we should send `true` instead of appMode 
https://github.com/kblok/puppeteer/blob/51645932b7dc3c3f3d958deb667c22b3519bcc91/lib/Launcher.js#L233

PR #536, opened 2018-08-14
Fix SetViewportAsync instances in Readme

jankcat:
Also updated `browser` instances from uppercase as they are instance methods, not static methods


Meir017:
@jankcat Thank you for your contribution! 👍

PR #538, opened 2018-08-21
Change accessibiilty on Payload.

### LINKED ISSUE #537: Payload class is internal - how then to perform request interception
### Description

I'm trying to set a http header for the initial request to a target webpage. I tried using `SetExtraHttpHeadersAsync` but that appears not to be it's intended use (ie it passes the header to *all* requests that the target web page is making)

I then tried to use request interception as per one of the [unit test](https://github.com/kblok/puppeteer-sharp/blob/aa6470f61ab167b1525ebb74468811f386736b79/lib/PuppeteerSharp.Tests/PageTests/SetRequestInterceptionTests.cs#L154). However I can't use this method due to the `Payload` class and it's `Headers` property being marked as internal.

Am I missing some other method of achieving this? Otherwise, can these symbols be made public?

Thanks
Christian

kblok:
As the Request class is never treated as a Payload we are removing that dependency.

closes #537

### DIFF
--- a/lib/PuppeteerSharp/Payload.cs
+++ b/lib/PuppeteerSharp/Payload.cs
@@ -23,27 +23,26 @@ namespace PuppeteerSharp
         /// </summary>
         /// <value>HTTP method.</value>
         [JsonProperty("method"), JsonConverter(typeof(HttpMethodConverter))]
-        public HttpMethod Method { get; internal set; }
+        public HttpMethod Method { get; set; }
         /// <summary>
         /// Gets or sets the post data.
         /// </summary>
         /// <value>The post data.</value>
         [JsonProperty("postData")]
-        public object PostData { get; internal set; }
+        public object PostData { get; set; }
         /// <summary>
         /// Gets or sets the HTTP headers.
         /// </summary>
         /// <value>HTTP headers.</value>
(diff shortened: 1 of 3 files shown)


lib/PuppeteerSharp/Payload.cs
```diff
@@ -23,67 +23,66 @@ internal Payload()
...
-                        {
-                            continue;
-                        }
-                        hash.Headers[item.Key] = item.Value;
-                    }
-                }
-
-                return JsonConvert.SerializeObject(hash);
-            }
+        internal string Hash
```
Meir017: Shouldn't this still have the JsonIgnore attribute?

kblok: Why would we need it if we are not serializing this object?

Meir017: In that case remove all of the attributes in this class

kblok: `JsonProperty` it's used for deserialization. e.g. https://github.com/kblok/puppeteer-sharp/blob/master/lib/PuppeteerSharp/Messaging/RequestInterceptedResponse.cs#L13

Issue #512, opened 2018-07-30
LaunchAsync - Not working code just hangs


vinayhfs:
Hi 

I am just trying demo code in my project. Code just hangs at line 
   await new BrowserFetcher().DownloadAsync(BrowserFetcher.DefaultRevision);

I am on windows 7 ,Visiual studio 2017. installed nuget version 1.4.0

I trying generated pdf in .net core 2.0.6 web api method.

kblok:
We had some reports here about that https://github.com/kblok/puppeteer-sharp/issues/90#issuecomment-407046241

The internals of the the WebSocket class is not implemented on Windows 7
https://stackoverflow.com/questions/11039438/using-websocket-on-windows-7

fernandobelo:
I'm having the same problem here. Win 10 OK, Server 2008R2 NOK.

Anyone used https://github.com/PingmanTools/System.Net.WebSockets.Client.Managed ? It is supposed to be a drop-in replacement for System.Net.WebSockets, as it implements the abstract classes from it. 

Still haven´t got the chance to try, if it works, would remove a lot of work from changing the code to use websocket-sharp for example.

ggeurts:
I am working on pull request that uses the System.Net.WebSockets.Client.Managed NuGet package.

kblok:
@ggeurts that project doesn't look so active...

ggeurts:
It defaults to the regular ClientWebSocket usage unless it throws a Platform or supported option in which case it falls back to a port of Microsoft's managed WebSocket implementation.

ggeurts:
Pull request #574 introduces a Puppeteer extension point that enables use of custom WebSocket implementations. I have used this extension point in combination with the https://github.com/PingmanTools/System.Net.WebSockets.Client.Managed implementation to run Puppeteer successfully on Windows 2008 R2.

Ariex:
just for record, this is how it make works for me:
``` C#
var browser = await Puppeteer.LaunchAsync(new LaunchOptions
{
	WebSocketFactory = (uri, socketOptions, cancellationToken) =>
	{
		return System.Net.WebSockets.SystemClientWebSocket.ConnectAsync(uri, cancellationToken);
	},
	Headless = true
});
```

PR #574, opened 2018-08-28
Extension point enabling custom web sockets

ggeurts:
Microsoft's default `WebSocket` implementation does not work on Windows 7 and Windows 2008 R2 platforms. Alternative web sockets implementation exist that do work on those platforms. Therefore it makes sense to add an extension point, so library users can use Puppeteer on these platforms, without Puppeteer becoming dependent on other `WebSocket` implementations.


ggeurts:
This pull request addresses issue #512.

lib/PuppeteerSharp.Tests/PuppeteerTests/PuppeteerLaunchTests.cs
```diff
@@ -454,12 +471,12 @@ private string GetSubprocessWorkingDir(string dir)
...
-                dir,
-                "bin",
-                build,
-                "net471");
+            return Path.Combine(
+            TestUtils.FindParentDirectory("lib"),
+            dir,
+            "bin",
+            build,
+            "net471");
```
kblok: Let's retore the tabs here.

lib/PuppeteerSharp/Connection.cs
```diff
@@ -276,14 +276,27 @@ private void ProcessResponse(string response)
...
-        internal static async Task<Connection> Create(string url, int delay = 0, int keepAliveInterval = 60, ILoggerFactory loggerFactory = null)
+        /// <summary>
+        /// Gets default <see cref="WebSocketFactory"/> implementation.
+        /// </summary>
+        public static readonly WebSocketFactory DefaultWebSocketFactory = async (uri, options, cancellationToken) =>
+        {
+            var result = new System.Net.WebSockets.ClientWebSocket();
+#pragma warning disable 618
+            result.Options.KeepAliveInterval = TimeSpan.FromMilliseconds(options.KeepAliveInterval);
+#pragma warning restore 618
```
kblok: Why are we doing this?

lib/PuppeteerSharp/Connection.cs
```diff
@@ -276,14 +276,27 @@ private void ProcessResponse(string response)
...
+
         #region Static Methods
 
-        internal static async Task<Connection> Create(string url, int delay = 0, int keepAliveInterval = 60, ILoggerFactory loggerFactory = null)
+        /// <summary>
+        /// Gets default <see cref="WebSocketFactory"/> implementation.
+        /// </summary>
+        public static readonly WebSocketFactory DefaultWebSocketFactory = async (uri, options, cancellationToken) =>
+        {
+            var result = new System.Net.WebSockets.ClientWebSocket();
```
kblok: How about adding a using?

lib/PuppeteerSharp/WebSocketFactory.cs
```diff
@@ -0,0 +1,16 @@
...
+namespace PuppeteerSharp
+{
+    /// <summary>
+    /// Factory method for connected web sockets.
+    /// </summary>
+    /// <param name="uri">The uri to connect to.</param>
+    /// <param name="options">Web socket creation options.</param>
+    /// <param name="cancellationToken">Cancellation token.</param>
+    /// <returns>A <see cref="WebSocket"/> that is connected to the given <paramref name="uri"/>.</returns>
+    public delegate Task<WebSocket> WebSocketFactory(Uri uri, IWebSocketOptions options, CancellationToken cancellationToken);
```
kblok: I'd just declare that as an Action instead of creating a delegate.

lib/PuppeteerSharp/IWebSocketOptions.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using System.Net.WebSockets;
+
+namespace PuppeteerSharp
+{
+    /// <summary>
+    /// Options for <see cref="WebSocket"/> creation.
+    /// </summary>
+    public interface IWebSocketOptions
```
kblok: Do we have any class which is IWebSocketOptions but it's not IConnectionOptions?

lib/PuppeteerSharp/Connection.cs
ggeurts: KeepAliveInterval is marked obsolete

lib/PuppeteerSharp/IWebSocketOptions.cs
ggeurts: No, we don't. `IConnectionOptions` inherits `IWebSocketOptions`. `IWebSocketOptions` defines options for `WebSocketFactory` delegate, which is a subset of what is required by the `Connection` class. My preference was to keep `WebSocketFactory` delegate as independent as possible of the rest of the code.

lib/PuppeteerSharp/WebSocketFactory.cs
ggeurts: That would become a `Func<Uri, IWebSocketOptions, CancellationToken, Task<WebSocket>>` which is significantly less self-documenting than the proposed delegate.

lib/PuppeteerSharp/Connection.cs
kblok: I don't see it marked as obsolete https://github.com/dotnet/corefx/blob/master/src/System.Net.WebSockets.Client/src/System/Net/WebSockets/ClientWebSocketOptions.cs#L155

Do you have any doc?

kblok:
We have a real fail in AppVeyor

ggeurts:
The AppVeyor fail is related to `Connection.CloseAsync` not being suffciently re-entrant/thread-safe. I am happy to propose an adjustment, but it isn't related to this pull request.

kblok:
@ggeurts We can't merge a red build, so we'd need a complete solution.

ggeurts:
Fine, I will commit a fix

ggeurts:
Many thanks for considering, reviewing and using this contribution!

kblok:
Thank for contributing @ggeurts !

PR #575, opened 2018-08-29
Introduced workers

### LINKED ISSUE #458: feat(Page): introduce workers
* https://github.com/kblok/puppeteer/pull/5/commits/93fe2b57d69d99bce037420422bbd7bee92471d3
* https://github.com/kblok/puppeteer/pull/5/commits/1c2adf61e971577a6bc96621fcb78ee88a9a9ff8
* https://github.com/kblok/puppeteer/pull/5/commits/35e3f12afaceb6f3b3d4ee6b7e7111e3f6dd941a
* https://github.com/kblok/puppeteer/pull/5/commits/2ff0adcad8207f85865ba35533605361a66515e6

We also need to take into consideration this change https://github.com/kblok/puppeteer/commit/a0584689480c4e2102279969b515a837021f6f24

### LINKED ISSUE #473: fix: emit all arguments from worker console logs
https://github.com/kblok/puppeteer/pull/5/commits/75ba86f41ac4196dcfee515900265edc44ffa2cd

kblok:
Based on #539 
closes #458 
closes #473


lib/PuppeteerSharp.TestServer/wwwroot/worker/worker.js
```diff
@@ -0,0 +1,13 @@
...
+ function workerFunction() {
+  return 'worker function result';
+}
+ self.addEventListener('message', event => {
+  console.log('got this data: ' + event.data);
+});
+ (async function() {
+  while (true) {
+    self.postMessage(workerFunction.toString());
+    await new Promise(x => setTimeout(x, 10000));
```
Meir017: change to 100

lib/PuppeteerSharp.Tests/TestConstants.cs
```diff
@@ -10,16 +10,16 @@ namespace PuppeteerSharp.Tests
 {
     public static class TestConstants
     {
-        public const int Port = 8907;
+        public const int Port = 8081;
```
Meir017: why the change?

lib/PuppeteerSharp/Worker.cs
```diff
@@ -0,0 +1,117 @@
...
+            => await (await ExecutionContextTask.ConfigureAwait(false)).EvaluateExpressionHandleAsync(script).ConfigureAwait(false);
+
+        internal Task<ExecutionContext> ExecutionContextTask => _executionContextCallback.Task;
+
+        internal async void OnMessageReceived(object sender, MessageEventArgs e)
+        {
+            switch (e.MessageID)
+            {
+                case "Runtime.executionContextCreated":
+                    if (_jsHandleFactory == null)
```
Meir017: move each case to a private method

lib/PuppeteerSharp.Tests/TestConstants.cs
kblok: So we don’t use the same port as puppeteer.
If we do that we can debug both projects at the same time when we are desperate:joy:
Do you think this might hurt us?

Meir017: probably not, I just wasn't sure why you made the change

kblok:
Done @Meir017

PR #589, opened 2018-09-07
Nicer protocol error messages

### LINKED ISSUE #481: feat: nicer protocol error messages
https://github.com/kblok/puppeteer/pull/7/commits/73f9c48081017c228ec1f1318c3c7946321e0538

kblok:
Breaking Change: Now EvaluateFunctionHandleAsync and EvaluateExpressionHandleAsync will throw an EvaluationException.

closes #481

### DIFF
--- a/lib/PuppeteerSharp.Tests/PageTests/CloseTests.cs
+++ b/lib/PuppeteerSharp.Tests/PageTests/CloseTests.cs
@@ -17,7 +17,7 @@ namespace PuppeteerSharp.Tests.PageTests
             // Put into a var to avoid warning
             var t = Page.CloseAsync();
 
-            var exception = await Assert.ThrowsAsync<TargetClosedException>(async () => await neverResolves);
+            var exception = await Assert.ThrowsAnyAsync<PuppeteerException>(async () => await neverResolves);
 
             Assert.Contains("Protocol error", exception.Message);
         }
--- a/lib/PuppeteerSharp.Tests/PageTests/CloseTests.cs
+++ b/lib/PuppeteerSharp.Tests/PageTests/CloseTests.cs
@@ -17,8 +17,8 @@ namespace PuppeteerSharp.Tests.PageTests
             // Put into a var to avoid warning
             var t = Page.CloseAsync();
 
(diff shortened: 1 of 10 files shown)


lib/PuppeteerSharp.Tests/PageTests/CloseTests.cs
```diff
@@ -17,7 +17,7 @@ public async Task ShouldRejectAllPromisesWhenPageIsClosed()
             // Put into a var to avoid warning
             var t = Page.CloseAsync();
 
-            var exception = await Assert.ThrowsAsync<TargetClosedException>(async () => await neverResolves);
+            var exception = await Assert.ThrowsAnyAsync<PuppeteerException>(async () => await neverResolves);
```
Meir017: this should use `ThrowsAsync` instead

lib/PuppeteerSharp.Tests/PuppeteerTests/PuppeteerLaunchTests.cs
```diff
@@ -97,7 +97,7 @@ public async Task ShouldRejectAllPromisesWhenBrowserIsClosed()
             {
                 var neverResolves = page.EvaluateFunctionHandleAsync("() => new Promise(r => {})");
                 await browser.CloseAsync();
-                var exception = await Assert.ThrowsAsync<TargetClosedException>(() => neverResolves);
+                var exception = await Assert.ThrowsAnyAsync<PuppeteerException>(() => neverResolves);
```
Meir017: this should use `ThrowsAsync` instead

lib/PuppeteerSharp.Tests/PageTests/CloseTests.cs
kblok: I changted this to `Any` because this changed from `TargetClosedException` to `EvaluationException`(which has `TargetClosedException` as inner exception).
So here I changed to any because we don't care if which exception is as long as we throw something.

What do you think?

Meir017: since the test tests specifically that the method throws an exception I think this test should care about the exception type

PR #617, opened 2018-09-14
Improve Request overrides on ContinueAsync

### LINKED ISSUE #616: Request intercept ContinueAsync(Payload) hanging
I am trying to get a POST request working, but am getting navigation timeouts anytime a Payload is passed to Request.ContinueAsync.
Payload does have an internal constructor, indicating this probably isn't intended to be working yet, but considering how well everything else works, I got my hopes up.

```csharp
    public class Program
    {
        public static void Main(string[] args)
        {
            MainAsync(args).GetAwaiter().GetResult();
        }
        public static async Task MainAsync(string[] args)
        {
            var url = "https://www.google.com";
            var options = new LaunchOptions
            {
                Headless = true,
                Timeout = 10000
            };
            
            await new BrowserFetcher().DownloadAsync(BrowserFetcher.DefaultRevision);
            
            using (var browser = await Puppeteer.LaunchAsync(options))
            {
                {
                    await page.SetRequestInterceptionAsync(true);
                    page.Request += async (sender, e) =>
                    {
                        var payload = CreateInstance<Payload>();
                        payload.Headers = e.Request.Headers;
                        payload.Method = HttpMethod.Get; // need to eventually set to POST, and pass down some postData

                        System.Diagnostics.Debug.WriteLine("request hook BEFORE " + e.Request.Url);

                        // always hangs whenever a payload is passed in
                        await e.Request.ContinueAsync(payload);

                        // works if no payload passed in
                        //await e.Request.ContinueAsync();

                        // works if explicit response
                        //await e.Request.RespondAsync(new PuppeteerSharp.ResponseData
                        //{
                        //    Status = HttpStatusCode.OK,
                        //    Body = "<html>hello world</html>"
                        //});

                        System.Diagnostics.Debug.WriteLine("request hook AFTER " + e.Request.Url);
                    };
                    await page.EmulateMediaAsync(MediaType.Screen);

                    await page.GoToAsync(url);
                    
                    var pdfOptions = new PdfOptions
                    {
                        PrintBackground = true
                    };

                    await page.PdfAsync(Path.Combine(Directory.GetCurrentDirectory(), "testPost.pdf"), pdfOptions);
                }
            }
        }

        public static T CreateInstance<T>(params object[] args)
        {
            // apologies in advance
            var type = typeof(T);
            var instance = type.Assembly.CreateInstance(
                type.FullName, false,
                BindingFlags.Instance | BindingFlags.NonPublic,
                null, args, null, null);
            return (T)instance;
        }
    }
```

### Expected behavior:

Ability to continue with payload like POST data.

### Actual behavior:

Navigation timeout.

### Versions:

* Which version of PuppeteerSharp are you using?
1.5

* Which .NET runtime and version are you targeting?
.NET Core 2.1
.NET Framework 4.7.2

### Related Issues?
* closed Issue https://github.com/kblok/puppeteer-sharp/issues/537
* open Issue https://github.com/kblok/puppeteer-sharp/issues/598

Excellent work on this project btw, very impressive.

kblok:
Closes #616

### DIFF
--- a/lib/PuppeteerSharp.Tests/Issues/Issue0616.cs
+++ b/lib/PuppeteerSharp.Tests/Issues/Issue0616.cs
@@ -1,0 +1,50 @@
+﻿using Microsoft.AspNetCore.Http;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Net;
+using System.Net.Http;
+using System.Text;
+using System.Threading.Tasks;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace PuppeteerSharp.Tests.PageTests
+{
+    [Collection("PuppeteerLoaderFixture collection")]
+    public class Issue0616 : PuppeteerPageBaseTest
+    {
+        public Issue0616(ITestOutputHelper output) : base(output)
+        {
+        }
+
+        [Fact]
+        public async Task ShouldBeAbleToChangeToPost()
+        {
+            await Page.SetRequestInterceptionAsync(true);
+            Page.Request += async (sender, e) =>
+            {
+                var payload = new Payload()
+                {
+                    Method = HttpMethod.Post,
+                    PostData = "foo"
+                };
+                await e.Request.ContinueAsync(payload);
+            };
+
+            var requestTask = Server.WaitForRequest<string>("/grid.html", request =>
+            {
+                return request.Form.ToString();
+            });
+            var gotoTask = Page.GoToAsync(TestConstants.ServerUrl + "/grid.html");
+
+            await Task.WhenAll(
+                requestTask,
+                gotoTask
+            );
+
+            Assert.Equal("http://google.com/", requestTask.Result);
+        }
+    }
+}
--- a/lib/PuppeteerSharp.Tests/Issues/Issue0616.cs
+++ b/lib/PuppeteerSharp.Tests/Issues/Issue0616.cs
@@ -10,7 +10,7 @@ using System.Threading.Tasks;
 using Xunit;
 using Xunit.Abstractions;
 
-namespace PuppeteerSharp.Tests.PageTests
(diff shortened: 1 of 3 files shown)


lib/PuppeteerSharp/Payload.cs
```diff
@@ -13,7 +13,10 @@ namespace PuppeteerSharp
     /// </summary>
     public class Payload
     {
-        internal Payload()
+        /// <summary>
+        /// Initializes a new instance of the <see cref="T:PuppeteerSharp.Payload"/> class.
```
Meir017: T: 😕

Meir017:
small fix and merge

lib/PuppeteerSharp/Payload.cs
kblok: IDEs....

lib/PuppeteerSharp.Tests/Issues/Issue0616.cs
```diff
@@ -0,0 +1,50 @@
...
+            {
+                using (var reader = new StreamReader(context.Request.Body, Encoding.UTF8))
+                {
+                    var request = await reader.ReadToEndAsync();
+                    await context.Response.WriteAsync(request);
+                }
+            });
+            var response = await Page.GoToAsync(TestConstants.ServerUrl + "/grid.html");
+
+            Assert.Equal(TestConstants.ServerUrl + "/grid.html", response.Url);
```
Meir017: I think you should test the response method too, it should be `POST` not `GET`

kblok: I checked that on Puppeteer. The `Network.setRequestInterception` message doesn't override the original request values, so it returns a `GET`
screen shot 2018-09-14 at 08 36 17

PR #624, opened 2018-09-16
Typing emoji

### LINKED ISSUE #559: Typing emoji
https://github.com/kblok/puppeteer/pull/10/commits/3335d369d3c5054517db7cdc541aba31184b9d07

kblok:
closes #559

### DIFF
--- a/lib/PuppeteerSharp.Tests/InputTests/InputTests.cs
+++ b/lib/PuppeteerSharp.Tests/InputTests/InputTests.cs
@@ -605,6 +605,7 @@ namespace PuppeteerSharp.Tests.InputTests
                 "👹 Tokyo street Japan \uD83C\uDDEF\uD83C\uDDF5",
                 await Page.QuerySelectorAsync("textarea").EvaluateFunctionAsync<string>("t => t.value"));
         }
+
         [Fact]
         public async Task ShouldTypeEmojiIntoAniframe()
         {
--- a/lib/PuppeteerSharp.Tests/InputTests/InputTests.cs
+++ b/lib/PuppeteerSharp.Tests/InputTests/InputTests.cs
@@ -600,9 +600,9 @@ namespace PuppeteerSharp.Tests.InputTests
         public async Task ShouldTypeEmoji()
         {
             await Page.GoToAsync(TestConstants.ServerUrl + "/input/textarea.html");
-            await Page.TypeAsync("textarea", "👹 Tokyo street Japan");
+            await Page.TypeAsync("textarea", "👹 Tokyo street Japan \uD83C\uDDEF\uD83C\uDDF5");
             Assert.Equal(
-                "👹 Tokyo street Japan",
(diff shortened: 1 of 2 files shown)


lib/PuppeteerSharp.Tests/InputTests/InputTests.cs
```diff
@@ -595,5 +595,28 @@ public async Task ShouldThrowOnUnknownKeys()
 
             await Assert.ThrowsAsync<KeyNotFoundException>(() => Page.Keyboard.PressAsync("😊"));
         }
+
+        [Fact]
+        public async Task ShouldTypeEmoji()
+        {
+            await Page.GoToAsync(TestConstants.ServerUrl + "/input/textarea.html");
+            await Page.TypeAsync("textarea", "👹 Tokyo street Japan \uD83C\uDDEF\uD83C\uDDF5");
```
kblok: I'm adding the Japanese flag as Unicode because it breaks the IDE :joy:

lib/PuppeteerSharp/Input/Keyboard.cs
```diff
@@ -109,8 +103,11 @@ public async Task TypeAsync(string text, TypeOptions options = null)
             {
                 delay = (int)options.Delay;
             }
-            foreach (var letter in text)
+
+            var textParts = StringInfo.GetTextElementEnumerator(text);
```
Meir017: why was this change required?

lib/PuppeteerSharp/Input/Keyboard.cs
```diff
@@ -41,7 +42,7 @@ public Task DownAsync(string key, DownOptions options = null)
 
             var text = options?.Text == null ? description.Text : options.Text;
 
-            return _client.SendAsync("Input.dispatchKeyEvent", new Dictionary<string, object>(){
+            return _client.SendAsync("Input.dispatchKeyEvent", new Dictionary<string, object>{
```
Meir017: formatting ;)

kblok: Done

lib/PuppeteerSharp/Input/Keyboard.cs
kblok: I was just preparing a post about this. Emojis (let's call it Unicode chars) uses 4 bytes. So you need 2 chars to represent one Unicode char. If you do `string.ToCharArray` and send one by one to Chromium you would get invalid characters (a.k.a. □) on the other side.
So you need a method, `GetTextElementEnumerator`, that returns the "elements" of the text. `GetTextElementEnumerator` is able to return a \u1F479 (which are 2 chars: \u1F47 and 9) as one element.
From the [docs](https://docs.microsoft.com/en-us/dotnet/api/system.globalization.stringinfo.gettextelementenumerator?view=netframework-4.7.2):
>The .NET Framework defines a text element as a unit of text that is displayed as a single character, that is, a grapheme. A text element can be a base character, a surrogate pair, or a combining character sequence.

PR #643, opened 2018-09-24
Roll Chromium to r588429

kblok:


Meir017:
@kblok looks like it might be a real fail

kblok:
Yeah, that's why I changed it to WIP. I'll fix it this evening.

PR #644, opened 2018-09-24
CDPSession SendAsync access modifiers fixed.

asalih:


kblok:
@Meir017 I don't have a problem either. I didn't want to make it public not to open that "door". But I think that if the user reads the CDP Session docs and finds a method useful for her, she can go can make the call.
I was interested in your opinion.

asalih:
Thanks for the great project guys.

Issue #429, opened 2018-07-18
Implement CI for Linux on AppVeyor


kblok:
We already have a CI project on AppVeyor https://ci.appveyor.com/project/kblok/puppeteer-sharp-esipd.
This project is waiting for a config file `appveyor-linux.yml`.
We need a config file for linux which needs to bit similar but simpler than https://github.com/kblok/puppeteer-sharp/blob/master/appveyor.yml:

 * We don't need docfx
 * It should only have one matrix
 * It doesn't need to build docs nor publish to nuget.

We already have done some work here https://github.com/kblok/puppeteer-sharp/pull/430. So we would need to branch from that PR instead of master.

Meir017:
@kblok AppVeyor supports Linux so it might be easier to use it then Travis

HarpyWar:
I tried setup travis but without a success https://github.com/HarpyWar/puppeteer-sharp/commits/master

* With `netcoreapp2.0` on Linux  tests hangs up at beginning and stopped by timeout https://travis-ci.org/HarpyWar/puppeteer-sharp/jobs/432685571. I had the same behavior on my Linux machine in my program, until I installed chrome dependencies with `apt-get` https://docs.browserless.io/blog/2018/04/25/chrome-linux.html. But here it did not help. May be it's required to add `--no-sandbox` args to chrome because of running under root? But on my linux I received appropriate warning message and here not.
* On Mac `netcoreapp2.0` failed with errors https://travis-ci.org/HarpyWar/puppeteer-sharp/jobs/432685572

* With `netcoreapp2.1` both failed with error `It was not possible to find any compatible framework version` https://travis-ci.org/HarpyWar/puppeteer-sharp/jobs/432676494. I have added 2.1 target in all csproj files, may be it's required to add something else?


To activate travis-ci on the repository you have to login on https://travis-ci.org, select the project and click on activate button. Before add `.travis.yml` into the repository.

Status badge for travis can be added this way:
```
[![Build Status](https://travis-ci.org/kblok/puppeteer-sharp.svg?branch=master)](https://travis-ci.org/kblok/puppeteer-sharp)
```

HarpyWar:
Some progress https://travis-ci.org/HarpyWar/puppeteer-sharp/builds/432858789
I get the same output in travis as on my Linux machine.

Anyone knows how to correctly run tests of puppeteer-sharp on Linux and Mac?

PR #430, opened 2018-07-18
[WIP] Ensure executable permissions in Linux and OSX

kblok:
I implemented a platform check because it doesn't make sense to implement this on Windows, I bet NodeJS ignores it.

I also created the issue #429, You should test this on Linux.

closes #377 
closes #402

PR #646, opened 2018-09-25
Travis-CI

HarpyWar:
https://github.com/kblok/puppeteer-sharp/issues/429


lib/PuppeteerSharp/NavigatorWatcher.cs
```diff
@@ -37,7 +37,7 @@ public NavigatorWatcher(FrameManager frameManager, Frame mainFrame, int timeout,
 
             _expectedLifecycle = waitUntil.Select(w =>
             {
-                var protocolEvent = _puppeteerToProtocolLifecycle.GetValueOrDefault(w);
+                var protocolEvent = DictionaryExtensions.GetValueOrDefault(_puppeteerToProtocolLifecycle, w);
```
Meir017: why can't the method be called as an extension?

HarpyWar: I think compiler is right, and this will persist in the following dotnet versions. 
What difference of this extension method between `System.Collections.Generic.CollectionExtensions.GetValueOrDefault`?

Meir017: @kblok

HarpyWar: The problem is `GetValueOrDefault` exists in different namespaces `System.Collections.Generic` and `DictionaryExtensions.GetValueOrDefault`. Both namespaces are used, because of that compile error with ambiguous calls. 
I dont understand the purpose of DictionaryExtensions, at first sight it does the same as standart library.

kblok: We can't use `System.Collections.Generic.CollectionExtensions`
Puppeteer-Sharp project must be a NetStandard Library.

lib/PuppeteerSharp.DevicesFetcher/PuppeteerSharp.DevicesFetcher.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <TargetFrameworks>netcoreapp2.0;netcoreapp2.1</TargetFrameworks>
```
kblok: These projects are for internal use; we can bump them to 2.1 instead of making them multi-target.

lib/PuppeteerSharp.TestServer/PuppeteerSharp.TestServer.csproj
```diff
@@ -1,6 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk.Web">
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netcoreapp2.0;netcoreapp2.1;netstandard2.0</TargetFrameworks>
```
kblok: Same here

lib/PuppeteerSharp.Tests.CloseMe/PuppeteerSharp.Tests.CloseMe.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>netcoreapp2.0;net471</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.0;netcoreapp2.1;net471</TargetFrameworks>
```
kblok: Same here

lib/PuppeteerSharp.Tests.DumpIO/PuppeteerSharp.Tests.DumpIO.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>netcoreapp2.0;net471</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.0;netcoreapp2.1;net471</TargetFrameworks>
```
kblok: Same here

lib/PuppeteerSharp.Tests/PuppeteerSharp.Tests.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.0;net471</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.0;netcoreapp2.1;net471</TargetFrameworks>
```
kblok: Same here

lib/PuppeteerSharp/PuppeteerSharp.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netcoreapp2.0;netcoreapp2.1;netstandard2.0</TargetFrameworks>
```
kblok: As I mentioned PuppeteerSharp should only be a netstandard project

lib/PuppeteerSharp.Tests/PuppeteerSharp.Tests.csproj
HarpyWar: Other projects ok, but if not keep both `netcoreapp2.0;netcoreapp2.1;` here then it will not be possible to run tests on different dotnet core versions, isn't it? 
```
 dotnet test -f netcoreapp2.0 -s test.runsettings
 dotnet test -f netcoreapp2.1 -s test.runsettings
```
Or it should be tested only by the latest dotnet?

kblok: You're right. Maybe I wouldn't test both versions on builds (I would test 2.1), but we could leave these frameworks available for testing.
So leave all of these but PuppeteerSharp, which needs to be NetStandard.

lib/PuppeteerSharp/PuppeteerSharp.csproj
HarpyWar: If remove it from `PuppeteerSharp.csproj` and `PuppeteerSharp.TestServer.csproj` then it's not possible to build the dll:
```
dotnet build -f netcoreapp2.1
```
How to do it without a definition inside csproj?

kblok: I think you should be able to use msbuild on Linux https://ci.appveyor.com/project/kblok/puppeteer-sharp/build/1.0.1626/job/tlxntjkyfpxd465y#L121

lib/PuppeteerSharp/NavigatorWatcher.cs
HarpyWar: So it's proper solution or it should be done other way like @Meir017 wrote? 
I get ambiguous calls on every version of dotnet core (2.0 and 2.1).

kblok: The thing is that PuppetterSharp is NetStandard, not NetCore. Can you try removing the Target frameworks you added?

HarpyWar: Ok, I reverted these changes. I'm able to build the project with mono msbuild but I did succeed with dotnet core. 

dotnet core has built-it msbuild which can be run with `dotnet msbuild`, by default it's required .NETFramework,Version=v4.7.1 for this project, but this framework only for Windows. I tried run it like this:
`dotnet msbuild /p:TargetFrameworkIdentifier=.NETCoreApp /p:TargetFrameworkVersion=2.1`, but actully it's the same as `dotnet build -f dotnet netcoreapp2.1` and it's required the appropriate target to be defined in csproj.

kblok: @HarpyWar on Linux we only need to run the tests. We would package the app on Windows. This should work on Linux:
```
dotnet test -f netcoreapp2.0
```

`dotnet test` will build all the dependencies.

lib/PuppeteerSharp/PuppeteerSharp.csproj
HarpyWar: With mono msbuild all projects compile successfully except `PuppeteerSharp.Tests.csproj`. What could be wrong there? https://travis-ci.org/HarpyWar/puppeteer-sharp/jobs/433140784#L2276-L2277

Also should mono be included into travis to build the solution, or it's enough in appveyour?

kblok: AppVeyor now supports Linux, so We would build there.
But your work here would help us to setup AppVeyor, in fact, we could close this task setting up AppVeyor.
So:
1. I think you only need to `dotnet test -f netcoreapp2.0 -s test.runsettings` You don't need anything else.
2. We need to be able to create a self-signed certificate as we do in Windows https://github.com/kblok/puppeteer-sharp/blob/master/appveyor.yml#L22
3. You would need to incorporate this PR https://github.com/kblok/puppeteer-sharp/pull/430. It will do a chmod on the downloaded file.

We have our channel (Puppeteer-Sharp) on the Puppeteer's slack workspace https://puppeteer.slack.com/messages fill free to jump there to chat.

PR #659, opened 2018-09-30
Implement BrowserContext.OverridePermissionsAsync

kblok:
closes #600

PR #661, opened 2018-10-01
Page geolocation

### LINKED ISSUE #601: Introduce page.setGeolocation method
https://github.com/kblok/puppeteer/pull/11/commits/17029281a92446fcab2357cc52dcdfa7239aa4f5

kblok:
closes #601

based on #659

### DIFF
--- a/lib/PuppeteerSharp/BrowserContext.cs
+++ b/lib/PuppeteerSharp/BrowserContext.cs
@@ -14,6 +14,27 @@ namespace PuppeteerSharp
         private readonly IConnection _connection;
         private readonly string _id;
 
+        private static readonly Dictionary<OverridePermission, string> _webPermissionToProtocol = new Dictionary<OverridePermission, string>
+        {
+            [OverridePermission.Geolocation] = "geolocation",
+            [OverridePermission.Midi] = "midi",
+            [OverridePermission.Notifications] = "notifications",
+            [OverridePermission.Push] = "push",
+            [OverridePermission.Camera] = "videoCapture",
+            [OverridePermission.Microphone] = "audioCapture",
+            [OverridePermission.BackgroundSync] = "backgroundSync",
+            [OverridePermission.AmbientLightSensor] = "sensors",
+            [OverridePermission.Accelerometer] = "sensors",
+            [OverridePermission.Gyroscope] = "sensors",
+            [OverridePermission.Magnetometer] = "sensors",
+            [OverridePermission.AccessibilityEvents] = "accessibilityEvents",
+            [OverridePermission.ClipboardRead] = "clipboardRead",
+            [OverridePermission.ClipboardWrite] = "clipboardWrite",
+            [OverridePermission.PaymentHandler] = "paymentHandler",
+            // chrome-specific permissions we have.
+            [OverridePermission.MidiSysex] = "midiSysex"
+        };
+
         internal BrowserContext(IConnection connection, Browser browser, string contextId)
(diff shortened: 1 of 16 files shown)


Meir017:
@kblok  missing the cr commit...

kblok:
@Meir017 from the other branch? if it's that: done :)

lib/PuppeteerSharp/BrowserContext.cs
```diff
@@ -83,6 +107,46 @@ public Task CloseAsync()
...
+        /// <example>
+        /// <![CDATA[
+        /// var context = browser.DefaultBrowserContext;
+        /// await context.OverridePermissionsAsync("https://html5demos.com", new List<string> {"geolocation"});
+        /// ]]>
+        /// </example>
+        /// <seealso href="https://developer.mozilla.org/en-US/docs/Glossary/Origin"/>
+        public async Task OverridePermissionsAsync(string origin, IEnumerable<OverridePermission> permissions)
+        {
+            var permissionList = permissions.Select(permission =>
```
Meir017: since `permissions` is strongly typed could we just add an `EnumMember` attribute to each enum member?
if this works maybe also cleanup [`WaitUntilNavigation`](https://github.com/kblok/puppeteer-sharp/blob/master/lib/PuppeteerSharp/NavigatorWatcher.cs#L11) ?

lib/PuppeteerSharp/Page.cs
```diff
@@ -310,6 +310,32 @@ public class Page : IDisposable
...
 
+        /// <summary>
+        /// Sets the page's geolocation.
+        /// </summary>
+        /// <returns>The task.</returns>
+        /// <param name="options">Geolocation options.</param>
+        /// <remarks>
+        /// Consider using <seealso cref="BrowserContext.OverridePermissionsAsync(string, IEnumerable{OverridePermission})"/> to grant permissions for the page to read its geolocation.
+        /// </remarks>
+        public async Task SetGeolocationAsync(GeolocationOption options)
```
Meir017: this method doesn't have to use the `async` keyword, just return the last statement

lib/PuppeteerSharp/BrowserContext.cs
```diff
@@ -122,21 +101,22 @@ public Task CloseAsync()
         /// ]]>
         /// </example>
         /// <seealso href="https://developer.mozilla.org/en-US/docs/Glossary/Origin"/>
-        public async Task OverridePermissionsAsync(string origin, IEnumerable<OverridePermission> permissions)
+        public Task OverridePermissionsAsync(string origin, IEnumerable<OverridePermission> permissions)
         {
-            var permissionList = permissions.Select(permission =>
+            var invalidPermission = permissions
```
Meir017: can `invalidPermission` actually have a value?

kblok: Yes here. https://github.com/kblok/puppeteer-sharp/pull/661/files#diff-89c4d2b1dab8fadf851afd2b813aff2dR42

I didn't ping you yet because I'm trying to find better ways to code that part but "it's working"

Meir017: I don't think we should care for that case.
you can always come up with weird cases, I think having it use an enum is enough

kblok:
Done @Meir017.

Let's leave the wait for implementation for another PR because there are more cycle events coming from Chromium than the ones listed there.

Issue #666, opened 2018-10-06
roll Chromium to r594312


kblok:
https://github.com/kblok/puppeteer/pull/12/commits/e75e36b9c2c0fc47b5bc97ff1f556e9ce1fa2e18

PR #686, opened 2018-10-08
bump the version of chromium

bdurrani:
this address #666 
Ran all the test locally as well.


Meir017:
@bdurrani Thanks for your contribution!

Issue #668, opened 2018-10-06
Make sure zero-width screenshots don't hang


kblok:
https://github.com/kblok/puppeteer/pull/12/commits/c644a3bbd3ee9722705a5a672a09828b9b90651b

bdurrani:
I'll take a stab at this one

kblok:
Go for it @bdurrani

bdurrani:
I'm getting a TargetClosedException with the following test

```
        public async Task ShouldNotHangWithZeroWidthHeightElement()
        {
            try
            {
            await Page.SetContentAsync(@"<div style='width: 0; height: 0'></div>");
            var elementHandle = await Page.QuerySelectorAsync("div");
            var screenshot = await elementHandle.ScreenshotDataAsync();
            Assert.NotNull(screenshot); 
            }
            catch(TargetClosedException ex)
            {
                ex.ToString();
            }
        }

```
The specific error is

`Message: PuppeteerSharp.TargetClosedException : Protocol error(Page.captureScreenshot): Target closed.`


I'm using Windows 10. I get the exception if either width or height are 0. 
Any thoughts on how to proceed?
According to[ this issue](https://github.com/GoogleChrome/puppeteer/issues/2672), looks like this is a known issue for linux

kblok:
@bdurrani push your PR so we can take a look.

bdurrani:
@kblok done!

PR #691, opened 2018-10-10
Add zero width element test

bdurrani:
Handles issue #668 
Fails currently a `TargetClosedException`


kblok:
Well, this is one of the "fun" issues. They created the test just to track the issue. It's expected to fail. When you see `xit` on Puppeteer, it means that the test is being ignored. Just ignore this test like this https://github.com/kblok/puppeteer-sharp/blob/7a8bd934180d1b5536c6a52bab5b15f588cae3b8/lib/PuppeteerSharp.Tests/PuppeteerTests/HeadfulTests.cs#L60

bdurrani:
Ah ok. sorry about that. I read the original issue and thought it was only a bug in linux. I just tried the nodejs version on windows and it does hang there as well, so it's a general problem. Thanks for taking a look at this.

PR #692, opened 2018-10-11
Clean up the main examples page

bdurrani:
- some of the examples were generate a "not a function" error. 
- docs made it look like `Browser` class has a static `NewPageAsync()` method. 
- could not find a `SetViewPort()` method. Looks like the correct method is `SetViewPortAsync()`


kblok:
Awesome!

PR #652, opened 2018-09-28
Remove unnecessary use of dynamics

tommymonk:
The use of dynamic throughout the code made it difficult to safely make a change relating to #650 
Dynamics are terrible for hiding refactor mistakes until run-time.

I've plumbed through JObject for response and JToken for method interface, this clarity allowed me to do a lot of streamlining of the response processing and in some places identify where we are serialize/de-serialize the same data many times.

I didn't want to add to the string debt so I pulled out a large amount of the key strings into a Consts class to help with quality and refactor and reduce risk of future mistake.

As discussed in #650 it seemed like a sensible idea to remove the 'dynamic' or 'object overloads of methods like EvaluateFunctionAsync which are a cause of confusion, however if someone wants to use the type 'object' they will now get the correct types serialized for numerics.

All of the existing tests are passing for me, I wanted to add extra tests to cover #650 but wasn't sure which test class to use, maybe you could guide me ?

PR #695, opened 2018-10-12
Fix bug affecting evaluation of string using Evaluate* methods

tommymonk:
Fixes bug relating to #652 caused by insufficient test of JToken when returning null

### DIFF
--- a/lib/PuppeteerSharp.Tests/PageTests/EvaluateTests.cs
+++ b/lib/PuppeteerSharp.Tests/PageTests/EvaluateTests.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
+using Newtonsoft.Json.Linq;
 using Xunit;
 using Xunit.Abstractions;
 
@@ -240,6 +241,39 @@ namespace PuppeteerSharp.Tests.PageTests
                 return executionContext.EvaluateFunctionAsync("() => null");
             });
             Assert.Contains("navigation", ex.Message);
+        }
+
+        [Fact]
+        public async Task ShouldWorkWithoutGenerics()
+        {
+            var objectEmpty = await Page.EvaluateExpressionAsync("{}");
+            var arrayEmpty = await Page.EvaluateExpressionAsync("[]");
+            var stringEmpty = await Page.EvaluateExpressionAsync("''");
+            
+            var objectPopulated = await Page.EvaluateExpressionAsync("var obj = {a:1}; obj;");
+            var arrayPopulated = await Page.EvaluateExpressionAsync("[1]");
+            var stringPopulated = await Page.EvaluateExpressionAsync("'1'");
+
+            var numberInt16Populated = await Page.EvaluateExpressionAsync("1");
+            var numberInt32Populated = await Page.EvaluateExpressionAsync("11111111");
+            var numberInt64Populated = await Page.EvaluateExpressionAsync("11111111111111");
+            var numberFloatPopulated = await Page.EvaluateExpressionAsync("1.1");
+
+            Assert.Null(objectEmpty);
+            Assert.Null(arrayEmpty);
+            Assert.Null(stringEmpty);
+
+            Assert.NotNull(objectPopulated);
+            Assert.Equal(1, objectPopulated["a"]);
+
+            Assert.IsType<JArray>(arrayPopulated);
+            Assert.Equal(1, ((JArray)arrayPopulated)[0]);
+
+            Assert.Equal("1", stringPopulated);
+            Assert.Equal(1, numberInt16Populated);
+            Assert.Equal(11111111, numberInt32Populated);
+            Assert.Equal(11111111111111, numberInt64Populated);
+            Assert.Equal(1.1, numberFloatPopulated);
         }
     }
 }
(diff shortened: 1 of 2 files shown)


lib/PuppeteerSharp.Tests/PageTests/EvaluateTests.cs
```diff
@@ -241,5 +242,38 @@ public async Task ShouldThrowANiceErrorAfterANavigation()
...
+            Assert.Equal(1, objectPopulated["a"]);
+
+            Assert.IsType<JArray>(arrayPopulated);
+            Assert.Equal(1, ((JArray)arrayPopulated)[0]);
+
+            Assert.Equal("1", stringPopulated);
+            Assert.Equal(1, numberInt16Populated);
+            Assert.Equal(11111111, numberInt32Populated);
+            Assert.Equal(11111111111111, numberInt64Populated);
+            Assert.Equal(1.1, numberFloatPopulated);
```
kblok: Though I know you are following the Act-Assert pattern, we are not using that pattern. So you could just simply do:
```cs
Assert.Null(await Page.EvaluateExpressionAsync("foo"));
```

lib/PuppeteerSharp.Tests/PageTests/EvaluateTests.cs
```diff
@@ -241,5 +242,38 @@ public async Task ShouldThrowANiceErrorAfterANavigation()
             });
             Assert.Contains("navigation", ex.Message);
         }
+
+        [Fact]
+        public async Task ShouldWorkWithoutGenerics()
+        {
+            var objectEmpty = await Page.EvaluateExpressionAsync("{}");
+            var arrayEmpty = await Page.EvaluateExpressionAsync("[]");
+            var stringEmpty = await Page.EvaluateExpressionAsync("''");
```
kblok: Can we support returning empty arrays and empty strings?

lib/PuppeteerSharp/ExecutionContext.cs
```diff
@@ -261,5 +261,11 @@ private static string GetExceptionMessage(EvaluateExceptionDetails exceptionDeta
             }
             return message;
         }
+
+        private static bool IsJTokenEmpty(JToken token) => (token == null) ||
+                   (token.Type == JTokenType.Array && !token.HasValues) ||
+                   (token.Type == JTokenType.Object && !token.HasValues) ||
+                   (token.Type == JTokenType.String && token.ToString() == string.Empty) ||
+                   (token.Type == JTokenType.Null);
```
kblok: I would format it like this:
```cs
private static bool IsJTokenEmpty(JToken token) 
    => (token == null) ||
        (token.Type == JTokenType.Array && !token.HasValues) ||
        (token.Type == JTokenType.Object && !token.HasValues) ||
        (token.Type == JTokenType.String && token.ToString() == string.Empty) ||
        (token.Type == JTokenType.Null);
```

lib/PuppeteerSharp.Tests/PageTests/EvaluateTests.cs
tommymonk: Yes, it would require a change to that new method, I am kinda on the fence as to which is the expected behavior of a user ?

tommymonk: There is some expectation in the unit tests already I believe

kblok: I think it would be more clear for the user if `Page.EvaluateExpressionAsync` is a `void/Task` method. If you call that you want to execute you don't want to return data.
If you want to return data use `Page.EvaluateExpressionAsync<T>` instead.

Thoughts @Meir017 ?

tommymonk: I think we discussed this in the original PR, if we wanted to add these overloads we would need to differentiate the names
In the interest of patching this bug, we should probably tackle that separately.

kblok: Fair enough, so let's avoid treating things that are not null as null: `{}`, `[]`, `''` should not return null.

tommymonk: I know that there are unit tests that expect empty object `{}` to be null, do you want them changing ?

kblok: Which one?

tommymonk: I was wrong, there are 3x tests that expect empty array to be null

`InputTests.ShouldClickOnCheckboxInputAndToggle`
`InputTests.ShouldClickOnCheckboxLabelAndToggle`
`EvaluateTests.ShouldWorkWithoutGenerics`

kblok: result.check is null https://github.com/kblok/puppeteer-sharp/blob/master/lib/PuppeteerSharp.TestServer/wwwroot/input/checkbox.html#L11 

EvaluateTests.ShouldWorkWithoutGenerics is yours :joy:

tommymonk: Ok, I've made the change, actually simplifies the change and removes the need for the method.

It does however mean that this becomes a breaking change we should mention in change notes.

kblok: Great Job @tommymonk !

PR #705, opened 2018-10-17
Introducing Burst mode

### LINKED ISSUE #703: Taking Multiple Successive screenshots
### Description

I am trying to take multiple screenshots of a web page for a duration of time every 200ms..
This is so I can have successive screenshots showing any animation/sideshow that may be on the page. Is seems the screenshot process is quite slow, is it possible to speed this up, given my code.

```
using System;
using System.IO;
using System.Threading;
using System.Threading.Tasks;
using PuppeteerSharp;
using Xunit;

namespace Tests
{
    public class ImageGeneratorTests
    {

        public static async Task<List<string>> ImagesFromUrl(string url, string tempFolder)
        {
            try
            {
                await new BrowserFetcher().DownloadAsync(BrowserFetcher.DefaultRevision);

                {
                    Headless = true
                });

                //page.Viewport.Width = 1366;

                var result = new List<string>();

                var duration = 20; // Duration in seconds
                var intervalTimeSpan = new TimeSpan(0, 0, 0, 0, 200); // 200 Milliseconds

                var screenShotOptions = new ScreenshotOptions
                {
                    FullPage = true,
                };
                
                // Load Url
                await page.GoToAsync(url);

                //await page.WaitForNavigationAsync(new NavigationOptions { WaitUntil = new[] { WaitUntilNavigation.Networkidle0 } });

                // Take Initial Screen Shot
                var outputFilePath1 = Path.Combine(tempFolder, $"{Guid.NewGuid()}.png");
                await page.ScreenshotAsync(outputFilePath1, screenShotOptions);
                result.Add(outputFilePath1);

                var cancellationTokenSource = new CancellationTokenSource();
                var cancellation = cancellationTokenSource.Token;

                var endTime = DateTime.MinValue;
                async Task TakeScreenShots(TimeSpan interval, CancellationToken cancellationToken)
                {
                    while (true)
                    {
                        if(cancellationToken.IsCancellationRequested)
                        {
                            break;
                        }

                        var outputFilePath = Path.Combine(tempFolder, $"{Guid.NewGuid()}_{DateTime.Now:yyyyMMddHHmmssFFF}.png");
                        await page.ScreenshotAsync(outputFilePath, screenShotOptions);
                        result.Add(outputFilePath);
                        
                        await Task.Delay(interval, cancellationToken);

                        if (DateTime.Now >= endTime)
                        {
                            cancellationTokenSource.Cancel();
                            Console.WriteLine("Cancelled");
                        }
                    }
                }

                var startTime = DateTime.Now;
                endTime = startTime.AddSeconds(duration);
                await TakeScreenShots(intervalTimeSpan, cancellation);

                return result;
            }
            catch (Exception exception)
            {
                Console.WriteLine(exception);
                return new List<string>();
            }
        }

        [Fact]
        public async void TakeScreenshots()
        {
            try
            {
                var screenShots = await ImagesFromUrl("https://www.take-a-view.co.uk", @"c:\temp");

                //AnimatedGifs.Create(screenShots);

                var passed = screenShots.Count > 0;

                Assert.True(passed, "Error generating screenshots");
            }
            catch (Exception exception)
            {
                Console.WriteLine(exception);
                throw;
            }
        }
    }
}
```

### Expected behaviour:

I would expect screenshots to happen quicker?

### Actual behaviour:

Taking screenshots every 200ms for 20 seconds creates 5 screenshots.

### Versions

PuppeteerSharp from Nuget version 1.8.0
.NET Core 2.1.

### Additional Information

At present I am writing these files to disk, I have just considered that saving the byte[] arrays to memory and saving later might speed things up marginally? But haven't tried this yet.

kblok:
`PerformScreenshot` can make up to 5 WebSocket calls before performing the screenshot:
 * Target.activateTarget
 * Page.getLayoutMetrics 
 * Page.getLayoutMetrics
 * Emulation.setDeviceMetricsOverride
 * Emulation.setDefaultBackgroundColorOverride

Then it calls `Page.captureScreenshot` and then it might need to call `Emulation.setDefaultBackgroundColorOverride` and `SetViewportAsync`.

This hurt the performance when trying to perform screenshots in "burst mode".
With "burst mode" I mean taking a series of screenshots to the same page during a certain period of time. 

With this new `BurstMode` we would perform those first five calls only on the first screenshot. Then we would call only `Page.captureScreenshot`. This will make this burst process at least 2.5x times faster.

After screenshots are taken, the user will need to call `SetBurstModeOff` explicitelly. So we run `Emulation.setDefaultBackgroundColorOverride` and `SetViewportAsync`.

closes #703

### DIFF
--- a/lib/PuppeteerSharp/Page.cs
+++ b/lib/PuppeteerSharp/Page.cs
@@ -1572,13 +1572,15 @@ namespace PuppeteerSharp
         /// Resets the background color and Viewport after taking Screenshots using BurstMode.
         /// </summary>
         /// <returns>The burst mode off.</returns>
-        public async Task SetBurstModeOff()
+        public Task SetBurstModeOff()
         {
             _screenshotBurstModeOn = false;
             if (_screenshotBurstModeOptions != null)
             {
-                await ResetBackgroundColorAndViewport(_screenshotBurstModeOptions).ConfigureAwait(false);
+                ResetBackgroundColorAndViewport(_screenshotBurstModeOptions);
             }
+
+            return Task.CompletedTask;
         }
         #endregion
 
--- a/lib/PuppeteerSharp/Page.cs
+++ b/lib/PuppeteerSharp/Page.cs
@@ -43,6 +43,9 @@ namespace PuppeteerSharp
         private readonly Dictionary<string, Worker> _workers;
         private readonly ILogger _logger;
         private bool _ensureNewDocumentNavigation;
+        private PageGetLayoutMetricsResponse _metrics;
+        private bool _screenshotBurstModeOn;
+        private ScreenshotOptions _screenshotBurstModeOptions;
 
         private static readonly Dictionary<string, decimal> _unitToPixels = new Dictionary<string, decimal> {
             {"px", 1},
@@ -1565,6 +1568,18 @@ namespace PuppeteerSharp
         /// <param name="options">Navigation parameters.</param>
         public Task<Response> GoForwardAsync(NavigationOptions options = null) => GoAsync(1, options);
 
+        /// <summary>
+        /// Resets the background color and Viewport after taking Screenshots using BurstMode.
+        /// </summary>
+        /// <returns>The burst mode off.</returns>
+        public async Task SetBurstModeOff()
+        {
+            _screenshotBurstModeOn = false;
+            if (_screenshotBurstModeOptions != null)
+            {
+                await ResetBackgroundColorAndViewport(_screenshotBurstModeOptions);
+            }
+        }
         #endregion
 
         #region Private Method
@@ -1645,10 +1660,13 @@ namespace PuppeteerSharp
 
         private async Task<string> PerformScreenshot(ScreenshotType type, ScreenshotOptions options)
         {
-            await Client.SendAsync("Target.activateTarget", new
+            if (!_screenshotBurstModeOn)
             {
-                targetId = Target.TargetId
-            }).ConfigureAwait(false);
+                await Client.SendAsync("Target.activateTarget", new
+                {
+                    targetId = Target.TargetId
+                }).ConfigureAwait(false);
+            }
 
             var clip = options.Clip?.Clone();
             if (clip != null)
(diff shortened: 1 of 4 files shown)


Meir017:
@kblok I'm not completely sure how this burst feature works, can you briefly explain it?

lib/PuppeteerSharp/Page.cs
```diff
@@ -43,6 +43,9 @@ public class Page : IDisposable
         private readonly Dictionary<string, Worker> _workers;
         private readonly ILogger _logger;
         private bool _ensureNewDocumentNavigation;
+        private PageGetLayoutMetricsResponse _metrics;
```
Meir017: I think this should be renamed to `_burstModeMetrics` or similar
```suggestion
        private PageGetLayoutMetricsResponse _burstModeMetrics;
```

lib/PuppeteerSharp/Page.cs
```diff
@@ -1727,19 +1756,37 @@ private async Task<string> PerformScreenshot(ScreenshotType type, ScreenshotOpti
...
+            else
+            {
+                await ResetBackgroundColorAndViewport(options);
+            }
+            return result.Data;
+        }
+
+        private async Task ResetBackgroundColorAndViewport(ScreenshotOptions options)
+        {
+            var tasks = new List<Task>();
```
Meir017: maybe create two task variable with a `Task.Completed` value and then call 
```cs
var omitBackgroundTask = options?.OmitBackground == true ? 
    Client.SendAsync("Emulation.setDefaultBackgroundColorOverride") : Task.Completed;
var setViewPortTask = (options?.FullPage == true && Viewport != null) ?
    SetViewportAsync(Viewport) : Task.Completed;
return Task.WhenAll(omitBackgroundTask, setViewPortTask);
```

kblok:
Description updated @Meir017

Meir017:
just a few changes

kblok:
done @Meir017

brightertools:
Hi,
Will we be able to set the duration and interval in burst mode?
Thanks

Meir017:
You can achieve this using a loop or similar, I don't think this behavior should be baked in

brightertools:
Oh I see, I get it now..
We switch bust mode on, then take the screenshots on our duration/interval, then switch it off.
👍

PR #727, opened 2018-10-31
[WIP] CancellationTokenSource implementation

### LINKED ISSUE #709: Use CancellationTokenSource.CancelAfter instead of a Timeout task
We could use [CancellationTokenSource.CancelAfter](https://docs.microsoft.com/en-us/dotnet/api/system.threading.cancellationtokensource.cancelafter?view=netframework-4.7.2) to timeout tasks

nessross77:
closes #709

### DIFF
--- a/lib/PuppeteerSharp/FrameManager.cs
+++ b/lib/PuppeteerSharp/FrameManager.cs
@@ -67,25 +67,21 @@ namespace PuppeteerSharp
             var watcher = new NavigatorWatcher(_client, this, frame, _networkManager, timeout, options);
 
             var navigateTask = NavigateAsync(_client, url, referrer, frame.Id);
-            await Task.WhenAny(
-                watcher.TimeoutOrTerminationTask,
-                navigateTask).ConfigureAwait(false);
+            await Task.Run(() => navigateTask, watcher.TimedCancellationToken).ConfigureAwait(false);
 
-            AggregateException exception = null;
+            Exception exception = null;
             if (navigateTask.IsFaulted)
             {
                 exception = navigateTask.Exception;
             }
             else
             {
-                await Task.WhenAny(
-                    watcher.TimeoutOrTerminationTask,
-                    _ensureNewDocumentNavigation ? watcher.NewDocumentNavigationTask : watcher.SameDocumentNavigationTask
-                ).ConfigureAwait(false);
 
-                if (watcher.TimeoutOrTerminationTask.IsCompleted && watcher.TimeoutOrTerminationTask.Result.IsFaulted)
+                await Task.Run(() => _ensureNewDocumentNavigation ? watcher.NewDocumentNavigationTask : watcher.SameDocumentNavigationTask, watcher.TimedCancellationToken).ConfigureAwait(false);
+
+                if (watcher.TimedCancellationToken.IsCancellationRequested)
                 {
-                    exception = watcher.TimeoutOrTerminationTask.Result.Exception;
+                    exception = new TaskCanceledException("Timed cancellation token timed out.");
                 }
             }
 
@@ -119,18 +115,12 @@ namespace PuppeteerSharp
             var timeout = options?.Timeout ?? DefaultNavigationTimeout;
             var watcher = new NavigatorWatcher(_client, this, frame, _networkManager, timeout, options);
 
-            var raceTask = await Task.WhenAny(
-                watcher.NewDocumentNavigationTask,
-                watcher.SameDocumentNavigationTask,
-                watcher.TimeoutOrTerminationTask
-            ).ConfigureAwait(false);
+            var raceTask = await Task.Run(() => Task.WhenAny(watcher.NewDocumentNavigationTask, watcher.SameDocumentNavigationTask), watcher.TimedCancellationToken).ConfigureAwait(false);
 
-            var exception = raceTask.Exception;
-            if (exception == null &&
-                watcher.TimeoutOrTerminationTask.IsCompleted &&
-                watcher.TimeoutOrTerminationTask.Result.IsFaulted)
+            Exception exception = raceTask.Exception;
+            if (exception == null && watcher.TimedCancellationToken.IsCancellationRequested)
             {
-                exception = watcher.TimeoutOrTerminationTask.Result.Exception;
+                exception = new TaskCanceledException("Timed cancellation token timed out.");
             }
             if (exception != null)
             {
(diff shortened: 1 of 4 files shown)


lib/PuppeteerSharp/FrameManager.cs
```diff
@@ -67,25 +67,21 @@ public async Task<Response> NavigateFrameAsync(Frame frame, string url, Navigati
             var watcher = new NavigatorWatcher(_client, this, frame, _networkManager, timeout, options);
 
             var navigateTask = NavigateAsync(_client, url, referrer, frame.Id);
-            await Task.WhenAny(
-                watcher.TimeoutOrTerminationTask,
-                navigateTask).ConfigureAwait(false);
+            await Task.Run(() => navigateTask, watcher.TimedCancellationToken).ConfigureAwait(false);
```
kblok: Clever 😏

lib/PuppeteerSharp/FrameManager.cs
```diff
@@ -67,25 +67,21 @@ public async Task<Response> NavigateFrameAsync(Frame frame, string url, Navigati
...
             }
             else
             {
-                await Task.WhenAny(
-                    watcher.TimeoutOrTerminationTask,
-                    _ensureNewDocumentNavigation ? watcher.NewDocumentNavigationTask : watcher.SameDocumentNavigationTask
-                ).ConfigureAwait(false);
 
-                if (watcher.TimeoutOrTerminationTask.IsCompleted && watcher.TimeoutOrTerminationTask.Result.IsFaulted)
+                await Task.Run(() => _ensureNewDocumentNavigation ? watcher.NewDocumentNavigationTask : watcher.SameDocumentNavigationTask, watcher.TimedCancellationToken).ConfigureAwait(false);
```
kblok: Maybe we can move all this ` _ensureNewDocumentNavigation ? watcher.NewDocumentNavigationTask : watcher.SameDocumentNavigationTask` to a variable, so it's more clear.

lib/PuppeteerSharp/Helpers/CancellationTokenSourceHelper.cs
```diff
@@ -0,0 +1,32 @@
...
+            {
+                cancellationTokenSource.CancelAfter(-1);
+            }
+            else
+            {
+                cancellationTokenSource.CancelAfter(timeout);
+            }
+
+            var cancellationToken = cancellationTokenSource.Token;
+            return cancellationToken;
```
kblok: We can just return `cancellationTokenSource.Token;`

lib/PuppeteerSharp/Helpers/CancellationTokenSourceHelper.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// <summary>
+        /// Creates a CancellationToken with a timeout.
+        /// </summary>
+        /// <param name="timeout">Timeout in milliseconds.</param>
+        /// <returns>The timed cancellation token.</returns>
+        public static CancellationToken GetCancellationToken(int timeout)
+        {
+            var cancellationTokenSource = new CancellationTokenSource();
+
+            if (timeout == 0)
```
kblok: Can't we do only this?
```cs
if (timeout > 0)
{
    cancellationTokenSource.CancelAfter(timeout);
}
```

lib/PuppeteerSharp/NavigatorWatcher.cs
```diff
@@ -77,7 +78,7 @@ internal class NavigatorWatcher
         public Task<bool> SameDocumentNavigationTask => _sameDocumentNavigationTaskWrapper.Task;
         public Task<bool> NewDocumentNavigationTask => _newDocumentNavigationTaskWrapper.Task;
         public Response NavigationResponse => _navigationRequest?.Response;
-        public Task<Task> TimeoutOrTerminationTask => Task.WhenAny(_timeoutTask, _terminationTaskWrapper.Task);
+        public CancellationToken TimedCancellationToken => CancellationTokenSourceHelper.GetCancellationToken(_timeout);
```
kblok: Let's call it `TimeoutCancellationToken`

kblok: We should trigger the cancellation token here https://github.com/kblok/puppeteer-sharp/blob/a890badffa63bb8dca833e8b95005e06ed54d7ba/lib/PuppeteerSharp/NavigatorWatcher.cs#L120

lib/PuppeteerSharp/FrameManager.cs
```diff
@@ -67,25 +67,21 @@ public async Task<Response> NavigateFrameAsync(Frame frame, string url, Navigati
...
+
+                if (watcher.TimedCancellationToken.IsCancellationRequested)
                 {
-                    exception = watcher.TimeoutOrTerminationTask.Result.Exception;
+                    exception = new TaskCanceledException("Timed cancellation token timed out.");
```
kblok: We can throw this exception https://github.com/kblok/puppeteer-sharp/blob/a890badffa63bb8dca833e8b95005e06ed54d7ba/lib/PuppeteerSharp/Helpers/TaskHelper.cs#L28 so we don't break compatibility

kblok:
@nessross77 also take a look at CodeFactor errors. And also thank you for stepping in :)

kblok:
AppVeyor got stuck

lib/PuppeteerSharp/Helpers/CancellationTokenSourceHelper.cs
```diff
@@ -0,0 +1,32 @@
...
+    /// CancellationTokenSource Helper.
+    /// </summary>
+    public static class CancellationTokenSourceHelper
+    {
+        /// <summary>
+        /// Creates a CancellationToken with a timeout.
+        /// </summary>
+        /// <param name="timeout">Timeout in milliseconds.</param>
+        /// <returns>The timed cancellation token.</returns>
+        public static CancellationToken GetCancellationToken(int timeout)
```
Meir017: Maybe rename to something like `GetTimingOutCancellationToken` ? Or rename the class to `TimingOutCancellationToken` and then name the method just `Get`

Meir017:
Do we need the `Task.Run` calls?

kblok:
I'm afraid this is not going to work. The cancellation token is a cooperative cancellation. We need to do something with that token or send that token to a lower level API.

kblok:
Closed due to inactivity

Issue #722, opened 2018-10-29
How to map complex javascript objects to C#


kblok:
https://github.com/kblok/puppeteer-sharp/issues/693

bdurrani:
Working on this one next

Meir017:
just for reference:
inside of a test class:
```cs
[Fact]
public async Task ComplexObject()
{
    using (var browser = await Puppeteer.LaunchAsync(new LaunchOptions()))
    using (var page = await browser.NewPageAsync())
    { 
        var obj = await page.EvaluateFunctionAsync<MyComplexObject>(@"() => {
            return { FirstName: 'Meir', LastName: 'Blachman', Age: 24, Nested: { A: 'a value', B: 'b value' } };
        }");
        Console.WriteLine("test");
    }
}

class MyComplexObject
{
    public string FirstName { get; set; }
    public string LastName { get; set; }
    public int Age { get; set; }
    
    public MyNestedComplexObject Nested { get; set; }
}

class MyNestedComplexObject
{
    public string A { get; set; }
    public string B { get; set; }
}
```

Meir017:
@kblok maybe we should look into allowing the user to use `camelcase` for serializing?

```cs
JsonConvert.SerializeObject(json, new JsonSerializerSettings
{
    ContractResolver = new Serialization.CamelCasePropertyNamesContractResolver()
});
```

kblok:
@Meir017 agreed but not for this issue. I'll create a new one for that.

bdurrani:
Thanks @Meir017 for the example. I've linked what I have so far, but I like the way you did it. 
I'll change my example to match yours.

bdurrani:
@kblok you can close this issue. Thanks

PR #737, opened 2018-11-06
Example/complex js

bdurrani:
#722 

Example of mapping complex js object


bdurrani:
@kblok ready for review

Issue #717, opened 2018-10-26
Unrecoverable exceptions being frequently thrown


lloydjatkinson:
### Description

Unrecoverable exceptions are being thrown by Puppeteer Sharp. This prevents any kind of retry or graceful exception handling and results in the process being terminated.

### Complete minimal example reproducing the issue

```csharp
    // Arrange
            var launchOptions = new LaunchOptions()
            {
                Headless = true,
            };

            var viewPortOptions = new ViewPortOptions()
            {
                Width = 1920,
                Height = 1080
            };

            var sites = new List<string>()
            {
                "wix.com",
                "facebook.com",
                "twitter.com",
                "google.com",
                "youtube.com",
                "instagram.com",
                "linkedin.com",
                "wordpress.org",
                "pinterest.com",
                "wikipedia.org",
                "wordpress.com",
                "blogspot.com",
                "apple.com",
                "adobe.com",
                "tumblr.com",
            }
    // Act
            try
            {
                await new BrowserFetcher().DownloadAsync(BrowserFetcher.DefaultRevision);
                using (var browser = await Puppeteer.LaunchAsync(launchOptions))
                {
                    //var context = await browser.CreateIncognitoBrowserContextAsync();
                    //var page = await context.NewPageAsync();
                    var page = await browser.NewPageAsync();
                    await page.SetViewportAsync(viewPortOptions);

                    foreach (var site in sites)
                    {
                        try
                        {
                            await page.GoToAsync($"http://{site}");
                            Console.WriteLine(await page.GetTitleAsync());
                            await page.ScreenshotAsync($"D:\\bin\\screenshots\\{site}.png");
                        }
                        catch (Exception exception)
                        {
                            // Catches most exceptions such as timeouts but does not catch others, see below.
                            Console.WriteLine($"Unable to take screenshot of: {site}. Exception: {exception.Message}");
                        }
                    }
                }
            }
            catch (Exception exception)
            {
                // Never enters the catch.
                Console.WriteLine($"Unable to proceed: {exception.Message}");
            }
```

### Expected behavior:

The library to take a screenshot of the URL given.

### Actual behavior:

Often crashes with exceptions which cannot be caught. There appears to be no common cause of the issue and it happens at random. The issue does not seem to be related to a particular site.

Same happens with normal and Incognito sessions.

### Versions

* Which version of PuppeteerSharp are you using? 1.8.0
* Which .NET runtime and version are you targeting? .NET Core 2.1

### Additional Information

[...]

PR #753, opened 2018-11-11
Improve internal exception handling

### LINKED ISSUE #752: Improve Exception handle on message processing
Related to #717

kblok:
If we fail to process a message coming from chromium we can't just fail trusting that the user will catch that exception, because that call won't not part of the call stack coming from the user but from the WebSocket listener. This was causing the [unrecoverable exceptions](https://github.com/kblok/puppeteer-sharp/issues/717).

In order to fix this, we are going to close the `IConnection` (`CDPSession` or `Connection`) if we fail to process a message. We will add that exception as a `CloseReason` to the `IConnection` and also to the `TargetClosedException` in order to help us and the user to have more information about the problem we might have.

closes #752

### DIFF
--- a/lib/PuppeteerSharp.Tests/BrowserTests/Events/DisconnectedTests.cs
+++ b/lib/PuppeteerSharp.Tests/BrowserTests/Events/DisconnectedTests.cs
@@ -90,6 +90,7 @@ namespace PuppeteerSharp.Tests.BrowserTests.Events
                 //Whether from the Connection rejecting a message from the CDPSession 
                 //or from the CDPSession trying to send a message to a closed connection
                 Assert.IsType<TargetClosedException>(exception.InnerException);
+                Assert.Equal("Connection disposed", ((TargetClosedException)exception.InnerException).CloseReason);
             }
         }
     }
--- a/lib/PuppeteerSharp.Tests/PageTests/CloseTests.cs
+++ b/lib/PuppeteerSharp.Tests/PageTests/CloseTests.cs
@@ -13,13 +13,12 @@ namespace PuppeteerSharp.Tests.PageTests
         public async Task ShouldRejectAllPromisesWhenPageIsClosed()
         {
             var neverResolves = Page.EvaluateFunctionAsync("() => new Promise(r => {})");
-
-            // Put into a var to avoid warning
-            var t = Page.CloseAsync();
+            _ = Page.CloseAsync();
 
             var exception = await Assert.ThrowsAsync<EvaluationFailedException>(async () => await neverResolves);
             Assert.IsType<TargetClosedException>(exception.InnerException);
             Assert.Contains("Protocol error", exception.Message);
+            Assert.Equal("Target.detachedFromTarget", ((TargetClosedException)exception.InnerException).CloseReason);
         }
 
         [Fact]
--- a/lib/PuppeteerSharp/Browser.cs
+++ b/lib/PuppeteerSharp/Browser.cs
@@ -330,19 +330,28 @@ namespace PuppeteerSharp
 
         private async void Connect_MessageReceived(object sender, MessageEventArgs e)
         {
-            switch (e.MessageID)
+            try
             {
-                case "Target.targetCreated":
-                    await CreateTargetAsync(e.MessageData.ToObject<TargetCreatedResponse>()).ConfigureAwait(false);
-                    return;
+                switch (e.MessageID)
+                {
+                    case "Target.targetCreated":
+                        await CreateTargetAsync(e.MessageData.ToObject<TargetCreatedResponse>()).ConfigureAwait(false);
+                        return;
 
-                case "Target.targetDestroyed":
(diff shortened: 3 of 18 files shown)


lib/PuppeteerSharp.Tests/BrowserTests/Events/DisconnectedTests.cs
```diff
@@ -90,6 +90,7 @@ public async Task ShouldRejectWaitForSelectorWhenBrowserCloses()
                 //Whether from the Connection rejecting a message from the CDPSession 
                 //or from the CDPSession trying to send a message to a closed connection
                 Assert.IsType<TargetClosedException>(exception.InnerException);
+                Assert.Equal("Connection disposed", ((TargetClosedException)exception.InnerException).CloseReason);
```
Meir017: maybe write something like this?
```cs
Assert.Equal("Connection disposed", Assert.IsType<TargetClosedException>(exception.InnerException).CloseReason);
```

lib/PuppeteerSharp.Tests/PageTests/CloseTests.cs
```diff
@@ -13,13 +13,12 @@ public class CloseTests : PuppeteerPageBaseTest
...
             var neverResolves = Page.EvaluateFunctionAsync("() => new Promise(r => {})");
-
-            // Put into a var to avoid warning
-            var t = Page.CloseAsync();
+            _ = Page.CloseAsync();
 
             var exception = await Assert.ThrowsAsync<EvaluationFailedException>(async () => await neverResolves);
             Assert.IsType<TargetClosedException>(exception.InnerException);
             Assert.Contains("Protocol error", exception.Message);
+            Assert.Equal("Target.detachedFromTarget", ((TargetClosedException)exception.InnerException).CloseReason);
```
Meir017: same here

lib/PuppeteerSharp/Browser.cs
```diff
@@ -330,19 +330,28 @@ internal async Task DisposeContextAsync(string contextId)
...
 
-                case "Target.targetInfoChanged":
-                    ChangeTargetInfo(e.MessageData.ToObject<TargetCreatedResponse>());
-                    return;
+                    case "Target.targetInfoChanged":
+                        ChangeTargetInfo(e.MessageData.ToObject<TargetCreatedResponse>());
+                        return;
+                }
+            }
+            catch (Exception ex)
```
Meir017: maybe we can have a utility method somewhere that will contain the switch case? we have this in each XXX_MessageReceived method

something like this:
```cs
public static void OnMessageReceived(IReadOnlyDictionary<string, Action<JToken>> messagesHandler, MessageEventArgs e)
{
    try
    {
        if (messagesHandler.TryGetValue(e.MessageID, out var handler))
        {
            handler?.Invoke(e.MessageData);
        }
    }
    catch (Exception ex)
    {
        var message = $"Browser failed to process {e.MessageID}. {ex.Message}. {ex.StackTrace}";
        _logger.LogError(ex, message);
        Connection.Close(message);
        throw;
    }
}

// usage:
OnMessageReceived(new Dictionary<string, Action<JToken>>
{
    ["Target.targetCreated"] = async data => await CreateTargetAsync(e.MessageData.ToObject<TargetCreatedResponse>()).ConfigureAwait(false),
    ["Target.targetDestroyed"] = async data => await DestroyTargetAsync(e.MessageData.ToObject<TargetDestroyedResponse>()).ConfigureAwait(false),
    ["Target.targetInfoChanged"] = data => ChangeTargetInfo(e.MessageData.ToObject<TargetCreatedResponse>()),
}, e);
```

Meir017:
how about creating a base class like this?
```cs
using System;
using Microsoft.Extensions.Logging;
using Newtonsoft.Json.Linq;

namespace PuppeteerSharp
{
    public abstract class MessageReceivedHandler
    {
        protected readonly ILogger Logger;
        protected readonly IConnection Connection;

        protected MessageReceivedHandler(IConnection connection, ILogger logger)
        {
            Connection = connection;
            Logger = logger;
        }

        protected void OnMessageReceived(IReadOnlyDictionary<string, Action<JToken>> messagesHandler, MessageEventArgs e)
        {
            try
            {
                if (messagesHandler.TryGetValue(e.MessageID, out var handler))
                {
                    handler?.Invoke(e.MessageData);
                }
            }
            catch (Exception ex)
            {
                var message = $"Browser failed to process {e.MessageID}. {ex.Message}. {ex.StackTrace}";
                Logger.LogError(ex, message);
                Connection.Close(message);
                throw;
            }
        }
    }
}
```

lib/PuppeteerSharp.Tests/BrowserTests/Events/DisconnectedTests.cs
kblok: IMHO I don't like the idea of having to asserts in the same line.

kblok:
@Meir017 how about getting rid of events and doing that in the `IConnection` itself?

_client.On("Target.targetCreated", async data => await CreateTargetAsync(e.MessageData.ToObject<TargetCreatedResponse>()).ConfigureAwait(false));

PR #763, opened 2018-11-17
Snapshot the accessibility tree

### LINKED ISSUE #747: Snapshot the accessibility tree
https://github.com/kblok/puppeteer/pull/13/commits/9ba3261571f3cb636f83cfd367f8ceea7e6e3588

kblok:
closes #747


lib/PuppeteerSharp/PageAccessibility/AXNode.cs
```diff
@@ -0,0 +1,255 @@
...
+namespace PuppeteerSharp.PageAccessibility
+{
+    internal class AXNode
+    {
+        internal AccessibilityGetFullAXTreeResponse.AXTreeNode Payload { get; }
+        public List<AXNode> Children { get; }
+        public bool Focusable { get; set; };
+
+        private readonly string _name;
+        private string _role;
```
Meir017: could this be an enum?

kblok: It's not clear what's the complete list of roles. We had some headaches in the past because our enums were incomplete.
We could use the official list https://www.w3.org/1999/xhtml/vocab. But we have items like `doc-cover` which are not there.

Issue #736, opened 2018-11-05
browser.NewPageAsync() dies in IIS


taz4270:
### browser.NewPageAsync() dies in IIS same problem as #375 
Hello again, today I found a issue that was asked before on #375, the `browser.NewPageAsync` is not returning anything, don't know if there is an argument missing (probably not), or there was something that wasn't right at the _master code_, so i tried to find a solution, i've kinda made it work but unfortunatly, i can't find how can i open the other ISS' port (something like that but this part doesn't matter, just know that it worked), and when i tried to migrate the code the other side, it blow up
### There is some part of my code
**Remember, Im using my addon, so any method missing `Async`, is my doing**
```csharp
private Browser GetBrowser()
{
        loggerFactory = new LoggerFactory();
        return PuppeteerEx.Launch(new LaunchOptions()
        {
            Headless = true,
            ExecutablePath = _ExecutionPath_, //It goes to the HostingEnvirounment
            LogProcess = true,
        },
        loggerFactory;
        );
        private Page GetNavigatedPage(Browser browser)
        {
            Page page = browser.NewPage();
            page.GoTo(Link, 30000);
            return page;
        }
        catch (Exception e)
        {
            Debug.WriteLine(e);
            throw e.GetBaseException();
        }

       private Page GetPage()
       { 
            var browser = GetBrowser();
            var Navigate = GetNavigatedPage(browser);
            return Navigate;
       }
}
```

taz4270:
There is something inside of `Browser.CreatePageInContextAsync` or a property missing, that doesn't allow the program the program proceed

taz4270:
### Something is not completly right
Because I've been searching your code inside-out and even tried to find a result,
I tried to make a method to find how deep i could go to find more precise as I could:
```csharp
private static async Task<Page> SepareteContext(Browser browser){
        var browserContextArray = browser.BrowserContexts();
        var browserContext = browserContextArray.LastOrDefault();
        if (browserContext is null)
            return null;
        var page = await browserContext.NewPageAsync(); //Even tried with .ConfigureAwait(false)
        return page;
}
```
The bug occurrs when page is awaiting for `browserContext.NewPageAsync()`, it never has the anwser back

kblok:
What is `NewPage` doing?

taz4270:
Sort of, it's a "weird specie of a bird", That is synchronous but it waits for Tasks to be executed

kblok:
[...]

PR #776, opened 2018-11-27
AspNet WebSocket Transport

kblok:
Based on #775 
Related to #736  

As it's not recommended to use `Task.Run` in ASP.NET Full framework, we are creating a new package (.NET Full Framework 4.7.2) called PuppeteerSharp.AspNetFramework.

This package has a class named `AspNetWebSocketTransport` which uses `HostingEnvironment.QueueBackgroundWorkItem` instead of `Task.Run`.


kblok:
@Meir017 would you mind taking a look at this guy? I really like to have your feedback here.

I had to change a few things on our build process because we have not a .NET Framework library in our solution.

kblok:
Not ready to be reviewed, I will create another solution and AppVeyor project because it's breaking my Mac setup

Jojoshua:
I noticed this was mentioned in the December 18 update but it's not yet merged right?

kblok:
@Jojoshua we will ship it with v1.11 next week :)

Jojoshua:
Hey I saw this was merged. I just updated to 1.11.1 but I am getting ....

The type or namespace name 'AspNetWebSocketTransport' could not be found (are you missing a using directive or an assembly reference?)	

Any ideas?

kblok:
@Jojoshua you will need to add this package https://www.nuget.org/packages/PuppeteerSharp.AspNetFramework/

Jojoshua:
Ahh I missed that. Thanks!

PR #792, opened 2018-12-07
Use SendAsync when in asynchronous context

Im5tu:


kblok:
Any reason for this change @Im5tu?

[Send](https://github.com/kblok/puppeteer-sharp/blob/master/lib/PuppeteerSharp/CDPSession.cs#L109) is more a fire-and-forget method than a Synchronous one. See #690

PR #793, opened 2018-12-07
Fixed event handler leak

Im5tu:
In a scenario where a session is kept open for a pro-longed period with multiple navigations, then the event handlers can be kept in memory causing a prolonged memory leak.

This PR wraps the `NavigationManager` in a `using` statement and cleans up the event handlers on `Dispose`


Im5tu:
@kblok added an additional event handler removal, originally found by @tommymonk

kblok:
Great job @Im5tu!

PR #794, opened 2018-12-07
Removed couple of common allocations

Im5tu:
- Removed unused `TaskCompletitionSource`
- Used static array instead of creating one on each navigation


kblok:
Thank you for your contributions @Im5tu! 🎉

PR #807, opened 2018-12-18
Improve Headful test

kblok:


lib/PuppeteerSharp.Tests/FrameTests/GoToTests.cs
```diff
@@ -76,9 +76,10 @@ public async Task ShouldReturnMatchingResponses()
             for (var i = 0; i < 3; ++i)
             {
                 serverResponses[i].TrySetResult(serverResponseTexts[i]);
-                var response = await navigations[i];
+                var response = await navigations[i];
+                var responseText = await response.TextAsync();
                 Assert.Same(frameTasks[i].Result, response.Frame);
-                Assert.Equal(serverResponseTexts[i], await response.TextAsync());
+                Assert.Equal(serverResponseTexts[i], responseText);
```
kblok: This might not make sense, but it makes Linux builds more stable

PR #808, opened 2018-12-19
Browser.IsClosed should be true when the connection is closed

### LINKED ISSUE #749: Browser.IsClosed is false even if Chromium process has quit
### Description

When Chromium quits by accident (e.g. memory allocation failed), `Browser.IsClosed` is still false.

Complete means the code snippet can be copied into a unit test method in a fresh C# project and run.
Minimal means it is stripped from code not related to reproducing the issue.

E.g.

```csharp
            {
                Headless = true,
                DumpIO = true,
            });
var context = await browser.CreateIncognitoBrowserContextAsync();
var page = await context.NewPageAsync();
await page.GoToAsync(
                    "https://www.google.com",
                    300000,
                    new[] {WaitUntilNavigation.Networkidle0}
                );

// Do something to make system out of memory
// e.g. limit the control group via systemd, run some other memory-consuming program or just open a extra large page

// Then make Chromium allocate some memory
await page.ScreenshotAsync("test.png"); 
// Chromium: [1109/040905.664281:FATAL:memory_linux.cc(42)] Out of memory.

Console.WriteLine(browser.IsClosed); // false
// then send any command to the browser
await page.ScreenshotAsync("test.png"); 
// raises PuppeteerSharp.MessageException: Protocol error(Target.sendMessageToTarget): Target closed. ---> PuppeteerSharp.TargetClosedException: Protocol error(Target.sendMessageToTarget): Target closed.
```

### Versions

* PuppeteerSharp 1.9.0
* .Net Core 2.1

kblok:
Closes #749

We are also removing the CloseMe project. we can accomplish the same by launching a chrome instance.


Meir017:
@kblok some codefactor stuff...

PR #816, opened 2018-12-23
Fix race conditions on Worker test

kblok:
This would fix this error https://ci.appveyor.com/project/kblok/puppeteer-sharp/builds/21197746/job/p57wdynrjh47xnua#L265

### DIFF
--- a/lib/PuppeteerSharp.Tests/WorkerTests/WorkerTests.cs
+++ b/lib/PuppeteerSharp.Tests/WorkerTests/WorkerTests.cs
@@ -1,6 +1,7 @@
 using System.Threading.Tasks;
 using Xunit;
 using Xunit.Abstractions;
+using PuppeteerSharp.Helpers;
 
 namespace PuppeteerSharp.Tests.WorkerTests
 {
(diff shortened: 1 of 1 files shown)


lib/PuppeteerSharp.Tests/WorkerTests/WorkerTests.cs
```diff
@@ -14,17 +15,22 @@ public WorkerTests(ITestOutputHelper output) : base(output)
...
-                    Page.GoToAsync(TestConstants.ServerUrl + "/worker/worker.html"));
+                workerCreatedTcs.Task,
+                Page.GoToAsync(TestConstants.ServerUrl + "/worker/worker.html"));
             var worker = Page.Workers[0];
             Assert.Contains("worker.js", worker.Url);
 
             Assert.Equal("worker function result", await worker.EvaluateExpressionAsync<string>("self.workerFunction()"));
 
             await Page.GoToAsync(TestConstants.EmptyPage);
+            await workerDestroyedTcs.Task.WithTimeout(1000);
```
Meir017: why the timeout? what will happen if we just `await`?

kblok: If we fail because the worker is never destroyed I'd rather get an invalid worker on the next line than hanging the CI because the task never completes.

kblok: Sorry, We wouldn't get to the next line because we would get a TimeoutException, but I'd rather get that than hanging the CI.

PR #818, opened 2018-12-23
Improve WithConnectionCheck even handling

kblok:


kblok:
What do you think @Meir017, @Im5tu?

@Im5tu I can't wrap that in a try-catch because it would silence the exception we're throwing.

PR #820, opened 2018-12-24
Improve WithConnectionCheck extension method

kblok:


Im5tu:
:shipit: 

Event handlers can be evil :(

PR #823, opened 2018-12-27
We don't need no dynamics

### LINKED ISSUE #97: Use strongly typed objects instead of dynamic everywhere

kblok:
closes #97

### DIFF
--- a/lib/PuppeteerSharp/ExecutionContext.cs
+++ b/lib/PuppeteerSharp/ExecutionContext.cs
@@ -6,6 +6,7 @@ using System.Threading.Tasks;
 using Newtonsoft.Json.Linq;
 using PuppeteerSharp.Messaging;
 using PuppeteerSharp.Helpers;
+using static PuppeteerSharp.Messaging.RuntimeQueryObjectsResponse;
 
 namespace PuppeteerSharp
 {
@@ -116,12 +117,12 @@ namespace PuppeteerSharp
                 throw new PuppeteerException("Prototype JSHandle must not be referencing primitive value");
             }
 
-            var response = await _client.SendAsync("Runtime.queryObjects", new Dictionary<string, object>
+            var response = await _client.SendAsync<RuntimeQueryObjectsResponse>("Runtime.queryObjects", new Dictionary<string, object>
             {
                 {"prototypeObjectId", objectId.ToString()}
             }).ConfigureAwait(false);
 
-            return CreateJSHandle(response[MessageKeys.Objects]);
+            return CreateJSHandle(response.Objects);
         }
 
         internal async Task<JSHandle> EvaluateExpressionHandleAsync(string script)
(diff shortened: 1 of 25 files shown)


lib/PuppeteerSharp/ExecutionContext.cs
```diff
@@ -173,8 +174,8 @@ internal async Task<JSHandle> EvaluateFunctionHandleAsync(string script, params
             }
         }
 
-        internal JSHandle CreateJSHandle(dynamic remoteObject)
-            => (remoteObject.subtype == "node" && Frame != null)
+        internal JSHandle CreateJSHandle(JToken remoteObject)
```
Meir017: do we want to create a class for this?

lib/PuppeteerSharp/Messaging/EvaluateHandleResponse.cs
```diff
@@ -0,0 +1,10 @@
+﻿using Newtonsoft.Json.Linq;
+
+namespace PuppeteerSharp.Messaging
+{
+    internal class EvaluateHandleResponse
+    {
+        public JToken ExceptionDetails { get; set; }
```
Meir017: shouldn't this be a class?

lib/PuppeteerSharp/Page.cs
```diff
@@ -1865,7 +1866,7 @@ private async Task<object> ExecuteBinding(BindingCalledResponse e)
                 if (taskResult.GetType().IsGenericType)
                 {
                     // the task is already awaited and therefore the call to property Result will not deadlock
-                    result = ((dynamic)taskResult).Result;
+                    result = taskResult.GetType().GetProperty(taskResultPropertyName).GetValue(taskResult);
```
Meir017: the old way was weird but this also feels a bit odd, is there a clean way to do this?

lib/PuppeteerSharp/Page.cs
```diff
@@ -1904,13 +1905,13 @@ private async Task OnAttachedToTarget(MessageEventArgs e)
...
 
-        private void OnLogEntryAdded(LogEntryAddedResponse e)
+        private async Task OnLogEntryAddedAsync(LogEntryAddedResponse e)
         {
             if (e.Entry.Args != null)
             {
                 foreach (var arg in e.Entry?.Args)
                 {
-                    RemoteObjectHelper.ReleaseObject(Client, arg, _logger);
+                    await RemoteObjectHelper.ReleaseObjectAsync(Client, arg, _logger);
```
Meir017: is this missing `.ConfigureAwait(false)` ?

lib/PuppeteerSharp/ExecutionContext.cs
Meir017: fair enough

kblok: I will implement a class @Meir017 the shape will be a composition of that ;)

lib/PuppeteerSharp/Page.cs
kblok: I open to new ideas 😂

kblok: Created an issue for this https://github.com/kblok/puppeteer-sharp/issues/824

kblok:
Ready to review again @Meir017

PR #826, opened 2019-01-01
New BrowserFetcher proxy

### LINKED ISSUE #432: BrowserFetcher: download through a proxy
Hello,
at the moment I have to download and unpack the browser myself.
browserFetcher.DownloadAsync fails in my ASP.net Application because I am behind a proxy.
Can you please implement something like this:
`            
var browserFetcher = new BrowserFetcher(new BrowserFetcherOptions
 {
     Path = TEMP_PATH,
     PROXY = new WebProxy(PROXY_URL) { UseDefaultCredentials = true };
 });
`
Then we can use this proxy for the webclient.

kblok:
Closes #432

### DIFF
--- a/lib/PuppeteerSharp.Tests/PuppeteerTests/BrowserFetcherTests.cs
+++ b/lib/PuppeteerSharp.Tests/PuppeteerTests/BrowserFetcherTests.cs
@@ -78,27 +78,23 @@ namespace PuppeteerSharp.Tests.PuppeteerTests
         [Fact]
         public async Task ShouldUseWebProxy()
         {
-            var tcs = new TaskCompletionSource<string>();
+            var tcs = new TaskCompletionSource<bool>();
 
             var browserFetcher = new BrowserFetcher()
             {
-                Proxy = new WebProxy(TestConstants.ServerUrl)
+                Proxy = new WebProxy(TestConstants.HttpsPrefix)
             };
 
-            Server.SetRoute(string.Empty, context =>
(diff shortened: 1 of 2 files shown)


Abrissirba:
When will this be merged to master?

kblok:
It's not in stand by. I wasn't able to make the tests to pass on .NET Framework.
https://ci.appveyor.com/project/kblok/puppeteer-sharp/builds/21324137/job/q349i76ash0753jb/tests

Any help will be appreciated.

Abrissirba:
I gave it a try but didn't manage to solve it either. The HttpsServer doesn't seem to get the requests made by the WebClient when using the proxy. The WebClient makes a request with the HTTP Verb CONNECT which the server doesn't seem to handle. I will fork the the repo in the meantime until a solution for this is found, if we really need to have a unit test for it ;)

Abrissirba:
I didn't get the test to work with the **Https**Server. The server middleware was never executed. When I changed to the Server (http), the middleware was called and the test passed. 

If the Host in BrowserFetcherOptions is set to an **http** uri instead of **https** the test fails due to how WebClient forwards the request to the proxy. When the request is made to an **https** uri, WebClient will make a **CONNECT** request to the proxy with an empty path. If the request is made to an http uri, WebClient will make a **GET** request with the path set to the path of the original request. Since the unit test is set up to listen on an empty path no routes will be found and the test will fail due to no response.

Any ideas on why the **HttpsServer** doesn't get called? Would it be ok to change the unit test to use **Server** instead?

PR #840, opened 2019-01-05
Remove Task.Delays

kblok:
Based on this https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#always-create-taskcompletionsourcet-with-taskcreationoptionsruncontinuationsasynchronously

We are removing Task.Delay used for cancellations

### DIFF
--- a/lib/PuppeteerSharp/Helpers/TaskHelper.cs
+++ b/lib/PuppeteerSharp/Helpers/TaskHelper.cs
@@ -17,31 +17,13 @@ namespace PuppeteerSharp.Helpers
         /// <param name="task">Task to wait for.</param>
         /// <param name="milliseconds">Milliseconds timeout.</param>
         /// <param name="exceptionToThrow">Optional exception to be thrown.</param>
-        public static async Task WithTimeout(
+        public static Task WithTimeout(
             this Task task,
             int milliseconds = 1_000,
             Exception exceptionToThrow = null)
-        {
-            var tcs = new TaskCompletionSource<bool>();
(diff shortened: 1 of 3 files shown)


lib/PuppeteerSharp/Helpers/TaskHelper.cs
```diff
@@ -43,9 +47,78 @@ public static async Task WithTimeout(this Task task, int milliseconds = 1_000)
...
         /// <returns>The task result.</returns>
         /// <param name="task">Task to wait for.</param>
+        /// <param name="timeoutAction">Action to be executed on Timeout.</param>
         /// <param name="milliseconds">Milliseconds timeout.</param>
+        public static async Task WithTimeout(
+            this Task task,
+            Func<Task> timeoutAction,
+            int milliseconds = 1_000)
+        {
+            var tcs = new TaskCompletionSource<bool>();
```
Meir017: maybe create a private method for creating the `TaskCompletionSource` & `CancellationTokenSource` and canceling after the timeout? this code is duplicated multiple times
something like:
```cs
private static (TaskCompletionSource<bool> tcs, CancellationTokenSource cancellationSource) CreateSources(int milliseconds = 1_000)
{
    var tcs = new TaskCompletionSource<bool>();
    var cancellationSource = new CancellationTokenSource();
    if (milliseconds > 0)
    {
        cancellationToken.CancelAfter(milliseconds);
    }
    return (tcs, cancellationSource);
}
```

kblok:
What do you think now @Meir017?

PR #842, opened 2019-01-07
Add overloads of EvaluateFunctionAsync with disposeHandle flag

hlaueriksson:
I would like to add a flag to the `EvaluateFunctionAsync` methods, to be able to opt out of the disposal of the element/array handle.

If the page function is pure with no side effects on the element/array handle, then this could be possible.

Consider this scenario:

```
await Page.SetContentAsync("<section id='testAttribute'>43543</section>");
var section = await Page.QuerySelectorAsync("section");
var idAttribute = await section.EvaluateFunctionAsync<string>("e => e.id", disposeHandle: false);
var text = await section.EvaluateFunctionAsync<string>("e => e.textContent", disposeHandle: false);
```

Here we can avoid the `EvaluationFailedException` (JSHandle is disposed!), when evaluating two functions on the same element.

### DIFF
--- a/lib/PuppeteerSharp/Extensions.cs
+++ b/lib/PuppeteerSharp/Extensions.cs
@@ -68,7 +68,7 @@ namespace PuppeteerSharp
         /// <typeparam name="T">The type of the response</typeparam>
         /// <param name="elementHandle">An <see cref="ElementHandle"/> that will be used as the first argument in <paramref name="pageFunction"/></param>
         /// <param name="pageFunction">Function to be evaluated in browser context</param>
-        /// <param name="disposeHandle">If set to <c>false</c> the <paramref name="elementHandle"/> will not be disposed. Only opt out of disposure if the <paramref name="pageFunction"/> is a pure function with no side effects on the <paramref name="elementHandle"/>.</param>
+        /// <param name="disposeHandle">If set to <c>false</c> the <paramref name="elementHandle"/> will not be disposed. Only opt out of disposal if the <paramref name="pageFunction"/> is a pure function with no side effects on the <paramref name="elementHandle"/>.</param>
         /// <param name="args">Arguments to pass to <c>pageFunction</c></param>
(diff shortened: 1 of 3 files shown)


lib/PuppeteerSharp/Extensions.cs
```diff
@@ -62,6 +62,36 @@ public static async Task<T> EvaluateFunctionAsync<T>(this ElementHandle elementH
...
+        /// </summary>
+        /// <typeparam name="T">The type of the response</typeparam>
+        /// <param name="elementHandle">An <see cref="ElementHandle"/> that will be used as the first argument in <paramref name="pageFunction"/></param>
+        /// <param name="pageFunction">Function to be evaluated in browser context</param>
+        /// <param name="disposeHandle">If set to <c>false</c> the <paramref name="elementHandle"/> will not be disposed. Only opt out of disposal if the <paramref name="pageFunction"/> is a pure function with no side effects on the <paramref name="elementHandle"/>.</param>
+        /// <param name="args">Arguments to pass to <c>pageFunction</c></param>
+        /// <returns>Task which resolves to the return value of <c>pageFunction</c></returns>
+        /// <exception cref="SelectorException">If <paramref name="elementHandle"/> is <c>null</c></exception>
+        public static async Task<T> EvaluateFunctionAsync<T>(this ElementHandle elementHandle, string pageFunction, bool disposeHandle, params object[] args)
```
kblok: I'm not against this feature, but this would be a breaking change for people sending a `bool` as the first `params object[] args`

kblok: I have 2 ideas here.
 1. A property on the ElementHandle. `AutoDispose`.
 2. Move the `disposeHandle`to the second position on this signature.

Thoughts @Meir017?

hlaueriksson: Good catch

kblok:
Closed due to inactivity. Feel free to reopen it if needed.

PR #852, opened 2019-01-21
Remove message keys

kblok:


lib/PuppeteerSharp/Messaging/BrowserGrantPermissionsRequest.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Collections.Generic;
+
+namespace PuppeteerSharp.Messaging
+{
+    internal class BrowserGrantPermissionsRequest
+    {
+        public string Origin { get; set; }
+        public string BrowserContextId { get; set; }
+        public IEnumerable<OverridePermission> Permissions { get; set; }
```
Meir017: This could be an array since it represents a json object

PR #894, opened 2019-02-02
Relax request matching heuristic

### LINKED ISSUE #872: Relax request matching heuristic
https://github.com/kblok/puppeteer/pull/15/commits/04fbbd7cf00852fa5b8ecf878f7cb34c9796453d

Meir017:
closes #872

### DIFF
--- a/lib/PuppeteerSharp/Payload.cs
+++ b/lib/PuppeteerSharp/Payload.cs
@@ -14,6 +14,16 @@ namespace PuppeteerSharp
     /// </summary>
     public class Payload
     {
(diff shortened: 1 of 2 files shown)


lib/PuppeteerSharp/Payload.cs
```diff
@@ -14,6 +14,16 @@ namespace PuppeteerSharp
     /// </summary>
     public class Payload
     {
+        private static readonly ISet<string> IgnoredHeaders = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
```
kblok: Why not just a simple List?

Meir017: a bit faster

PR #900, opened 2019-02-04
[WIP] Fix unobserved task exception

kblok:
Super WIP


ggeurts:
Any usage of `ContinueWith` that result in a task whose result is never observed (a fire-and-forget task) will can cause unhandled exceptions. Instead of `someTask.ContinueWith(t => FireAndForget())` it would be better to use the pattern

    _ = FireAndForgetAsync();
    Task FireAndForgatAsync()
    {
        await someTask.ConfigureAwait(false);
        ForeAndForget();
    }

ggeurts:
All `async void` methods are also candiates for the same treatment for the same reason.

PR #906, opened 2019-02-06
Introduce Popup event

### LINKED ISSUE #859: Introduce "Popup" event
https://github.com/kblok/puppeteer/pull/15/commits/8aaca4eb1ed4020c56d03e3957c1e64e9a0e8b45

kblok:
closes #859

### DIFF
--- a/lib/PuppeteerSharp/Page.cs
+++ b/lib/PuppeteerSharp/Page.cs
@@ -193,6 +193,11 @@ namespace PuppeteerSharp
         public event EventHandler Close;
 
         /// <summary>
+        /// Raised when the page opens a new tab or window.
+        /// </summary>
+        public event EventHandler<PopupEventArgs> Popup;
+
+        /// <summary>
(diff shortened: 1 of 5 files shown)


lib/PuppeteerSharp/Page.cs
```diff
@@ -312,6 +317,7 @@ public int DefaultNavigationTimeout
         public Accessibility Accessibility { get; }
 
         internal bool JavascriptEnabled { get; set; } = true;
+        internal bool HasPopupEventListeners => Popup?.GetInvocationList().Any() == true;
         #endregion
```
Meir017: Why do we need this?

kblok: I can only access the Invocation list from the inside of the class.
So here https://github.com/kblok/puppeteer-sharp/pull/906/files#diff-304795bf53e0aba6fcc27d50cbc6d7dfR46 if  we have no listeners we don't need to await for `PageAsync`.

PR #908, opened 2019-02-06
filter out "intervention" header

### LINKED ISSUE #877: Filter out "intervention" header
https://github.com/kblok/puppeteer/pull/15/commits/9fd4b67d0c6597f32a5907f241f081add0d492d3

Meir017:
closes #877

### DIFF
--- a/lib/PuppeteerSharp/Payload.cs
+++ b/lib/PuppeteerSharp/Payload.cs
@@ -21,7 +21,8 @@ namespace PuppeteerSharp
             "x-devtools-emulate-network-conditions-client-id",
             "cookie",
             "origin",
-            "content-type"
+            "content-type",
+            "intervention"
         };
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


lib/PuppeteerSharp/Payload.cs
```diff
@@ -21,7 +21,8 @@ public class Payload
             "x-devtools-emulate-network-conditions-client-id",
             "cookie",
             "origin",
-            "content-type"
+            "content-type",
+            "intervention"
```
kblok: The other day I kept thinking about the HashKey. Then I realized, Shouldn't a `const string[]` be even faster and lightweight?

Meir017: Hmm, I'm not sure.
You can run a benchmark to test that

kblok: Not strong on this, just a thought

PR #915, opened 2019-02-09
Warn on nested js handle

kblok:
We are not getting a `Converting circular structure to JSON` when we try to serialize a JSHandle, and I also consider that's not the best way to infer that, we can do better in C#
So I created a JSHandlerCoverter which will throw that exception is someone is trying to send a JSHandle.


Meir017:
Does this still work sending in an `ElementHandle`?

kblok:
Test added @Meir017

PR #917, opened 2019-02-10
Add browserUrl option to Puppeteer.ConnectAsync

### LINKED ISSUE #869: Add browserUrl option to puppeteer.connect
https://github.com/kblok/puppeteer/pull/15/commits/15af75f9a24ad9a91ab7cd7a57290e212c8c1c49 and 
https://github.com/kblok/puppeteer/pull/15/commits/d346cb57b4e21f5b7dbff4051b45c6d1e9da68d6

kblok:
closes #869

### DIFF
--- a/lib/PuppeteerSharp/Launcher.cs
+++ b/lib/PuppeteerSharp/Launcher.cs
@@ -4,6 +4,8 @@ using System.Linq;
 using System.IO;
 using Microsoft.Extensions.Logging;
 using PuppeteerSharp.Messaging;
+using System.Net.Http;
+using Newtonsoft.Json;
 
 namespace PuppeteerSharp
 {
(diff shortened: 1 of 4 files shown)


lib/PuppeteerSharp/Launcher.cs
```diff
@@ -100,6 +111,25 @@ public async Task<Browser> ConnectAsync(ConnectOptions options)
             }
         }
 
+        private async Task<string> GetWSEndpointAsync(string browserURL)
+        {
+            try
+            {
+                if (Uri.TryCreate(new Uri(browserURL), "/json/version", out var endpointURL))
+                {
+                    var client = new HttpClient();
```
Meir017: shouldn't we create a static `HttpClient` or at least a single instance for the browser?

kblok: I don’t think this part is critical. When do you hit Launcher.ConnectAsync will be call too many time so as a local HttpClient instance would be a problem?

kblok: I don’t think this part is critical. When do you hit Launcher.ConnectAsync will be call too many time so as a local HttpClient instance would be a problem?

Meir017: OK, I wasn't sure about this but just wanted to be sure

PR #920, opened 2019-02-14
Move to flatten protocol

### LINKED ISSUE #880: Move to flatten protocol
https://github.com/kblok/puppeteer/pull/15/commits/89a5c396bf71fdf6d2dccc935190e353d6eb671f

kblok:
closes #880


lib/PuppeteerSharp/Tracing.cs
```diff
@@ -75,7 +75,17 @@ public Task StartAsync(TracingOptions options)
             {
                 TransferMode = "ReturnAsStream",
                 Categories = string.Join(", ", categories)
-            });
+            });
+
+            /*
```
Meir017: should this be deleted?

lib/PuppeteerSharp/Helpers/AsyndDictionaryHelper.cs
```diff
@@ -0,0 +1,44 @@
...
+            var tcs = new TaskCompletionSource<TValue>(TaskCreationOptions.RunContinuationsAsynchronously);
+            _pendingRequests.Add(key, tcs);
+
+            if (Dictionary.TryGetValue(key, out var item))
+            {
+                _pendingRequests.Delete(key, tcs);
+                return item;
+            }
+
+            return await tcs.Task.WithTimeout(WaitForRequestDelay, new PuppeteerException(string.Format(TimeoutMessage, key)));
```
Meir017: maybe we could pass in a `Func<Exception>` instead of creating an exception object? less allocation.
then we could create a static method that will be passed in here

lib/PuppeteerSharp/Helpers/AsyndDictionaryHelper.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System.Collections.Concurrent;
+using System.Threading.Tasks;
+
+namespace PuppeteerSharp.Helpers
+{
+    internal class AsyncDictionaryHelper<TKey, TValue>
+    {
+        public ConcurrentDictionary<TKey, TValue> Dictionary { get; }
```
Meir017: I think this should be private or at least exposed as an `IDictionary`

lib/PuppeteerSharp/Tracing.cs
kblok: It's WIP ;)

kblok:
Ready  to review @Meir017

PR #925, opened 2019-02-19
Bug (NetworkManger): The key already existed in the dictionary 

Im5tu:
Recently we have been seeing the following under load: 

```
ex_msg: The key already existed in the dictionary. | stacktrace: System.ArgumentException: The key already existed in the dictionary.
   at System.Collections.Concurrent.ConcurrentDictionary`2.System.Collections.Generic.IDictionary<TKey,TValue>.Add(TKey key, TValue value)
   at PuppeteerSharp.NetworkManager.<OnRequestWillBeSentAsync>d__44.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.GetResult()
   at PuppeteerSharp.NetworkManager.<Client_MessageReceived>d__36.MoveNext()
```

This PR looks at some of the causes for this. We are still testing the changes, but wanted to get this here for your visibility and thoughts also.

I've managed to:
- Remove two of the dictionaries that were previously required
- Optimise the way some of the dictionaries were used (ie: using TryRemove instead of Get/Remove)
- Fixed up some event invocation handling
- Fixed a few minor race conditions (ie: MultiMap.Add - note that there is still a problem when the list needs to grow the array, but from what I can see the default capacity should be enough to warrant not locking)

... all whilst _hopefully_ kept the same semantics.

_It'd be really nice if chrome fixed the mentioned bug and started to send back the request id, then the entire class could be handled with a single ConcurrentDictionary. I know this is not you though ;)_

### DIFF
--- a/lib/PuppeteerSharp/NetworkManager.cs
+++ b/lib/PuppeteerSharp/NetworkManager.cs
@@ -14,11 +14,8 @@ namespace PuppeteerSharp
         #region Private members
 
         private readonly CDPSession _client;
-        private readonly IDictionary<string, Request> _requestIdToRequest = new ConcurrentDictionary<string, Request>();
-        private readonly IDictionary<string, RequestWillBeSentPayload> _requestIdToRequestWillBeSentEvent =
-            new ConcurrentDictionary<string, RequestWillBeSentPayload>();
+        private readonly ConcurrentDictionary<string, NetworkRequest> _requests = new ConcurrentDictionary<string, NetworkRequest>();
         private readonly MultiMap<string, string> _requestHashToRequestIds = new MultiMap<string, string>();
-        private readonly MultiMap<string, string> _requestHashToInterceptionIds = new MultiMap<string, string>();
         private readonly ILogger _logger;
         private Dictionary<string, string> _extraHTTPHeaders;
         private bool _offine;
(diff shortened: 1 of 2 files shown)


kblok:
I'll go throw this tomorrow. But I get your point. Sometimes it sucks branching from Puppeteer's code but this makes sense.
Let's see what AppVeyor thinks about this :p

Im5tu:
So it seems appveyor is having the same trouble I am running the test suite. Is this a known issue or more likely something that i've messed up?

kblok:
@Im5tu can you pull from master?

Im5tu:
@kblok - Done :)

lib/PuppeteerSharp/NetworkManager.cs
```diff
@@ -247,53 +242,44 @@ private async Task OnRequestInterceptedAsync(RequestInterceptedResponse e)
...
             }
-            if (!_requestIdToRequest.TryGetValue(e.RequestId, out var currentRequest) ||
-              currentRequest.Frame == null)
+            if (!_requests.TryGetValue(e.RequestId, out var request) ||
+              request.RawRequest.Frame == null)
             {
                 var frame = await FrameManager?.GetFrameAsync(e.FrameId);
 
-                request = new Request(
+                request.RawRequest = new Request(
```
kblok: WorkerTests.ShouldEmitCreatedAndDestroyedEvents is failing around here. `request` is null, so we fail setting `RawRequest`

kblok:
I think we still have something on this test `ShouldNotCrashWhileRedirectingIfOriginalRequest`

Im5tu:
It appears that we get the redirect response prior to getting the request will send portion. I haven't managed to take a look further yet - should be able to get back to resolving this tomorrow :)

PR #941, opened 2019-02-28
Target PuppeteerSharp.AspNetFramework at .NET 4.6.1

ggeurts:
The current framework dependency for PuppeteerSharp.AspNetFramework unnecessarily targets .NET 4.7.2 whereas .NET 4.6.1 is sufficient.


kblok:
Sounds good @ggeurts, but why are we removing files there?

ggeurts:
The .nuspec and packages.config files are no longer needed and neither is the assemblyinfo.cs when using Microsoft.NET.Sdk style projects. The AppVeyor post-build step to build the package no longer is needed either, but has not yet been committed

kblok:
I don't see the package in AppVeyor @ggeurts 
image

kblok:
@ggeurts can you set the version to v1.12? so I publish this package.

PR #946, opened 2019-03-03
Make Target class resposible for its own state management

ggeurts:
State management for the Target class is currently spread out over the Browser, Target and Page classes which makes it hard to verify correctness. Therefore it makes sense to localize all code that determines when certain Target operations can and/or should be performed in the Target class itself.


kblok:
Hey @ggeurts I don't want our code to differ too much from the original (Puppeteer code). Unless it's something that we do need to different in C#

This helps us keep up with the changes they make in their own code.

If you have new ideas, I think it would be more productive for you to begin with an issue, so we can debate the implementation first.

Issue #948, opened 2019-03-03
Ensure Targets are always closed


ggeurts:
`Target` closure is initiated by the `Page` class by a task that is scheduled in its constructor. This task will only complete when `Target.CloseTask` completes. However, `Target.CloseTask` only completes when `Browser` receives a DestroyTarget message. If for some reason this message never arrives, the `Page` will hang indefinitely on closure.

kblok:
Is there any way to reproduce this with a test?

ggeurts:
Yep,

        [Fact(Timeout =  5000)]
        public async Task ShouldCloseWhenConnectionBreaksPrematurely()
        {
            Browser.Connection.Dispose();
            await Page.CloseAsync();
        }

kblok:
Awesome!

ggeurts:
This tests also highlights a bug in ChromeProcess kill timeout

PR #949, opened 2019-03-03
Fix for #948: Improve target destruction reliability

ggeurts:


lib/PuppeteerSharp/ChromiumProcess.cs
```diff
@@ -193,7 +193,7 @@ public async Task<bool> WaitForExitAsync(TimeSpan? timeout)
                 await _exitCompletionSource.Task.WithTimeout(() =>
                 {
                     taskCompleted = false;
-                }, timeout.Value.Milliseconds).ConfigureAwait(false);
+                }, ToTimeoutInt32(timeout.Value)).ConfigureAwait(false);
```
kblok: Why do we need this?

ggeurts: `timeout.Value.Milliseconds` only returns the milliseconds portion of the TimeSpan. Instead TotalMilliseconds should be used, but that may result in a value greater than `int.MaxValue`.

kblok: Maybe we can improve `WithTimeout`, adding an overload with a `TimeSpan` parameter (which would be the normal flow internally)

ggeurts: `TimeSpan.Milliseconds` returns only the milliseconds portion of the `TimeSpan`. Instead, `TotalMilliseconds` should be used, but that returns a `double` that mat be greater than `int.MaxValue`.

kblok:
@ggeurts I think we need both. In many places are sending milliseconds from option arguments.

lib/PuppeteerSharp/Helpers/TaskHelper.cs
```diff
@@ -9,22 +9,33 @@ namespace PuppeteerSharp.Helpers
     /// </summary>
     public static class TaskHelper
     {
+        private static readonly Func<int, Exception> DefaultExceptionFactory =
```
kblok: How about if we make it TimeSpan by default?
The methods expecting milliseconds can use TimeSpan.FromMilliseconds(int)
And we wouldn't need a `ToTimeoutInt32`

ggeurts: True, and I considered it, but `CancellationTokenSource` will convert the `TimeSpan` back to `int` and also throw a `ArgumentOutOfRangeException`.

kblok: Genius!

PR #951, opened 2019-03-04
WebSocketTransport initialization, shutdown and extension

ggeurts:
While testing the latest commits I no longer experienced hangs, but the debugger did break on some exceptions that are not really necessary in the `WebSocketTransport`. This pull request puts some additional belts and braces around the starting, stopping and cancelling of read operations.

### DIFF
--- a/lib/PuppeteerSharp/ConnectOptions.cs
+++ b/lib/PuppeteerSharp/ConnectOptions.cs
@@ -1,12 +1,9 @@
 ﻿using System;
 using System.Net.WebSockets;
+using PuppeteerSharp.Transport;
 
 namespace PuppeteerSharp
 {
-    using System.Threading;
-    using System.Threading.Tasks;
-    using PuppeteerSharp.Transport;
-
     /// <summary>
     /// Options for connecting to an existing browser.
     /// </summary>
@@ -48,7 +45,7 @@ namespace PuppeteerSharp
         /// Optional factory for <see cref="WebSocket"/> implementations.
         /// If <see cref="Transport"/> is set this property will be ignored.
         /// </summary>
-        public Func<Uri, IConnectionOptions, CancellationToken, Task<WebSocket>> WebSocketFactory { get; set; }
+        public WebSocketFactory WebSocketFactory { get; set; }
 
         /// <summary>
         /// Gets or sets the default Viewport.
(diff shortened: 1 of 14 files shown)


kblok:
Can we try to identify how can we get to calls to StartReading?

ggeurts:
I am testing some changes. The design of `IConnectionOptions.Transport` seems flawed, because it allows the same 'WebSocketTransport' to be initialized multiple times. It will only work the first time, any subsequent calls will fail.

kblok:
@ggeurts and how can we initialize it more than once?

ggeurts:
When reusing the same ConnectionOptions instance with a custom IConnectionTransport specified in the Transport property, this transport will be initialized multiple times.

ggeurts:
If only thing that needs to be customizable is the long-running task scheduler, I would introduce BackgroundTaskScheduler property of type Action<Task> on the IConnectionOptions.

kblok:
I was looking for the websocket for windows 7, and I found you there :p

https://github.com/kblok/puppeteer-sharp/pull/574/files

The scenario is not only AspNet but also windows 7 (sorry if I’m missing something I’m mobile now)

ggeurts:
Ok, in that case it makes sense to replace `IConnectionOptions.WebSocketFactory` with a `IConnectionOptions.TransportFactory` and provide a few protected virtual methods in `WebSocketTransport` that will acts as extension points for custom implementations. I can send an implementation proposal tomorrow.

kblok:
If you come up with a 100% backward compatible solution is a 100% yes.

lib/PuppeteerSharp/ConnectOptions.cs
```diff
@@ -48,8 +48,14 @@ public class ConnectOptions : IBrowserOptions, IConnectionOptions
         /// Optional factory for <see cref="WebSocket"/> implementations.
         /// If <see cref="Transport"/> is set this property will be ignored.
         /// </summary>
+        [Obsolete("Use TransportFactory instead")]
         public Func<Uri, IConnectionOptions, CancellationToken, Task<WebSocket>> WebSocketFactory { get; set; }
```
kblok: I think we shouldn't mark this as `Obsolete`. We should still be able to do
```cs
WebSocketFactory = (uri, socketOptions, cancellationToken) =>
	{
		return System.Net.WebSockets.SystemClientWebSocket.ConnectAsync(uri, cancellationToken);
	},
```

ggeurts: Ok, I will submit a different pull-request with a consolidation of the changes in that case.

PR #990, opened 2019-03-06
Change default timeout back to 30000ms

kkempster94:
It looks like the default timeout was changed with this commit: https://github.com/kblok/puppeteer-sharp/commit/2e59b1eef049836b54767f09c72d345ac0f84227

I figured this was a typo because the comments seem to imply that it should still be 30 seconds. 

This is my first pull request, so let me know if I did something wrong.


kblok:
Ops, good catch!
I'll release a hotfix with this.

PR #1023, opened 2019-03-19
Click and Goto tests refactors

kblok:


lib/PuppeteerSharp.Tests/InputTests/ClickTests.cs
```diff
@@ -0,0 +1,254 @@
...
+using PuppeteerSharp.Input;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace PuppeteerSharp.Tests.InputTests
+{
+    [Collection("PuppeteerLoaderFixture collection")]
+    public class ClickTests : PuppeteerPageBaseTest
+    {
+        private const string Dimensions = @"function dimensions() {
```
Meir017: this constant appears also in `InputTests.cs`, maybe reference the same constant from both classes?

kblok: Good catch. Removed.

Issue #1050, opened 2019-04-01
Review Demo project CI fail


kblok:
Created by https://github.com/kblok/puppeteer-sharp/pull/1036

https://ci.appveyor.com/project/kblok/puppeteer-sharp-0c8w9/builds/23475340/job/vfpp4tm2d222nk7r

PR #1051, opened 2019-04-01
Isolating the Posix code so it doesn't get loaded in net471.

scabana:
Fixes https://github.com/kblok/puppeteer-sharp/issues/1050

As the code path is now called from a local function, the static fields of LinuxPermissionsSetter are not loaded by the .net until the local function is called. What was happening, is when ```DownloadAsync``` is invoked, it loads the ```LinuxPermissionsSetter``` class, thus loading the static field ```ExecutableFilePermissions``` which then loads the ```FileAccessPermissions``` enum from the Posix lib.

### DIFF
--- a/lib/PuppeteerSharp/BrowserFetcher.cs
+++ b/lib/PuppeteerSharp/BrowserFetcher.cs
@@ -204,7 +204,12 @@ namespace PuppeteerSharp
 
             if (revisionInfo != null && GetCurrentPlatform() == Platform.Linux)
             {
-                LinuxPermissionsSetter.SetExecutableFilePermissions(revisionInfo.ExecutablePath);
+                void SetPermissions(string fullPath)
+                {
+                    LinuxPermissionsSetter.SetExecutableFilePermissions(fullPath);
+                }
+
+                SetPermissions(revisionInfo.ExecutablePath);
             }
             return revisionInfo;
         }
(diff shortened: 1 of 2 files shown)


kblok:
Nice trick

lib/PuppeteerSharp/BrowserFetcher.cs
```diff
@@ -204,7 +204,12 @@ public async Task<RevisionInfo> DownloadAsync(int revision)
 
             if (revisionInfo != null && GetCurrentPlatform() == Platform.Linux)
             {
-                LinuxPermissionsSetter.SetExecutableFilePermissions(revisionInfo.ExecutablePath);
+                void SetPermissions(string fullPath)
```
kblok: Can we arrow this?
```cs
void SetPermissions(string fullPath) => LinuxPermissionsSetter.SetExecutableFilePermissions(fullPath);
```

scabana: 👍

scabana:
@kblok: I suggest changing the 480000 value from the /lib/PuppeteerSharp.Tests/test.runsettings to something higher... it's mostly always timing out because of that from what I see.

kblok:
@scabana Those tests are getting stuck, increasing the timeout won't help.

scabana:
480000 is 8 minutes. Are you sure they're getting stuck? There are well over 500 tests, at 1 second a test, it would take more than 8 minutes to run... I think it's just close to the limit. The linux build tests took 7.4743, which is just ~30 seconds from failing.

kblok:
You are right :) let's give it a shot. 15mins?

PR #1062, opened 2019-04-07
Paper Format Accessibility Changes

### LINKED ISSUE #1061: PaperFormat Should Be Immutable with Public Constructor
Recently, I was trying to define dimensions using `PaperFormat` for a "#10" envelope.  

I know that you can use `Height` and `Width` on the `PdfOptions` object, but for consistency's sake, I needed to use `PaperFormat`.

I noticed that `PaperFormat` does not have a public constructor.  I don't see a really good reason for this—especially considering that `Height` and `Width` have public setters.

So as an experiment, I modified the `Height` and `Width` of `PaperFormat.Letter` to 9.5 by 4.125. It yielded a bad side effect.  Because `PaperFormat.Letter` is a static object, all subsequent uses of it ended up being 9.5 by 4.125. While that behavior is conceptually expected, it could cause something that's very difficult to debug.

As a result, I propose that `Height` and `Width` be public readonly properties so that we don't undo the functionality from #407. 

Furthermore, I propose that we make a public constructor for `PaperFormat` with `Width` and `Height` as parameters so that people can consistently use `PaperFormat` rather than switching back and forth between using `PaperFormat` and `PdfOptions.Height` and  `PdfOptions.Width`.

aaroncadrian:
- Make a public constructor for `PaperFormat`
- Make `Height` and `Width` readonly properties.

closes #1061

### DIFF
--- a/lib/PuppeteerSharp/Media/PaperFormat.cs
+++ b/lib/PuppeteerSharp/Media/PaperFormat.cs
@@ -6,7 +6,7 @@ namespace PuppeteerSharp.Media
     /// <summary>
     /// Paper format.
     /// </summary>
-    /// <seealso cref="PdfOptions.Format" />
+    /// <seealso cref="PdfOptions.Format"/>
     public class PaperFormat : IEquatable<PaperFormat>
     {
         /// <summary>
--- a/lib/PuppeteerSharp/Media/PaperFormat.cs
+++ b/lib/PuppeteerSharp/Media/PaperFormat.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 
 namespace PuppeteerSharp.Media
@@ -6,7 +6,7 @@ namespace PuppeteerSharp.Media
     /// <summary>
     /// Paper format.
     /// </summary>
-    /// <seealso cref="PdfOptions.Format"/>
+    /// <seealso cref="P:PuppeteerSharp.PdfOptions.Format" />
     public class PaperFormat : IEquatable<PaperFormat>
     {
         /// <summary>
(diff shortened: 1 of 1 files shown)


lib/PuppeteerSharp/Media/PaperFormat.cs
```diff
@@ -6,11 +6,15 @@ namespace PuppeteerSharp.Media
     /// <summary>
     /// Paper format.
     /// </summary>
-    /// <seealso cref="PdfOptions.Format"/>
+    /// <seealso cref="P:PuppeteerSharp.PdfOptions.Format" />
```
kblok: We don't need this change.

lib/PuppeteerSharp/Media/PaperFormat.cs
```diff
@@ -20,56 +24,66 @@ private PaperFormat(decimal width, decimal height)
         /// Page width in inches
         /// </summary>
         /// <value>The width.</value>
-        public decimal Width { get; set; }
+        public readonly decimal Width;
```
kblok: I wouldn't like to break compatibility for this. I would leave this as it is.
What we could do is turning PaperFormats into getters. By doing this we would return a new PaperFormat every time they request it.

lib/PuppeteerSharp/Media/PaperFormat.cs
aaroncadrian: No problem. I've reverted it.

lib/PuppeteerSharp/Media/PaperFormat.cs
aaroncadrian: Great point about breaking compatibility.  I've reverted them back to getters and setters, and I've made the static properties getters instead of static properties.

PR #1073, opened 2019-04-16
[WIP] Puppeteer abstractions

### LINKED ISSUE #1049: Research: Analyze and create issues for interfaces
Let's see how PuppeteerSharp.Abstractions would look like and create issues to work on it.

Meir017:
Closes #1049


kblok:
What I was thinking, even before taking a look at your PR, is that the rule number #1 should be not breaking existing code.
So, for instance, the class BoundingBox (to mention any). I would move that class to abstractions,
but I would leave one class in the same namespace marking it as obsolete and inheriting from the abstraction class. We could remove it in the next iteration. 
The same for the interfaces. We need to implement explicit interfaces. So. for instance,
IBrowser.NewPageAsync returns IPage. but if  I'm currently using Browser.NewPage I would still get a Page.
Not breaking existing code base is priority number 1.

Meir017:
@kblok so a good indicator will be that the tests project should not change at all?

kblok:
@Meir017 exactly.

PR #1074, opened 2019-04-19
Microsoft Edge Friendly tests

kblok:
We made some minor changes to our tests so they run with Microsoft Edge:
 * Intervention header has the URL https://to-be-replaced.invalid/feature/5718547946799104 instead of https://www.chromestatus.com/feature/5718547946799104. We are changing the "Contains" part
 * We are removing some new lines and using Trim to compare TextAsync
 * Edge launches two pages with default arguments. We are going to check that  the pages array is not empty.

### DIFF
--- a/lib/PuppeteerSharp.Tests/ChromiumSpecificTests/PageTests.cs
+++ b/lib/PuppeteerSharp.Tests/ChromiumSpecificTests/PageTests.cs
@@ -34,7 +34,7 @@ namespace PuppeteerSharp.Tests.ChromiumSpecificTests
 
             await Page.GoToAsync(TestConstants.ServerUrl + "/intervention");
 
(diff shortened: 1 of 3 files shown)


Meir017:
> * Edge launches two pages with default arguments. We are going to check that the pages array is not empty.

@kblok did the puppeteer project make these changes as well?

kblok:
@Meir017 No. Do you think I should add some comments as a reference?
Chrome by default opens with one tab, whereas Edge opens with two.

Meir017:
I think a comment should be enough.
maybe also open a pull-request on puppeteer with this change? ;)

kblok:
@Meir017 both are great ideas.

kblok:
Comment added @Meir017

PR #1184, opened 2019-06-22
Allow customization of fetch patterns

syscafedevelopment:
Customization of fetch patterns including specifying the RequestStage and ResourceType

### DIFF
--- a/lib/PuppeteerSharp/Messaging/RequestWillBeSentPayload.cs
+++ b/lib/PuppeteerSharp/Messaging/RequestWillBeSentPayload.cs
@@ -1,4 +1,5 @@
-﻿using Newtonsoft.Json;
+﻿using System;
+using Newtonsoft.Json;
 
 namespace PuppeteerSharp.Messaging
 {
@@ -11,5 +12,7 @@ namespace PuppeteerSharp.Messaging
         public ResponsePayload Response { get; set; }
         public ResourceType Type { get; set; }
         public string FrameId { get; set; }
+
+        internal bool IsInterceptable => !Request.Url.StartsWith("data:", StringComparison.InvariantCultureIgnoreCase);
     }
 }
--- a/lib/PuppeteerSharp/Messaging/RequestWillBeSentPayload.cs
+++ b/lib/PuppeteerSharp/Messaging/RequestWillBeSentPayload.cs
@@ -8,7 +8,7 @@ namespace PuppeteerSharp.Messaging
         public string LoaderId { get; set; }
         public Payload Request { get; set; }
 
-        public ResponsePayload RedirectResponse { get; set; }
+        public ResponsePayload Response { get; set; }
         public ResourceType Type { get; set; }
         public string FrameId { get; set; }
     }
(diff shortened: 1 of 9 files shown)


kblok:
Hey @syscafedevelopment. Could you tell me a real use case scenario?
So I can try to propose this upstream (a.k.a puppeteer).

kblok:
We would also need some tests for this.

And by the way. Thank you for your time contributing :)

syscafedevelopment:
Hey @kblok, 

A scenario would be:
- the dev wants to download a certain resource (e.g., a video, a pdf, etc.)
- when the request interception is made, the dev will need to call Chrome's [Fetch.getResponseBody](https://chromedevtools.github.io/devtools-protocol/tot/Fetch#method-getResponseBody) or [Fetch.takeResponseBodyAsStream](https://chromedevtools.github.io/devtools-protocol/tot/Fetch#method-takeResponseBodyAsStream)
- however, this will require the interception to happen at `HeadersReceived` stage
- additionally, the dev only needs to intercept those specific resource types, having predefined URL patterns
- at the moment, all interceptions are at `Request` stage, with a hardcoded pattern of `"*"`, which does not allow response body interceptions, nor selective interceptions

Apparently there are several complains and complicated workarounds about allowing downloading things via Puppeteer, which would start to be solved by customizing the interception stage first.

About unit tests, I'll handle those as soon as possible.

lib/PuppeteerSharp/Messaging/RequestWillBeSentPayload.cs
```diff
@@ -8,7 +8,7 @@ internal class RequestWillBeSentPayload
         public string LoaderId { get; set; }
         public Payload Request { get; set; }
 
-        public ResponsePayload RedirectResponse { get; set; }
```
syscafedevelopment: I wonder if this is ever used, considering it has never been set before

syscafedevelopment:
@kblok, any idea why this is not passing? It seems it needs Mono.Posix assembly? Why?

kblok:
@syscafedevelopment it seems that ShouldEmitCreatedAndDestroyedEvents is getting stuck...

On the other hand, I like this feature, but I'm waiting for Puppeteer's feedback here https://github.com/GoogleChrome/puppeteer/issues/4624. 
I don't want us to implement a feature and then they implement the same feature in a different way and we would be forced to change our API.

kblok:
@syscafedevelopment could you share your case here https://github.com/GoogleChrome/puppeteer/issues/4624 ?

emiliavanderwerf:
@kblok @syscafedevelopment Is this planned to be merged soon please?

kblok:
@emiliavanderwerf I could try to push a little bit upstream to see if we get this implemented in puppeteer.

emiliavanderwerf:
@kblok Ah, so not even [Puppeteer](https://github.com/puppeteer/puppeteer/blob/v7.0.0/docs/api.md) implements this yet?

kblok:
@emiliavanderwerf exactly.

PR #1191, opened 2019-06-29
Adding support for .net framework on Win 7 and Server 2008

jbtule:
- Adds dependency on System.Net.WebSockets.Client.Managed.
- If on a platform that is missing ClientWebSocket, uses Managed version.

### DIFF
--- a/lib/PuppeteerSharp/PuppeteerSharp.csproj
+++ b/lib/PuppeteerSharp/PuppeteerSharp.csproj
@@ -39,6 +39,7 @@
     <PackageReference Include="Newtonsoft.Json" Version="10.0.3" />
     <PackageReference Include="System.Net.Http" Version="4.3.3" />
     <PackageReference Include="Microsoft.AspNetCore.WebUtilities" Version="2.0.2" />
+    <PackageReference Include="System.Net.WebSockets.Client.Managed" Version="1.0.22" />
     <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.1" />
   </ItemGroup>
 </Project>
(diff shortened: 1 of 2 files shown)


kblok:
@jbtule thanks for your contribution. There are a few things regarding `System.Net.WebSockets.Client.Managed`.
 * We decided in the past to keep as low as possible the number of dependencies. We don't want to add a dependency, which would help only to a particular platform.
 * To accomplish that we added [LaunchOptions.WebSocketFactory](https://www.puppeteersharp.com/api/PuppeteerSharp.LaunchOptions.html#PuppeteerSharp_LaunchOptions_WebSocketFactory) so you can put that same code but on the client side.

This is how I'm using that in my personal project.

```cs
WebSocketFactory = async (uri, socketOptions, cancellationToken) =>
{
    var client = SystemClientWebSocket.CreateClientWebSocket();
    if (client is System.Net.WebSockets.Managed.ClientWebSocket managed)
    {
        managed.Options.KeepAliveInterval = TimeSpan.FromSeconds(0);
        await managed.ConnectAsync(uri, cancellationToken);
    }
    else
    {
        var coreSocket = client as ClientWebSocket;
        coreSocket.Options.KeepAliveInterval = TimeSpan.FromSeconds(0);
        await coreSocket.ConnectAsync(uri, cancellationToken);
    }

    return client;
},
```

So, again, thank you for to contribution, but I'm closing this PR in favor of an existing feature solving this need.

jbtule:
Thanks for sharing about the  LaunchOptions.WebSocketFactory, and thank's for such a great project!

PR #1193, opened 2019-07-01
Add WebSocketFactory example

kblok:


jbtule:
Don't forget to mention on readme, instead of just the read more for the .net Framework Documentation.  https://github.com/kblok/puppeteer-sharp/blob/master/README.md#prerequisites

kblok:
@jbtule good idea!
I will add it [here](https://github.com/kblok/puppeteer-sharp/pull/1193). I added the example I shared with you there.

Issue #1016, opened 2019-03-14
TargetClosedException after Page.CloseAsync()


siready:
First of all, thank you for your tremendous effort that went into this cool project!

I'm trying to generate PDF file, as you will see in the snippet below and although everything works fine, even when I fire multiple concurrent requests that are opening multiple instances of chromium, I'm seeing some exceptions in the log files.

I've been able to narrow down that the issue occurs when trying to close the connection.
The following exception is being thrown:
* ```PuppeteerSharp.TargetClosedException: 'Navigation failed because browser has disconnected! (Target.detachedFromTarget)'```

Example reproducing the issue:

```csharp
            Stream stream;
            await new BrowserFetcher().DownloadAsync(BrowserFetcher.DefaultRevision);

            var browserOptions = new LaunchOptions
            {
                Headless = true,
                Args = new[] { "--no-sandbox" },
            };

            var browser = await Puppeteer.LaunchAsync(browserOptions, loggerFactory);
            var page = await puppeteerService.GetPage(browser);

            await page.SetContentAsync("<div>Testing</div>", puppeteerService.GetNavigationOptions());
            stream = await page.PdfStreamAsync(puppeteerService.GetPdfOptions());

            await page.CloseAsync(); // after this line the exception is thrown
            await browser.CloseAsync();
```

### Versions

PuppeteerSharp: 1.12.1
.NET: netcoreapp2.1
Platform: Windows 10

Attached log file (LogLevel.Trace): [puppeteer-sharp.log-2019-03-14.txt](https://github.com/kblok/puppeteer-sharp/files/2966308/puppeteer-sharp.log-2019-03-14.txt)

kblok:
Hey @siready, thanks for the report. Do you have the stack trace of the exception?

siready:
Unfortunately stack trace is not preserved. That's all I was able to get:
```at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()```

Also part of the reason why I included the log file.

kblok:
@siready  I just created a brand new console app with the latest Puppeteer-Sharp version and it's working for me

![image](https://user-images.githubusercontent.com/2198466/55954609-853e8580-5c35-11e9-8204-9ea0eea1ca45.png)

Using Windows 10 and Core 2.1

kblok:
Closed due to inactivity. Feel free to reopen it if needed.

R2D221:
[...]

PR #1218, opened 2019-07-16
Fixes bug #1016: Unhandled TargetClosedException

R2D221:
It solves the issue by collecting the lingering tasks, and manually making them throw so that the UnobservedTaskException event doesn't see them anymore.

### DIFF
--- a/lib/PuppeteerSharp.Tests/PageTests/SetContentTests.cs
+++ b/lib/PuppeteerSharp.Tests/PageTests/SetContentTests.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Threading.Tasks;
 using Xunit;
 using Xunit.Abstractions;
@@ -127,6 +128,36 @@ namespace PuppeteerSharp.Tests.PageTests
         {
             await Page.SetContentAsync("<div>\n</div>");
             Assert.Equal("\n", await Page.QuerySelectorAsync("div").EvaluateFunctionAsync<string>("div => div.textContent"));
+        }
+
+        [Fact]
+        public async Task ShouldWorkWithoutThrowingUnhandledTaskExceptions()
+        {
+            var exceptions = new List<TargetClosedException>();
+
+            TaskScheduler.UnobservedTaskException += (_, args) =>
+            {
+                args.Exception.Handle(ex =>
+                {
+                    if (ex is TargetClosedException exception)
+                    {
+                        exceptions.Add(exception);
+                        return true;
+                    }
+
+                    return false;
+                });
+            };
+
+            var page = await Browser.NewPageAsync();
+            await page.SetContentAsync("<html></html>");
+            await page.CloseAsync();
+            page = null;
+
+            GC.Collect();
+            GC.WaitForPendingFinalizers();
+
+            Assert.Empty(exceptions);
         }
     }
 }
--- a/lib/PuppeteerSharp.Tests/PageTests/SetContentTests.cs
+++ b/lib/PuppeteerSharp.Tests/PageTests/SetContentTests.cs
@@ -133,6 +133,9 @@ namespace PuppeteerSharp.Tests.PageTests
         [Fact]
         public async Task ShouldWorkWithoutThrowingUnhandledTaskExceptions()
         {
+            // Using this technique to test finalizers:
+            // https://www.inversionofcontrol.co.uk/unit-testing-finalizers-in-csharp/
+
             TaskScheduler.UnobservedTaskException += TaskScheduler_UnobservedTaskException;
 
             WeakReference<Page> weak = null;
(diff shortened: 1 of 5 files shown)


R2D221:
I tried two different methods (you can see there are two commits), but it's weird, because the tests run fine in my machine for either of them, but AppVeyor doesn't seem happy about them.

It may be a little harder than I thought.

kblok:
@R2D221 yes, not easy. But AppVeyor says which code is good and which isn't :p 

If you want to keep trying stuff and pushing code, go for it, I have no problem with that.

kblok:
What do you think about adding these 2 lines in this PR?  https://github.com/kblok/puppeteer-sharp/pull/1152/files#diff-408613782162a72f6aace17be118570aR36

If we want to guarantee that we are observing all exceptions we would need to test it across the board.

R2D221:
I think it makes sense, but if you throw an exception in the unobserved event, I'm not exactly sure where it will throw it.

The approach I was using is collecting the errors in a list, forcing GC to collect, and then see if the list is empty. I don't know if this can be set up globally for all tests.

kblok:
That's true. In my PR I added a `Collect` as you did. Maybe we could add that as a safety net when you finish this feature?
We should never get that. Or at list a `Console.WriteLine` to get that in CI?

R2D221:
Sure. Let me see how this run goes, and maybe we can add it as well.

lib/PuppeteerSharp.Tests/PageTests/SetContentTests.cs
```diff
@@ -128,5 +129,49 @@ public async Task ShouldWorkWithNewline()
...
+            GC.WaitForPendingFinalizers();
+
+            Assert.Empty(exceptions);
+
+            TaskScheduler.UnobservedTaskException -= TaskScheduler_UnobservedTaskException;
+        }
+
+        private static List<TargetClosedException> exceptions = new List<TargetClosedException>();
+
+        private void TaskScheduler_UnobservedTaskException(object sender, UnobservedTaskExceptionEventArgs args)
```
kblok: We can move the event inside the test. Like we do here https://github.com/kblok/puppeteer-sharp/blob/master/lib/PuppeteerSharp.Tests/PageTests/Events/ConsoleTests.cs#L21

lib/PuppeteerSharp.Tests/PageTests/SetContentTests.cs
```diff
@@ -128,5 +129,49 @@ public async Task ShouldWorkWithNewline()
...
+            await func();
+            GC.Collect(0, GCCollectionMode.Forced);
+            GC.WaitForPendingFinalizers();
+
+            Assert.Empty(exceptions);
+
+            TaskScheduler.UnobservedTaskException -= TaskScheduler_UnobservedTaskException;
+        }
+
+        private static List<TargetClosedException> exceptions = new List<TargetClosedException>();
```
kblok: We can move this inside the test.

lib/PuppeteerSharp/LifecycleWatcher.cs
```diff
@@ -70,8 +72,15 @@ internal class LifecycleWatcher : IDisposable
         public Task<bool> SameDocumentNavigationTask => _sameDocumentNavigationTaskWrapper.Task;
         public Task<bool> NewDocumentNavigationTask => _newDocumentNavigationTaskWrapper.Task;
         public Response NavigationResponse => _navigationRequest?.Response;
-        public Task TimeoutOrTerminationTask
-            => _terminationTaskWrapper.Task.WithTimeout(_timeout);
+        public Task TimeoutOrTerminationTask
+        {
+            get
+            {
+                var task = _terminationTaskWrapper.Task.WithTimeout(_timeout);
```
kblok: Can we try to find a solution so we can solve this inside the `WithTimeout` method?
This is not the only place where we use `WithTimeout`. We are using it in the `Page` class, in `Browser` and in `ChromumProcess` among others. 
I can buy if we need to create a kind of TimeoutTask class or something like this.

R2D221: Sure, that would be a cleaner solution. I don't think we need a new class, just a way to ensure we don't create new tasks on every property get.

kblok: BTW, Great Job!

kblok:
One more thing. Could you test this on BrowserDisconnectTests.ShouldRejectNavigationWhenBrowserCloses?
That was my headache on #1152

R2D221:
I was busy on the weekend, but I tried to take a look at that.

I saw you created a variable `LazyBodyLoadedTaskWrapper`, but I think that's not the best way to solve it. Instead of deferring the creation of the task, we should move the exception logic outside of it. Let the task resolve with True or False as a success flag, and only throw an exception when it resolved as False (in this case, when we try to get the response body of a redirect).

I'll push another commit with the way I'm solving it. Just give me a moment.

R2D221:
In general, I think the way to go is checking how the tasks are failing unobserved, and think around it and only fail after we observe them.

kblok:
Looking good, looking good. Thanks for your hard work!

HunteRoi:
Hey there @kblok @R2D221 !

Do you have any status for the PR ? It seems that I have an issue that is similar and the changes could fix it.

Cheers!

R2D221:
@HunteRoi Hi!

Well, I was working on this a while ago, but if I recall correctly, I couldn't completely fix it. So I'd have to see what's missing.

HunteRoi:
Keep us in the loop please :smile:

PR #1234, opened 2019-07-30
New Feature Page.BringToFrontAsync

### LINKED ISSUE #1221: multiple tabs issue, page.bringToFront wanted
### Description
When you work with multiple tabs, inactive tab may work incorrectly, for example not all elements loaded, or page is frozen until you activate it.
In the puppeteer there is useful method to activate particular tab: page.bringToFront
But here seems no way to activate the tab.. 

### Expected behavior:
Activate (bringToFront) special tab in the browser

kblok:
closes #1221

### DIFF
--- a/lib/PuppeteerSharp/Page.cs
+++ b/lib/PuppeteerSharp/Page.cs
@@ -1603,6 +1603,13 @@ namespace PuppeteerSharp
 
             return Task.CompletedTask;
         }
+
+        /// <summary>
+        /// Brings page to front (activates tab).
+        /// </summary>
+        /// <returns>A task which resolves when the message was sent to Chromium.</returns>
+        public Task BringToFrontAsync() => Client.SendAsync("Page.bringToFront");
+
         #endregion
 
         internal void OnPopup(Page popupPage) => Popup?.Invoke(this, new PopupEventArgs { PopupPage = popupPage });


lib/PuppeteerSharp/Page.cs
```diff
@@ -1603,6 +1603,13 @@ public Task SetBurstModeOffAsync()
 
             return Task.CompletedTask;
         }
+
+        /// <summary>
+        /// Brings page to front (activates tab).
+        /// </summary>
+        /// <returns>A task which resolves when the message was sent to Chromium.</returns>
```
LucasMetal: I think I would have written "that resolves" instead of "which resolves", but I'm not particularly against it, just a comment.

LucasMetal: Well, my curiosity led me to this:
https://www.writersdigest.com/online-editor/which-vs-that
so I think it needs to be changed to "that".

> If the sentence doesn’t need the clause that the word in question is connecting, use which. If it does, use that. (Pretty easy to remember, isn’t it?) Let me explain with a couple of examples

LucasMetal:
Also tests are missing for this feature. I think they could be implemented by using something like:
https://stackoverflow.com/questions/2720658/how-to-detect-when-a-tab-is-focused-or-not-in-chrome-with-javascript

PR #1248, opened 2019-08-15
Improve target doc

### LINKED ISSUE #1243: Improve PageAsync()
Explain how would you like this document to be impoved.
Source URL: https://github.com/kblok/puppeteer-sharp/blob/master/lib/PuppeteerSharp/Target.cs/#L91

In docs it says it creates a new page, while it returns currently targeted page.
It could say something like: "Returns targeted Page. If the target is not "page" or "background_page" returns null"

kblok:
closes #1243

### DIFF
--- a/lib/PuppeteerSharp/Target.cs
+++ b/lib/PuppeteerSharp/Target.cs
@@ -108,9 +108,9 @@ namespace PuppeteerSharp
         #endregion
 
         /// <summary>
-        /// A <see cref="Page"/> associated with the target. If the target is not <c>"page"</c> or <c>"background_page"</c> returns <c>null</c>
+        /// Returns the <see cref="Page"/> associated with the target. If the target is not <c>"page"</c> or <c>"background_page"</c> returns <c>null</c>
         /// </summary>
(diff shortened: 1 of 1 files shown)


lib/PuppeteerSharp/Target.cs
```diff
@@ -108,7 +108,7 @@ public class Target
         #endregion
 
         /// <summary>
-        /// Creates a new <see cref="Page"/>. If the target is not <c>"page"</c> or <c>"background_page"</c> returns <c>null</c>
+        /// A <see cref="Page"/> associated with the target. If the target is not <c>"page"</c> or <c>"background_page"</c> returns <c>null</c>
```
LucasMetal: I think I would start like `Returns the/a Page associated with the target...`

lib/PuppeteerSharp/Target.cs
```diff
@@ -108,7 +108,7 @@ public class Target
...
         /// </summary>
         /// <returns>a task that returns a new <see cref="Page"/></returns>
```
LucasMetal: Maybe you can remove `new` from this line?

PR #1302, opened 2019-10-08
Add SourceLink

kblok:


SimonCropp:
you could alternatively go with embedded symbols. then u get line numbers when people get an exception stacktrace in production environments

SimonCropp:
want a PR for embedded pdbs?

kblok:
> want a PR for embedded pdbs?

Let’s do it @SimonCropp !

PR #1324, opened 2019-11-04
Add EmulateMediaTypeAsync and EmulateMediaFeaturesAsync

### LINKED ISSUE #1313: Add page.emulateMedia{Type,Features}
https://github.com/GoogleChrome/puppeteer/commit/a86363fda695865ddc7c5eeb3c958f5c2da8e61b
https://github.com/GoogleChrome/puppeteer/commit/8132b39a6c494b2e2ff7db6b5c062d787b2e8785

kblok:
closes #1313


lib/PuppeteerSharp.Tests/PageTests/EmulateMediaFeaturesTests.cs
```diff
@@ -0,0 +1,46 @@
...
+{
+    [Collection(TestConstants.TestFixtureCollectionName)]
+    public class EmulateMediaFeaturesAsyncTests : PuppeteerPageBaseTest
+    {
+        public EmulateMediaFeaturesAsyncTests(ITestOutputHelper output) : base(output)
+        {
+        }
+
+        [Fact]
+        public async Task ShouldWork()
```
LucasMetal: Usually I try to not have so much code that is so similar, I'm sure this could be reduced a lot using something like NUnit `ValuesAttribute`, because it's really hard to maintain.
But on the other hand, this will probably never change, so it's ok I guess ;)

PR #1361, opened 2020-01-14
Updated to work with VS2019 and .NET Core 3.1

tig:
Pretty self-explanatory. I'm using VS2019 and .NET Core 3.1 for my project and needed to ensure puppeteersharp would work with it.


kblok:
Thanks!

PR #1364, opened 2020-01-20
Added ability to skip resolving file paths for UploadFileAsync. Fixes #1363

### LINKED ISSUE #1363: UploadFileAsync fails with browserless.io
### Description

UploadFileAsync doesn't work with browserless.io - note that I am using hosted version of browserless but the same applies to browserless.io as well.

The upload fails because the method ElementHandle.UploadFileAsync resolves the path using Path.GetFullPath. So when Puppeteer-Sharp is run on Windows, it resolves "/usr/src/app/workspace/filename.ext" to "C:\usr\src\app\workspace\filename.ext".

Since browserless is running on a remote server, they allow you to upload your file to your workspace (https://docs.browserless.io/docs/workspace.html) on the remote server and then reference it with puppeteer.

After uploading the file to your workspace, you can access it as "/usr/src/app/workspace/filename.ext".

```csharp
// Arrange

// Use HttpClient to upload a file to browserless.io workspace.

var url = "http://somefileuploadservice.com/";
var filename = "/usr/src/app/workspace/foo.txt";

string apikey = config["browserlessApiKey"];
var options = new ConnectOptions()
{
    BrowserWSEndpoint = $"wss://chrome.browserless.io?token={apikey}"
};

var browser = await Puppeteer.ConnectAsync(options);

// Act
await page.GoToAsync(url);

var el = page.QuerySelectorAsync("input[name='file1']");
await el.UploadFileAsync(filename);

// Use HttpClient to delete the file from browserless.io workspace.

// Assert
// Check if file was uploaded correctly to the website.
```

### Expected behavior:

UploadFileAsync to successfully upload the file.

### Actual behavior:

Due to the file getting resolved from "/usr/src/app/workspace/filename.ext" to "C:\usr\src\app\workspace\filename.ext", the upload fails.

### Versions

* PuppeteerSharp 2.0
* .NET framework 4.7.2

Code-DJ:
Added ability to skip resolving file paths for UploadFileAsync. Fixes #1363

### DIFF
--- a/lib/PuppeteerSharp.Tests/InputTests/InputTests.cs
+++ b/lib/PuppeteerSharp.Tests/InputTests/InputTests.cs
@@ -51,6 +51,56 @@ namespace PuppeteerSharp.Tests.InputTests
         }
 
         [Fact]
+        public async Task ShouldUploadTheFileIfResolveFilePathIsFalse()
+        {
+            await Page.GoToAsync(TestConstants.ServerUrl + "/input/fileupload.html");
+            var filePath = TestConstants.FileToUpload;
+            var input = await Page.QuerySelectorAsync("input");
+            await input.UploadFileAsync(false, filePath);
+            Assert.Equal("file-to-upload.txt", await Page.EvaluateFunctionAsync<string>("e => e.files[0].name", input));
+            Assert.Equal("contents of the file", await Page.EvaluateFunctionAsync<string>(@"e => {
+                const reader = new FileReader();
+                const promise = new Promise(fulfill => reader.onload = fulfill);
+                reader.readAsText(e.files[0]);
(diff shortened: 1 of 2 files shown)


Code-DJ:
How should it be written?

I understand it needs to be written in ElementHandleTests/ClickTests.cs and use fileupload.html on the server but how can I assert it without a post?

kblok:
I would try to copy `ShouldUploadTheFile`. There you could do

* If I upload a Relative path without the flag it works.
* If I upload a Relative path with the flag it will fail.
* If I upload a Full path with the flag it will work.

kblok:
@Code-DJ  this is is on v2.0.1 now.

Issue #1447, opened 2020-03-09
Random NavigationException exceptions on page.GoToAsync()


mr-wong:
The code is fairly simple:

        private void button1_Click(object sender,EventArgs e) {
            PuppeteerSharp.LaunchOptions opts = new PuppeteerSharp.LaunchOptions() {
                Headless=false,
                ExecutablePath=@"C:\Users\user\AppData\Local\Chromium\Application\chrome.exe",
                DefaultViewport=null,
                UserDataDir=@"d:\32323"
            };
            
            opts.IgnoredDefaultArgs=new string[] { "--enable-automation" };
            
            PuppeteerSharp.Launcher launcher = new PuppeteerSharp.Launcher();

            var browser=launcher.LaunchAsync(opts).GetAwaiter().GetResult();

            var pages=browser.PagesAsync().GetAwaiter().GetResult();

            var page = pages.ElementAt(0);

            for(int i = 0;i<20;i++) {
                Navigate(page,"https://distilnetworks.com");
                Navigate(page,"https://mail.com");
                Navigate(page,"https://distilnetworks.com");
                Navigate(page,"https://vk.com");
                Navigate(page,"https://distilnetworks.com");
                Navigate(page,"https://mail.com");
                Navigate(page,"https://distilnetworks.com");
                Navigate(page,"https://mail.com");
                Navigate(page,"about:blank");
            }
        }

        public void Navigate(PuppeteerSharp.Page page,string url) {
            page.MainFrame.GoToAsync(url,new PuppeteerSharp.NavigationOptions() { Timeout=0,WaitUntil=new PuppeteerSharp.WaitUntilNavigation[] { PuppeteerSharp.WaitUntilNavigation.DOMContentLoaded } }).GetAwaiter().GetResult();
        }

I always get random exceptions mostly like this one:

PuppeteerSharp.NavigationException: 'Protocol error(Page.navigate): Target closed. (NetworkManager failed to process Network.requestWillBeSent. Frame 53F547F3C375005301129799DB86C5A4 not found.    at PuppeteerSharp.Helpers.AsyncDictionaryHelper`2.<>c__DisplayClass4_0.<GetItemAsync>b__0() in D:\Downloads\puppeteer-sharp-master\lib\PuppeteerSharp\Helpers\AsyncDictionaryHelper.cs:line 32
[...]

PR #1450, opened 2020-03-18
[FAILING-TEST] Random NavigationException exceptions on page.GoToAsync()

ben-biddington:
Failing example for #1447


kblok:
Thanks for taking the time to make the example, I'll take a look.

PR #1458, opened 2020-04-08
Add IAsyncDisposable support for Page and Browser

### LINKED ISSUE #1456: Consider adding IAsyncDisposable support for Browser and Page
Thanks for building and maintaining this library!

Today, `Browser` and `Page` have `CloseAsync()` + `Dispose()`.

Now that we have `IAsyncDisposable` and `await using` syntax, it would be nice for these types to also implement that interface and expose a `DisposeAsync()` method which just calls `CloseAsync()`.

This could be done while still targeting netstandard2.0 using https://www.nuget.org/packages/Microsoft.Bcl.AsyncInterfaces/.

If there is interest in this feature, I'd be happy to contribute it (if that would be helpful).

madelson:
Resolves #1456

### DIFF
--- a/lib/PuppeteerSharp.Tests/PageTests/CloseTests.cs
+++ b/lib/PuppeteerSharp.Tests/PageTests/CloseTests.cs
@@ -1,4 +1,4 @@
-﻿using System.Threading.Tasks;
+using System.Threading.Tasks;
 using Xunit;
 using Xunit.Abstractions;
 
@@ -24,11 +24,21 @@ namespace PuppeteerSharp.Tests.PageTests
             Assert.Equal("Target.detachedFromTarget", exception.CloseReason);
         }
 
-        [Fact]
-        public async Task ShouldNotBeVisibleInBrowserPages()
+        [Theory]
+        [InlineData(false)]
+        [InlineData(true)]
+        public async Task ShouldNotBeVisibleInBrowserPages(bool useDisposeAsync)
         {
             Assert.Contains(Page, await Browser.PagesAsync());
-            await Page.CloseAsync();
+            if (useDisposeAsync)
+            {
+                // emulates what would happen in a C#8 await using block
+                await Page.DisposeAsync();
+            }
+            else
+            {
+                await Page.CloseAsync();
+            }
             Assert.DoesNotContain(Page, await Browser.PagesAsync());
         }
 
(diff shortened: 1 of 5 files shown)


lib/PuppeteerSharp.Tests/PageTests/CloseTests.cs
```diff
@@ -24,11 +24,21 @@ public async Task ShouldRejectAllPromisesWhenPageIsClosed()
             Assert.Equal("Target.detachedFromTarget", exception.CloseReason);
         }
 
-        [Fact]
-        public async Task ShouldNotBeVisibleInBrowserPages()
+        [Theory]
+        [InlineData(false)]
+        [InlineData(true)]
```
madelson, line 29: Apologies if this was not what you were going for with your suggestions on https://github.com/hardkoded/puppeteer-sharp/issues/1456#issuecomment-610633676. I wasn't sure and figured that it would be easier when looking at the code.

lib/PuppeteerSharp/PuppeteerSharp.csproj
```diff
@@ -34,7 +34,8 @@
     <PackageReference Include="Newtonsoft.Json" Version="10.0.3" />
     <PackageReference Include="System.Net.Http" Version="4.3.3" />
     <PackageReference Include="Microsoft.AspNetCore.WebUtilities" Version="2.0.2" />
-    <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.1" />
+    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="1.1.0" />
+    <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.2" />
```
madelson, line 38: This version bumped because it is a dependency from AsyncInterfaces. If you'd like, we could now drop the explicit reference to System.Threading.Tasks.Extensions and just consume it transitively through AsyncInterfaces.

madelson:
Thanks for approving @kblok ! 

Are there any remaining steps to get this merged?

Any sense of when these changes will be made available in a nuget.org release?

kblok:
>Thanks for approving @kblok !
>Are there any remaining steps to get this merged?
>Any sense of when these changes will be made available in a nuget.org release?

I was waiting for the CI to build, but it seems I need to change some configs due to the organization change. 
If it's green, I'll merge it and make a release.

Issue #1442, opened 2020-02-29
Enable CA2000 rule


kblok:
Change the action for CA2000 on the [ruleset file](https://github.com/kblok/puppeteer-sharp/blob/master/lib/PuppeteerSharp.ruleset) to Error, and make the solution build

kgar:
Is this one still up for grabs, or did the merge also resolve it?

PR #1463, opened 2020-04-28
Corrects CA2000 and IDE0008

undefinedfunction:
Corrects CA2000 (#1442 ) issue but also fixes IDE0008. This seems to be traced back to 

```
csharp_style_var_for_built_in_types = false:error
```

in `lib/.editorconfig`

This was unexpected but solution compiles now?

_(First public PR - go easy!)_


kblok:
Woohoo, Welcome @undefinedfunction!

PR #1470, opened 2020-05-07
fix CA1063

### LINKED ISSUE #1422: Enable CA1063 rule
Change the action for CA1063 on the [ruleset file](https://github.com/kblok/puppeteer-sharp/blob/master/lib/PuppeteerSharp.ruleset) to Error, and make the solution build

schulz3000:
fix #1422


kblok:
Thanks @schulz3000!

PR #1483, opened 2020-05-14
Add reference Microsoft.NETFramework.ReferenceAssemblies

kblok:


jzabroski:
@kblok What problem are you trying to solve by adding these reference assemblies?

kblok:
@jzabroski The test project is failing with that on macOS.
image

Do you think this has some side effects? Do you have a better idea?

jzabroski:
> @jzabroski The test project is failing with that on macOS.
> image
> 
> Do you think this has some side effects? Do you have a better idea?

I am trying to figure out what you want to achieve... I think I know how to solve pretty much any .NET dependency lib problem at this point in my career, but people often have unstated assumptions in what they want to achieve.  I don't know why you'd want to reference legacy .NET Framework here.  It doesn't make sense to me why you'd do that in a .netcoreapp2.2 TFM context.

kblok:
@jzabroski Puppeteer-Sharp is a netstandard library. Tests are multi-target because they test the same library on .NET Core and .NET Framework. .NET Framework is not legacy yet ;)

jzabroski:
@kblok Right.  I'm following you, but asking questions one at a time to make sure I don't make extra assumptions.

I spent 3 years trying to understand the right way to support .NET Framework and .NET Core.  Suffice to say, you need to do a Condition on the ItemGroup and break out your PackageReferences by which framework you're targetting.  Also, recommend you target netcoreapp3.1 as well, and target netstandard2.1 instead of 2.2, as 2.2 is not a LTS release.

Just trying to help.

kblok:
@jzabroski good idea about moving the reference to the conditional block. I won't upgrade to 3.1 yet.

jzabroski:
You can see how I'm doing it in FluentMigrator, at https://github.com/jzabroski/FluentMigrator/ - its working well locally for the last month and I'm hoping to merge it all soon

PR #1532, opened 2020-08-13
Add Directory.Build.props

SimonCropp:


lib/PuppeteerSharp/PuppeteerSharp.csproj
```diff
@@ -39,11 +39,7 @@
     <CodeAnalysisRuleSet>../PuppeteerSharp.ruleset</CodeAnalysisRuleSet>
   </PropertyGroup>
   <PropertyGroup>
-    <DocumentationFile>bin\$(Configuration)\netstandard2.0\PuppeteerSharp.xml</DocumentationFile>
```
SimonCropp, line 42: no need for full path here. GenerateDocumentationFile sorts that for you. and will also work if u ever decide to multitarget frameworks

lib/PuppeteerSharp.sln
```diff
@@ -1,22 +1,23 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
```
SimonCropp, line 3: let me know if u want the sln reverted back to VS15

kblok: This is perfect!

samples/get-all-links/get-all-links.csproj
```diff
@@ -4,7 +4,6 @@
     <OutputType>Exe</OutputType>
     <TargetFramework>netcoreapp2.1</TargetFramework>
     <RootNamespace>Example</RootNamespace>
-    <LangVersion>7.1</LangVersion>
```
kblok, line 7: These sample projects are not part of the solution. How about setting the LangVersion to 8 in these projects?

SimonCropp: VS walks the dir tree for Directory.Build.props, which is at the root, so it covers all the projects

SimonCropp: https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2019#search-scope

PR #1537, opened 2020-08-14
remove some redundant variables

SimonCropp:


lib/PuppeteerSharp/CDPSession.cs
```diff
@@ -198,7 +198,6 @@ internal void OnMessage(ConnectionResponse obj)
             else
             {
                 var method = obj.Method;
-                var param = obj.Params?.ToObject<ConnectionResponseParams>();
```
SimonCropp, line 201: not sure on this one?

lib/PuppeteerSharp.Tests/InputTests/PageWaitForFileChooserTests.cs
```diff
@@ -44,26 +44,26 @@ public async Task ShouldWorkWhenFileInputIsNotAttachedToDOM()
         }
 
         [Fact]
-        public async Task ShouldRespectTimeout()
+        public Task ShouldRespectTimeout()
         {
-            var ex = await Assert.ThrowsAsync<TimeoutException>(() => Page.WaitForFileChooserAsync(new WaitForFileChooserOptions
+            return Assert.ThrowsAsync<TimeoutException>(() => Page.WaitForFileChooserAsync(new WaitForFileChooserOptions
```
kblok, line 49: Let's make it an arrow function.

kblok: Same across the class

lib/PuppeteerSharp.Tests/PageTests/Events/ErrorTests.cs
```diff
@@ -19,10 +19,10 @@ public async Task ShouldThrowWhenPageCrashes()
         {
             string error = null;
             Page.Error += (sender, e) => error = e.Error;
-            var gotoTask = Page.GoToAsync("chrome://crash");
+            await Page.GoToAsync("chrome://crash");
```
kblok, line 22: The idea is not to await the task, so we wait for the error while the task is being executed.

lib/PuppeteerSharp.Tests/PuppeteerTests/PuppeteerConnectTests.cs
```diff
@@ -111,7 +111,7 @@ public async Task ShouldBeAbleToReconnectToADisconnectedBrowser()
         public async Task ShouldBeAbleToConnectToTheSamePageSimultaneously()
         {
             var browserOne = await Puppeteer.LaunchAsync(new LaunchOptions());
-            var browserTwo = await Puppeteer.ConnectAsync(new ConnectOptions
+            await Puppeteer.ConnectAsync(new ConnectOptions
```
kblok, line 114: This test is wrong. According to this https://github.com/puppeteer/puppeteer/blob/054d782c823c363c418bf478b0dbefa26673b358/test/launcher.spec.ts#L592 we should create a page with browserTwo in line 125.

lib/PuppeteerSharp/CDPSession.cs
kblok: interesting, but yeah, looks dead.

kblok:
We're having some fails in AppVeyor

PR #1549, opened 2020-09-09
Addresses deadlock, thread-safety in mixed sync/async apps

jpeirson:
Fixes issue https://github.com/hardkoded/puppeteer-sharp/issues/1354 and should generally address deadlock issues in mixed async/sync apps.

See also this example which hangs in v2.0.4, but does not hang in this PR:

```csharp
        private static async Task RunAsync()
        {
            // Async low-level use
            Console.WriteLine("launching browser");
            await using var browser = await Puppeteer.LaunchAsync(new LaunchOptions()).ConfigureAwait(false);

            Console.WriteLine("creating page");
            await using var page = await browser.NewPageAsync().ConfigureAwait(false);

            Console.WriteLine("navigating");
            await page.GoToAsync("http://ipecho.net/plain", WaitUntilNavigation.DOMContentLoaded).ConfigureAwait(false);

            Console.WriteLine("setting content");
            await page.SetContentAsync("<html><body>REPLACED</body></html>").ConfigureAwait(false);

            // Deep inside an existing mostly sync app...
            DoSomethingSync(page);

            // !! never reaches here
            Console.WriteLine("done");
        }

        private static void DoSomethingSync(Page page)
        {
            Console.WriteLine("getting content");
            Console.WriteLine(page.GetContentAsync().ConfigureAwait(false).GetAwaiter().GetResult());
        }
```

Ideally, we [shouldn't block on async code](https://blog.stephencleary.com/2012/07/dont-block-on-async-code.html) but many existing apps (e.g., mine) contain a lot of legacy code and aren't async all the way up, so calling async code synchronously is a valid use case here IMO.


lib/PuppeteerSharp/CDPSession.cs
```diff
@@ -184,22 +184,32 @@ internal void OnMessage(ConnectionResponse obj)
...
+                    });
                 }
             }
             else
             {
                 var method = obj.Method;
                 var param = obj.Params?.ToObject<ConnectionResponseParams>();
 
+                // Always call sync, since we need message (event) order to be determinate.
                 MessageReceived?.Invoke(this, new MessageEventArgs
```
jpeirson, line 213: Note that I left this unchanged since otherwise Chrome events would be raised out of order.

lib/PuppeteerSharp/CDPSession.cs
```diff
@@ -184,22 +184,32 @@ internal void OnMessage(ConnectionResponse obj)
         {
             var id = obj.Id;
 
-            if (id.HasValue && _callbacks.TryRemove(id.Value, out var callback))
+            if (id.HasValue)
```
jpeirson, line 187: I split this `if` to make this logic match the similar code in Connection.cs. If this is wrong, I can revert.

lib/PuppeteerSharp/FrameManager.cs
```diff
@@ -13,7 +13,7 @@ namespace PuppeteerSharp
 {
     internal class FrameManager
     {
-        private Dictionary<int, ExecutionContext> _contextIdToContext;
+        private ConcurrentDictionary<int, ExecutionContext> _contextIdToContext;
```
jpeirson, line 16: After testing my changes, I found races when accessing this dictionary, so I replaced with `ConcurrentDictionary`.

lib/PuppeteerSharp/Connection.cs
```diff
@@ -184,7 +183,7 @@ internal void Close(string closeReason)
         #region Private Methods
 
         private async void Transport_MessageReceived(object sender, MessageReceivedEventArgs e)
-            => await _callbackQueue.Enqueue(() => ProcessMessage(e)).ConfigureAwait(false);
+            => await ProcessMessage(e).ConfigureAwait(false);
```
jpeirson, line 186: This was necessary to prevent marshalling WS messages one by one; otherwise, my example would still have a deadlock while waiting for the semaphore before handling the next message.

jpeirson: Found that this change wasn't necessary, and caused Issue0716 test to fail. Reverted.

jpeirson:
Still getting one test failure, which I'm assuming was b/c of my changes. Will review this last failing test before marking this as ready for review...

kblok:
@jpeirson thank you for your contribution. Looking forward to some green builds :)

lib/PuppeteerSharp.Tests/CSSCoverageTests/CSSCoverageTests.cs
```diff
@@ -63,6 +63,11 @@ public async Task ShouldReportStylesheetsThatHaveNoCoverage()
         {
             await Page.Coverage.StartCSSCoverageAsync();
             await Page.GoToAsync(TestConstants.ServerUrl + "/csscoverage/unused.html");
+
```
jpeirson, line 66: This was the test that was failing some of the time. I'm not sure if I changed something to cause it to fail, or if this was failing before due to a race.

jpeirson:
@kblok were the linux tests failing before? A few are failing with:

```
PuppeteerSharp.ProcessException : Failed to launch Base! LaunchProcess: failed to execvp:
/home/appveyor/projects/puppeteer-sharp-esipd/lib/PuppeteerSharp.Tests/bin/Release/netcoreapp2.2/.local-chromium/Linux-706915/chrome-linux/nacl_helper
[1:1:0910/215617.421195:ERROR:nacl_fork_delegate_linux.cc(323)] Bad NaCl helper startup ack (0 bytes)
Received signal 11 SEGV_MAPERR fffff6685ab5b63e
```

lib/PuppeteerSharp/CDPSession.cs
```diff
@@ -184,22 +184,32 @@ internal void OnMessage(ConnectionResponse obj)
...
+                    {
+                        if (obj.Error != null)
+                        {
+                            callback.TaskWrapper.TrySetException(new MessageException(callback, obj.Error));
+                        }
+                        else
+                        {
+                            callback.TaskWrapper.TrySetResult(obj.Result);
+                        }
+                    });
```
kblok, line 204: My main concern here is the ASP.NET Full Framework Scenario. Could you test [this project](https://github.com/hardkoded/puppeteer-sharp/tree/master/samples/PupppeterSharpAspNetFrameworkSample) using these changes?

kblok (changes requested):

kblok:
@jpeirson Don't look at Linux fails. They are failing for a few months.

lib/PuppeteerSharp/CDPSession.cs
jpeirson: I only suspected that it was an issue, but I haven't observed any issue with the way it was. I'll revert it.

lib/PuppeteerSharp/CDPSession.cs
jpeirson: I'm not totally sure if I set up the sample correctly, but got it to run locally with IIS Express (since I don't have full IIS installed locally), and if I open in my own browser /PupppeterSharpAspNetFrameworkSample/api/test, I get back `"\"bar\""`. I set breakpoints in the debugger to at least make sure my `Task.Run()` tasks were being called and they are.

Is there anything else you can think of that I should test?

jpeirson: @kblok any luck on testing this in full IIS?

lib/PuppeteerSharp.Tests/CSSCoverageTests/CSSCoverageTests.cs
kblok: We shouldn't change this test.

kblok:
@jpeirson checking.

kblok:
We are good on IIS. But we need to pass the tests with no changes.

lib/PuppeteerSharp.Tests/CSSCoverageTests/CSSCoverageTests.cs
jpeirson: Reverted. I retested locally against both this PR and against master, and after 10 tried couldn't get it to fail, so I'm not sure why it was failing before on this PR.

kblok:
It seems that the failure is consistent on that test.

jpeirson:
Yep I see it now in the AppVeyor build. Will try to repro locally and see what's going on.

jpeirson:
@kblok I think the CSS test failure is a preexisting one. If I instead run these tests from the master branch, but make one change to [CSSCoverage](https://github.com/hardkoded/puppeteer-sharp/blob/master/lib/PuppeteerSharp/PageCoverage/CSSCoverage.cs#L105-L107) to simulate a race condition between stopping CSS coverage and the "CSS.styleSheetAdded" event being received, most of the tests fail.

I can reproduce the test failures by changing this in CSSCoverage:

```csharp
case "CSS.styleSheetAdded":
    await Task.Delay(1).ConfigureAwait(false); // <-- temp delay to repro race condition
    await OnStyleSheetAddedAsync(e.MessageData.ToObject<CSSStyleSheetAddedResponse>(true)).ConfigureAwait(false);
    break;
```

My guess is that the CSSCoverageTests GoTo a test page and immediately stop coverage, without actually giving the CSS file a chance to load.

Thoughts?

kblok:
I think [this change](https://github.com/hardkoded/puppeteer-sharp/pull/1549/files#diff-4a058ddb2ad9cb3e82e451d5e5963da8R259) is processing responses in parallel, which breaks the expected flow.

jpeirson:
The tests fail on master if you slow down the responses received from Chrome, even without the changes from this PR - the sample I change from my comment was made to master only. I think the existing CSSCoverageTests merely pass most of the time.

kblok:
re-running the builds

lib/PuppeteerSharp/PageCoverage/CSSCoverage.cs
```diff
@@ -37,8 +42,11 @@ internal Task StartAsync(CoverageStartOptions options)
 
             _resetOnNavigation = options.ResetOnNavigation;
             _enabled = true;
-            _stylesheetURLs.Clear();
-            _stylesheetSources.Clear();
+            lock (_stylesheetsLock)
```
jpeirson, line 45: Added this because async CDP changes from this PR means access to these dictionaries is no longer guaranteed to be thread-safe.

lib/PuppeteerSharp/PageCoverage/CSSCoverage.cs
```diff
@@ -57,10 +65,15 @@ internal async Task<CoverageEntry[]> StopAsync()
             _enabled = false;
 
             var trackingResponse = await _client.SendAsync<CSSStopRuleUsageTrackingResponse>("CSS.stopRuleUsageTracking").ConfigureAwait(false);
+
```
jpeirson, line 68: Fixes what I think is a pre-existing issue where the styleSheetAdded signal isn't necessarily received by the time the code reaches this point. I think this is pre-existing b/c if you inject a `await Task.Delay(10)` inside of `OnStyleSheetAddedAsync`  (i.e. simulates a slow response returned by Chromium), then the corresponding unit tests for this class fail. My async changes from this PR seem to exacerbate this issue.

The DeferredTaskQueue and calling DrainAsync() before gathering coverage results are I think what fixes this issue.

lib/PuppeteerSharp/Helpers/SendAsyncResponseQueue.cs
```diff
@@ -0,0 +1,105 @@
...
+namespace PuppeteerSharp.Helpers
+{
+    /// <summary>
+    /// Provides an async queue for responses for <see cref="CDPSession.SendAsync"/>, so that responses can be handled
+    /// async without risk callers causing a deadlock.
+    /// </summary>
+    /// <remarks>
+    /// See https://github.com/hardkoded/puppeteer-sharp/issues/1354
+    /// </remarks>
+    internal class SendAsyncResponseQueue : IDisposable
```
jpeirson, line 17: This class is mostly a refactoring of the initial commit I had with `Task.Run(() => ...)`, but also adds unhandled exception handling, and wrangles any pending tasks if the class that is handling the response async is disposed / closed.

lib/PuppeteerSharp/IConnectionOptions.cs
```diff
@@ -41,5 +41,10 @@ public interface IConnectionOptions
         /// If not <see cref="Transport"/> is set this will be use to determine is the default <see cref="WebSocketTransport"/> will enqueue messages.
         /// </summary>
         bool EnqueueTransportMessages { get; set; }
+
+        /// <summary>
+        /// Affects how responses to <see cref="CDPSession.SendAsync"/> are returned to the caller.
+        /// </summary>
+        bool EnqueueSendAsyncResponses { get; set; }
```
jpeirson, line 48: I added this as an option, kind of like EnqueueTransportMessages, to let end-users revert to the previous behavior. After fixing the race in CSSCoverage, I'm concerned that users of pptr# will have similar code that assumes CDP responses are handled synchronously, or maybe some scenario that could race but is made worse with this PR.

I can revert this and make this behavior permanent instead, but wanted to try to mitigate broken downstream apps and libs.

jpeirson:
@kblok I think I have this squared away now. I've made some fixes, some cleanup, and added a launch option to turn this new async behavior off. Please take a look.

lib/PuppeteerSharp/ConnectOptions.cs
```diff
@@ -73,5 +73,17 @@ public class ConnectOptions : IBrowserOptions, IConnectionOptions
...
+        /// Affects how responses to <see cref="CDPSession.SendAsync"/> are returned to the caller. If <c>true</c> (default), the
+        /// response is delivered to the caller on its own thread; otherwise, the response is delivered the same way <see cref="CDPSession.MessageReceived"/>
+        /// events are raised.
+        /// </summary>
+        /// <remarks>
+        /// This should normally be set to <c>true</c> (default) to support usage by applications that aren't <c>async</c> "all the way up"; i.e., the application
+        /// has legacy code that is not async which makes calls into PuppeteerSharp. If you experience issues, or your application is not mixed sync/async use, you
+        /// can set this to <c>false</c>.
+        /// </remarks>
+        public bool EnqueueSendAsyncResponses { get; set; } = true;
```
kblok, line 87: `false` by default.

kblok: I would call this `EnqueueAsyncMessages`.

lib/PuppeteerSharp/Connection.cs
```diff
@@ -245,26 +249,23 @@ private void ProcessIncomingMessage(ConnectionResponse obj)
             if (!string.IsNullOrEmpty(obj.SessionId))
             {
                 var session = GetSession(obj.SessionId);
-                session.OnMessage(obj);
+                session?.OnMessage(obj);
```
kblok, line 252: When would this be null?

lib/PuppeteerSharp/Connection.cs
```diff
@@ -85,6 +86,8 @@ internal Connection(string url, int delay, IConnectionTransport transport, ILogg
         /// <value>The logger factory.</value>
         public ILoggerFactory LoggerFactory { get; }
 
+        internal SendAsyncResponseQueue PendingSendAsyncResponses { get; }
```
kblok, line 89:
```suggestion
        internal SendAsyncResponseQueue MessageQueue { get; }
```

lib/PuppeteerSharp/Helpers/SendAsyncResponseQueue.cs
kblok: Let's call ti `AsyncMessageQueue`.

lib/PuppeteerSharp/Helpers/SendAsyncResponseQueue.cs
```diff
@@ -0,0 +1,105 @@
...
+
+namespace PuppeteerSharp.Helpers
+{
+    /// <summary>
+    /// Provides an async queue for responses for <see cref="CDPSession.SendAsync"/>, so that responses can be handled
+    /// async without risk callers causing a deadlock.
+    /// </summary>
+    /// <remarks>
+    /// See https://github.com/hardkoded/puppeteer-sharp/issues/1354
+    /// </remarks>
```
kblok, line 16:
```suggestion
```

lib/PuppeteerSharp/CDPSession.cs
```diff
@@ -186,20 +186,17 @@ internal void OnMessage(ConnectionResponse obj)
...
-                {
-                    callback.TaskWrapper.TrySetException(new MessageException(callback, obj.Error));
-                }
-                else
-                {
-                    callback.TaskWrapper.TrySetResult(obj.Result);
-                }
+                // This is a response to a SendAsync. Handle the callback async, since (a) we can, and (b)
+                // to avoid a situation where the continuation tries to call SendAsync again thus creating
+                // a deadlock.
```
kblok, line 191:
```suggestion
```

lib/PuppeteerSharp/CDPSession.cs
```diff
@@ -186,20 +186,17 @@ internal void OnMessage(ConnectionResponse obj)
...
+                Connection.PendingSendAsyncResponses.Enqueue(callback, obj);
             }
             else
             {
                 var method = obj.Method;
                 var param = obj.Params?.ToObject<ConnectionResponseParams>();
 
+                // Always call sync, since we need message (event) order to be determinate.
```
kblok, line 199:
```suggestion
```

lib/PuppeteerSharp/Connection.cs
```diff
@@ -21,7 +21,7 @@ public class Connection : IDisposable
         private readonly ILogger _logger;
         private TaskQueue _callbackQueue = new TaskQueue();
 
-        internal Connection(string url, int delay, IConnectionTransport transport, ILoggerFactory loggerFactory = null)
+        internal Connection(string url, int delay, bool enqueueSendAsyncResponses, IConnectionTransport transport, ILoggerFactory loggerFactory = null)
```
kblok, line 24: `enqueueAsyncMessages`

lib/PuppeteerSharp/Connection.cs
```diff
@@ -245,26 +249,23 @@ private void ProcessIncomingMessage(ConnectionResponse obj)
...
-                    {
-                        callback.TaskWrapper.TrySetException(new MessageException(callback, obj.Error));
-                    }
-                    else
-                    {
-                        callback.TaskWrapper.TrySetResult(obj.Result);
-                    }
+                    // This is a response to a SendAsync. Handle the callback async, since (a) we can, and (b)
+                    // to avoid a situation where the continuation tries to call SendAsync again thus creating
+                    // a deadlock.
```
kblok, line 262:
```suggestion
```

lib/PuppeteerSharp/Connection.cs
```diff
@@ -245,26 +249,23 @@ private void ProcessIncomingMessage(ConnectionResponse obj)
...
+                    PendingSendAsyncResponses.Enqueue(callback, obj);
                 }
             }
             else
             {
+                // Always call sync, since we need message (event) order to be determinate.
```
kblok, line 268:
```suggestion
```

lib/PuppeteerSharp/Helpers/DeferredTaskQueue.cs
```diff
@@ -0,0 +1,79 @@
...
+                {
+                    if (_pendingTasks.Count == 0)
+                    {
+                        break;
+                    }
+
+                    task = _pendingTasks[0];
+                }
+
+                // TODO: try/catch?
```
kblok, line 62:
```suggestion
```

lib/PuppeteerSharp/IConnectionOptions.cs
kblok: Same rename here.

lib/PuppeteerSharp/LaunchOptions.cs
```diff
@@ -159,5 +159,17 @@ public string[] IgnoredDefaultArgs
...
+        /// Affects how responses to <see cref="CDPSession.SendAsync"/> are returned to the caller. If <c>true</c> (default), the
+        /// response is delivered to the caller on its own thread; otherwise, the response is delivered the same way <see cref="CDPSession.MessageReceived"/>
+        /// events are raised.
+        /// </summary>
+        /// <remarks>
+        /// This should normally be set to <c>true</c> (default) to support usage by applications that aren't <c>async</c> "all the way up"; i.e., the application
+        /// has legacy code that is not async which makes calls into PuppeteerSharp. If you experience issues, or your application is not mixed sync/async use, you
+        /// can set this to <c>false</c>.
+        /// </remarks>
+        public bool EnqueueSendAsyncResponses { get; set; } = true;
```
kblok, line 173: false by default.

lib/PuppeteerSharp/PageCoverage/CSSCoverage.cs
kblok: Can't we have a concurrent dictionary?

kblok (changes requested):

lib/PuppeteerSharp/Connection.cs
jpeirson: This can technically be null since `_sessions.GetValueOrDefault(sessionId)` can return null, but I encountered this once during a test run, which seemed to happen at random.

Is this a possible path?
```csharp
            else if (method == "Target.detachedFromTarget")
            {
                // -> exec
                var sessionId = param.SessionId; // sessionId = "X"
                if (_sessions.TryRemove(sessionId, out var session) && !session.IsClosed)
                {
                    // -> exec
                    session.Close("Target.detachedFromTarget");
                }
            }

            if (!string.IsNullOrEmpty(obj.SessionId))
            {
                // -> exec
                var session = GetSession(obj.SessionId); // obj.SessionId = "X"
                // session = null since it was just removed above
                session?.OnMessage(obj);
            }
```

Not sure when/why `param.SessionId == obj.SessionId` though.

lib/PuppeteerSharp.Tests/TestConstants.cs
```diff
@@ -43,6 +43,7 @@ public static class TestConstants
         {
             SlowMo = Convert.ToInt32(Environment.GetEnvironmentVariable("SLOW_MO")),
             Headless = Convert.ToBoolean(Environment.GetEnvironmentVariable("HEADLESS") ?? "true"),
+            EnqueueSendAsyncResponses = Convert.ToBoolean(Environment.GetEnvironmentVariable("ENQUEUE_SENDASYNC_RESPONSES") ?? "true"),
```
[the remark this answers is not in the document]
jpeirson: I'll default this to false, and will take a stab at adding this as a variation in appveyor...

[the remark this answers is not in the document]
jpeirson: One issue I just noticed is that the tests for [Issue1354](https://github.com/hardkoded/puppeteer-sharp/pull/1549/files#diff-30ed2f8333676928330b164a382ecdadb90d222a696e5e316e232a311a5dd486R8) will always fail unless this option is set to true.

Thoughts on how to address that? Make the tests skip if that option is false? Force the option to true just within those two tests?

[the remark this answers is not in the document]
kblok: We can make the test pass if that variable is false.

appveyor.yml
```diff
@@ -10,6 +10,10 @@ environment:
   matrix:
     - framework: net48
     - framework: netcoreapp2.2
+    - framework: net48
+      ENQUEUE_ASYNC_MESSAGES: true
+    - framework: netcoreapp2.2
+      ENQUEUE_ASYNC_MESSAGES: true
```
kblok, line 16:
```suggestion
    - framework: net48
      ENQUEUE_ASYNC_MESSAGES: true
    - framework: netcoreapp2.2
      ENQUEUE_ASYNC_MESSAGES: true
    - framework: net48
      ENQUEUE_ASYNC_MESSAGES: false
    - framework: netcoreapp2.2
      ENQUEUE_ASYNC_MESSAGES: false
```

lib/PuppeteerSharp/CDPSession.cs
```diff
@@ -226,10 +219,11 @@ internal void Close(string closeReason)
             }
 
             _callbacks.Clear();
+
```
kblok, line 222:
```suggestion
```

jpeirson:
@kblok is there anything else I should do for this PR?

kblok:
Sorry @jpeirson. I've been quite busy. I'll get back to this tomorrow.

mrtristan:
@kblok any plans to cut a release? i'm hitting this deadlock that was solved here... :-(

If not, can you simply cut a "preview" or dev build so we can take advantage of some of the more recent updates?

thanks!

kblok:
Hey @mrtristan, Hopefully, I will release a new version in the following days.

PR #1559, opened 2020-10-01
Set CA2213 rule to Error

### LINKED ISSUE #1441: Enable CA2213 rule
Change the action for CA2213 on the [ruleset file](https://github.com/kblok/puppeteer-sharp/blob/master/lib/PuppeteerSharp.ruleset) to Error, and make the solution build

bananbe:
Set CA2213 rule to Error and adjusted code to this change.

Fixes #1441

### DIFF
--- a/lib/PuppeteerSharp.ruleset
+++ b/lib/PuppeteerSharp.ruleset
@@ -79,7 +79,7 @@
     <Rule Id="CA2000" Action="Error" /> <!-- Error CA2000: Call System.IDisposable.Dispose on object created by 'new Process()' before all references to it are out of scope. (CA2000) -->
     <Rule Id="CA2008" Action="None" /> <!-- Error ca2008: Pass TaskScheduler-->
     <Rule Id="CA2200" Action="None" /> <!-- Error CA2200: Re-throwing caught exception changes stack information. (CA2200)  -->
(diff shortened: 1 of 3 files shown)


kblok:
Thank you @bananbe!

bananbe:
There's one issue with this rule - it's not passing in [https://github.com/hardkoded/puppeteer-sharp/blob/master/lib/PuppeteerSharp/Transport/WebSocketTransport.cs](url), becuase `Dispose` it's not called on `_readerCancellationSource`. 

In fact it is done in `StopReading()` method, but `Cancel()` and `Dispose()` is called on `readerCts`, which contains `_readerCancellationSource`, so the rule is fulfilled, but not directly.

Do you some suggestions for that?

kblok:
@bananbe it doesn't make any sense. The code is right. Can we ignore the rule in that line?

kblok:
@bananbe We are getting a code style error we just merged. Could you pull from master a build again?

Thanks :)

bananbe:
Sure, I will do that 👍

bananbe:
@kblok One test failed because of timeout. I ran them locally and all, including the one which failed, are passing. Is it something specific to environment, that this is happend? Or this one is failing at random from time to time?

kblok:
@bananbe yeah, I'm running the builds again ;)

kblok:
Thanks!

PR #1571, opened 2020-10-13
Set SA1612 rule to Error

### LINKED ISSUE #1419: Enable SA1612 rule
Change the action for SA1612 on the [ruleset file](https://github.com/kblok/puppeteer-sharp/blob/master/lib/PuppeteerSharp.ruleset) to Error, and make the solution build

bananbe:
Set SA1612 rule to Error and adjusted code to this change.

Fixes #1419


kblok:
Thanks!

PR #1572, opened 2020-10-15
Set CA1816 to error

qtuu:
When I had a look at the class, i saw that it did not need a finalizer. LifecycleWatcher does not contain any unmanaged members. See [implementing dispose](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose) for reference: "It is possible for a base class to only reference managed objects, and implement the dispose pattern. In these cases, a finalizer is unnecessary. A finalizer is only required if you directly reference unmanaged resources."

In a second commit I removed the second dispose(bool) method and the finalizer.

I am uncertain about the SuppressMessage for CA2213:DisposableFieldsShouldBeDisposed in line 36 of the same file. The _terminationCancellationToken is only canceled in our dispose-method but never actually disposed. Can you elaborate on that?


qtuu:
i forgot to mention: this fixes issue #1437 ;D

kblok:
Thank you @qtuu  :)

kblok:
Thank you!

qtuu:
@kblok could you please add the hacktoberfest-accepted label to this PR, so it gets counted towards my progress? :)

kblok:
@qtuu Done!

PR #1573, opened 2020-10-15
Set ca2200 to error

qtuu:
this fixes issue #1439 

I did not add a unit test because inspecting the stack of a caught exception would end up being messy and flaky.


qtuu:
@kblok thank you very much :)

Issue #1427, opened 2020-02-29
Enable CA1064 rule


kblok:
Change the action for CA1064 on the [ruleset file](https://github.com/kblok/puppeteer-sharp/blob/master/lib/PuppeteerSharp.ruleset) to Error, and make the solution build

kgar:
Looks like this issue can be closed.

PR #1578, opened 2020-10-25
Enable ca1064 rule

qtuu:
this fixes issue #1427 . the class that was marked as an error was never used, so i removed it.
I noticed that there are some exception-classes inheriting from PuppeteerException and are thus not marked by the rule. Should i take a look at those as well?


qtuu:
the microsoft edge build seems to be using a different version of the .net sdk. C#8 should be supported with .net core 3.1 onwards i think...

anyway.. could you please add the hacktoberfest-accepted label to this PR, so it gets counted towards my progress? ;)

kblok:
> the microsoft edge build seems to be using a different version of the .net sdk. C#8 should be supported with .net core 3.1 onwards i think...

Don't worry about that build. It's a playground.

> anyway.. could you please add the hacktoberfest-accepted label to this PR, so it gets counted towards my progress? ;)

Sure!

Issue #1542, opened 2020-08-19
ResponseData.Headers can't deal with multiple headers with same name


L6Echo:
### Description

The Headers field of the ResponseData is a map of header name to header value. This assumes that every header type can be set only once, but this is assumption is not correct. E.g. the Set-Cookie may be listen multiple times.

In System.Net.Http this is solved by using a Map which maps the header name to a list of values.

### Complete minimal example reproducing the issue

When intercepting the http request, it is not possible to set multiple cookies:

E.g.

```csharp
        private async Task InterceptAsync(RequestEventArgs args)
        {
            var responseData = new ResponseData
            {
                Body = "{\"success\": true}",
                ContentType = "application/json; charset=utf-8",
                Headers = new Dictionary<string, object>(),
                Status = HttpStatusCode.OK
            };
            
            responseData.Headers.Add("Set-Cookie", "session=xyz; path=/; samesite=none");
            responseData.Headers.Add("Set-Cookie", "addition=abc; path=/; samesite=none");

            await args.Request.RespondAsync(responseData);
        }
```

### Expected behavior:

It should be possible to set headers multiple times with same name.

### Actual behavior:

Only one value per header name can be added.

### Versions

* PuppeteerSharp: 2.0.4
* .NET: .NET Core 3.1.3

kblok:
I believe [Chrome devtools protocol](https://chromedevtools.github.io/devtools-protocol/tot/Fetch/#type-HeaderEntry)  won't support duplicates there. 
Won't `Page.SetCookieAsync` work for you?

L6Echo:
The headers property is an Array of HeaderEntries. So it should be possible to set multiple headers with the same name. Except the protocol checks for uniqueness of header names. But I wouldn't expect this. Because it is a legit case.

Yes, currently i did a workaround with SetCookieAsync. But this bypasses the cookie handling of chrome and this may distort the test result. Additional the cookie header must be parsed manually which may also behave different than the chrome cookie handling itself.

venky8951:
Hi @kblok can I start looking into this ? seems like we need to implement RespondAsync for List<strings>(values)

kblok:
If you can make it work without breaking any test, go for it!

venky8951:
Hi @kblok I have tried couple of ways to address this issue but it seems having Dictionary<string,Object> is difficult to parse
can we use a List<KeyValuePair<string, object>> instead of a Dictionary<string,Object> ?

[...]

PR #1607, opened 2020-12-28
Added Support to Multiple Header with same name 

venky8951:
#1542 
Hi @kblok kindly review the changes, I have tried not to break the API kindly advice if it can be done better


lib/PuppeteerSharp/Request.cs
```diff
@@ -230,24 +230,31 @@ public async Task RespondAsync(ResponseData response)
 
             _interceptionHandled = true;
 
-            var responseHeaders = new Dictionary<string, string>();
+            var responseHeaders = new List<KeyValuePair<string, string>>();
```
kblok, line 233: Did you try using a passing `StringComparer.OrdinalIgnoreCase` to the Dictionary?
I would like to avoid having another property.

venky8951: Hi @kblok by passing that we can make Dictionary case insensitive but here we want Duplicate keys
like Set cookies header name can occur twice 
any further advice?

kblok: Yeah, that's true, good catch.

kblok:
Can we add a test?

venky8951:
Hi @kblok ya sure I will add a test for multiple header 
But one of the case in Core+Net Framework test seems to fail which has nothing to do with ResponseData 
its ElementHandleTests\BoundingBoxTests I tried to find out reason I think Frame is getting to null . what might be the remedy? any suggestions

kblok:
@venky8951 don't worry about the failing test.

lib/PuppeteerSharp/Request.cs
```diff
@@ -313,5 +320,11 @@ public async Task AbortAsync(RequestAbortErrorCode errorCode = RequestAbortError
 
         private Header[] HeadersArray(Dictionary<string, string> headers)
             => headers?.Select(pair => new Header { Name = pair.Key, Value = pair.Value }).ToArray();
+
+        private Header[] HeadersArray(List<KeyValuePair<string, string>> headers)
+            => headers?.Select(pair => new Header { Name = pair.Key, Value = pair.Value }).ToArray();
+
+        private bool ContainsKey(List<KeyValuePair<string, string>> headers, string key)
+           => headers?.Where(item => item.Key == key).ToList().Count > 0 ? true : false;
```
fr4gles, line 328: If I may add an opinion - you could rewrite this `headers?.Where(item => item.Key == key).ToList().Count` as follows `headers?.Count(item => item.Key == key)`. There is no need to create new list object while we can use LINQ 😄

lib/PuppeteerSharp/ResponseData.cs
```diff
@@ -39,5 +39,10 @@ public string Body
         /// </summary>
         /// <value>Status Code.</value>
         public HttpStatusCode? Status { get; set; }
+        /// <summary>
+        /// Multiple Response headers with same Header name. Header values will be converted to a string.
+        /// </summary>
+        /// <value>Headers.</value>
+        public List<KeyValuePair<string, object>> MultipleHeaders { get; set; }
```
amaitland, line 46: Using [NameValueCollection](https://docs.microsoft.com/en-us/dotnet/api/system.collections.specialized.namevaluecollection?view=netstandard-2.0) would likely be nicer here, it supports multiple values which can be obtained via `GetValues`. It's part of `.Net Standard 2.0`.

amaitland: Using https://docs.microsoft.com/en-us/dotnet/api/system.net.http.headers.httpheaders?view=net-5.0 would be even better if it's possible.

kblok:
closing in favor of https://github.com/hardkoded/puppeteer-sharp/pull/1831

PR #1616, opened 2021-01-27
Code refactoring - Removed unnecessary usings, added regions and comments

gauravgupta98:


lib/PuppeteerSharp/ContextPayload.cs
```diff
@@ -1,13 +1,20 @@
...
 
+        /// <summary>
+        /// Gets or sets the context payload aux data.
+        /// </summary>
         public ContextPayloadAuxData AuxData { get; set; }
 
+        /// <summary>
+        /// Gets or sets the name.
+        /// </summary>
         public string Name { get; set; }
```
kblok, line 18: We are not adding XML Docs for internal classes.

lib/PuppeteerSharp/EmulationManager.cs
```diff
@@ -1,21 +1,29 @@
...
 
 namespace PuppeteerSharp
 {
     internal class EmulationManager
     {
+        #region Members
         private readonly CDPSession _client;
         private bool _hasTouch;
         private bool _emulatingMobile;
+        #endregion Members
```
kblok, line 13: I'm trying to get out of the `#region` blocks. The community is not using them anymore.

kblok (changes requested):
Thanks for taking the time to improve the code.
I like removing using, but I don't think we need XML docs for the internals classes. And it seems that the community stop using regions so I might end up removing them.

lib/PuppeteerSharp/ContextPayload.cs
gauravgupta98: Thanks for the review @kblok.

lib/PuppeteerSharp/EmulationManager.cs
gauravgupta98: Wasn't aware of this. Thanks for letting know :)

gauravgupta98:
Closing the PR as we don't need the suggested changes

PR #1624, opened 2021-02-13
show md snippets usage

SimonCropp:
replaces by https://github.com/hardkoded/puppeteer-sharp/pull/1668


kblok:
That's a pretty cool idea. I love it. Although using regions for that feels odd. But I get the point.

kblok:
Don't we need to set up anything to make that action work?

SimonCropp:
u can use comments to delimitate snippets. that is what is used in other languages. but regions give u a enforced start and end. and u can set them to always expand

the action should need no other changes

SimonCropp:
thoughts on this one?

kblok:
I'm trying to wrap v3. I will get back to this after v3 and v8 plan.

kblok:
This would be ready to go once unconflicted @SimonCropp

SimonCropp:
repalced by https://github.com/hardkoded/puppeteer-sharp/pull/1668

PR #1625, opened 2021-02-13
fix some nulls

SimonCropp:


lib/PuppeteerSharp.DevicesFetcher/Program.cs
```diff
@@ -44,6 +44,7 @@ static async Task Main(string[] args)
             catch (Exception ex)
             {
                 Console.WriteLine($"FAILED: error parsing response - {ex.Message}");
+                return;
```
SimonCropp, line 47: if u dont return, u will get a null ref on the next line

lib/PuppeteerSharp.Tests/BrowserContextTests/DefaultBrowserContextTests.cs
```diff
@@ -29,7 +29,7 @@ public async Task PageGetCookiesAsyncShouldWork()
             await Page.GoToAsync(TestConstants.EmptyPage);
 
             await Page.EvaluateExpressionAsync("document.cookie = 'username=John Doe'");
-            var cookie = (await Page.GetCookiesAsync()).FirstOrDefault();
+            var cookie = (await Page.GetCookiesAsync()).First();
```
SimonCropp, line 32: no null check after a `FirstOrDefault` implies it can never be empty, and hence should be a `First`

lib/PuppeteerSharp.Tests/BrowserContextTests/DefaultBrowserContextTests.cs
```diff
@@ -52,7 +52,7 @@ public async Task PageSetCookiesAsyncShouldWork()
                 Value = "John Doe"
             });
 
-            var cookie = (await Page.GetCookiesAsync()).FirstOrDefault();
+            var cookie = (await Page.GetCookiesAsync()).First();
```
SimonCropp, line 55: no null check after a `FirstOrDefault` implies it can never be empty, and hence should be a `First`

lib/PuppeteerSharp.Tests/BrowserContextTests/DefaultBrowserContextTests.cs
```diff
@@ -88,7 +88,7 @@ public async Task PageDeleteCookieAsyncShouldWork()
             });
             Assert.Equal("cookie1=1", await Page.EvaluateExpressionAsync<string>("document.cookie"));
 
-            var cookie = (await Page.GetCookiesAsync()).FirstOrDefault();
+            var cookie = (await Page.GetCookiesAsync()).First();
```
SimonCropp, line 91: no null check after a `FirstOrDefault` implies it can never be empty, and hence should be a `First`

lib/PuppeteerSharp.Tests/ElementHandleTests/BoxModelTests.cs
```diff
@@ -29,7 +29,7 @@ public async Task ShouldWork()
 
             // Step 2: Add div and position it absolutely inside frame.
             var frame = Page.FirstChildFrame();
-            var divHandle = (await frame.EvaluateFunctionHandleAsync(@"() => {
+            var divHandle = (ElementHandle)await frame.EvaluateFunctionHandleAsync(@"() => {
```
SimonCropp, line 32: no null check after an `as` implied it can only be that type. hence a cast is better, since if the assumption is incorrect you will get a useful cast exception instead of a useless null ref exception

lib/PuppeteerSharp.Tests/ElementHandleTests/ClickTests.cs
```diff
@@ -25,7 +25,7 @@ public async Task ShouldWork()
         public async Task ShouldWorkForShadowDomV1()
         {
             await Page.GoToAsync(TestConstants.ServerUrl + "/shadow.html");
-            var buttonHandle = (await Page.EvaluateExpressionHandleAsync("button")) as ElementHandle;
+            var buttonHandle = (ElementHandle) await Page.EvaluateExpressionHandleAsync("button");
```
SimonCropp, line 28: no null check after an `as` implied it can only be that type. hence a cast is better, since if the assumption is incorrect you will get a useful cast exception instead of a useless null ref exception

lib/PuppeteerSharp.Tests/ElementHandleTests/ClickTests.cs
```diff
@@ -34,8 +34,8 @@ public async Task ShouldWorkForShadowDomV1()
         public async Task ShouldWorkForTextNodes()
         {
             await Page.GoToAsync(TestConstants.ServerUrl + "/input/button.html");
-            var buttonTextNode = (await Page.EvaluateExpressionHandleAsync(
-                "document.querySelector('button').firstChild")) as ElementHandle;
+            var buttonTextNode = (ElementHandle) await Page.EvaluateExpressionHandleAsync(
```
SimonCropp, line 37: no null check after an `as` implied it can only be that type. hence a cast is better, since if the assumption is incorrect you will get a useful cast exception instead of a useless null ref exception

lib/PuppeteerSharp.Tests/FrameUtils.cs
```diff
@@ -8,14 +8,14 @@ public static class FrameUtils
     {
         public static async Task<Frame> AttachFrameAsync(Page page, string frameId, string url)
         {
-            var handle = await page.EvaluateFunctionHandleAsync(@" async (frameId, url) => {
+            var handle = (ElementHandle)await page.EvaluateFunctionHandleAsync(@" async (frameId, url) => {
```
SimonCropp, line 11: no null check after an `as` implied it can only be that type. hence a cast is better, since if the assumption is incorrect you will get a useful cast exception instead of a useless null ref exception

lib/PuppeteerSharp/BrowserFetcher.cs
```diff
@@ -114,7 +114,7 @@ public async Task<bool> CanDownloadAsync(int revision)
                 var client = WebRequest.Create(url);
                 client.Proxy = _webClient.Proxy;
                 client.Method = "HEAD";
-                using (var response = await client.GetResponseAsync().ConfigureAwait(false) as HttpWebResponse)
+                using (var response = (HttpWebResponse)await client.GetResponseAsync().ConfigureAwait(false))
```
SimonCropp, line 117: no null check after an `as` implied it can only be that type. hence a cast is better, since if the assumption is incorrect you will get a useful cast exception instead of a useless null ref exception

kblok: Did you go one by one or using a tool?
It would be cool to find an analyzer checking this.

kblok (approved):

PR #1627, opened 2021-02-13
Broken xml doc links

SimonCropp:
since there are properties matching those type  names, the types need to be fully qualified


kblok:
Broken for who? It seems that DocFX is not complaining about it.
image

kblok:
Builds are red BTW.

SimonCropp:
the compiler was giving warnings. not sure how docfx resolves them since the type resolution rules mean they should resolve to the locally scoped properties. 

i will look at the build errors tomorrow

PR #1629, opened 2021-02-13
fix bug in ConnectToTheSamePageSimultaneously

SimonCropp:
i think the intent of the test was to test against the two diff instances?


kblok (approved):
Good catch

PR #1631, opened 2021-02-13
remove ToObject<ConnectionResponseParams

SimonCropp:
I assume this is redundant? but technically it could be breaking behavior change


kblok:
Let's see what AppVeyor thinks

PR #1633, opened 2021-02-14
use a static local func

SimonCropp:
https://www.c-sharpcorner.com/article/c-sharp-8-0-static-local-functions/

> In C# 8.0, we addressed this issue by declaring the local functions as static since we are aware that static methods won’t be allocating separate memory in each object and the same in its variables scoped to it.


kblok:
AppVeyor didn't like this

MrSimonC:
This is due to DevicesFetcher targeting .netcore 2.2 (_note: out of Microsoft support now_) which only supports C# 7.3. This issue is fixed either by bumping DevicesFetcher to .net core 3.1, or waiting for the .net 5 issue #1638 to be implemented ...

PR #1643, opened 2021-03-04
chmod +x nacl_helper

Androbin:
Fixes the `Bad NaCl helper startup ack (0 bytes)` error

### DIFF
--- a/lib/PuppeteerSharp/BrowserFetcher.cs
+++ b/lib/PuppeteerSharp/BrowserFetcher.cs
@@ -225,7 +225,7 @@ namespace PuppeteerSharp
                 string dirName = Path.GetDirectoryName(execPath);
 
                 int code = LinuxSysCall.Chmod(execPath, LinuxSysCall.ExecutableFilePermissions);
-                if (code == -1)
+                if (code != 0)
                 {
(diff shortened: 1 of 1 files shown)


kblok:
This is awesome. Thanks a lot.

kblok (approved):

lib/PuppeteerSharp/BrowserFetcher.cs
```diff
@@ -221,11 +221,20 @@ public async Task<RevisionInfo> DownloadAsync(int revision)
...
+                int code = LinuxSysCall.Chmod(execPath, LinuxSysCall.ExecutableFilePermissions);
                 if (code == -1)
                 {
                     throw new Exception("Chmod operation failed");
                 }
+
+                string naclPath = $"{dirName}/nacl_helper";
+                if (File.Exists(naclPath))
+                {
+                    LinuxSysCall.Chmod(naclPath, LinuxSysCall.ExecutableFilePermissions);
```
kblok, line 236: We should check the code Chmod is returning, as we do in line 228

Androbin: done

PR #1646, opened 2021-03-13
Cleanup: Remove BOMs, Use Unix style line endings

Androbin:
Some of the files had a BOM, some did not.
Some of the lines of some of the files used Windows style line endings, some did not.


kblok:
Thank you @Androbin!
Do you want to add an editor config for this?

Androbin:
@kblok Here's what I would suggest:
```
charset = utf-8
end_of_line = lf
insert_final_newline = true
trim_trailing_whitespace = true
```

kblok:
Do you think we can use this https://github.com/microsoft/playwright-sharp/blob/main/src/.editorconfig ?

Androbin:
I have sorted the properties for `[*]` and `[*.cs]` and made a diff:
https://www.diffchecker.com/O1X54bQN

I'll leave the pile of `[*.cs]` settings to you, but I would like to insist on `[*] end_of_line = lf`

kblok:
I forgot we had an editorconfig. I'm good with your settings :)
Let's leave this PR after Firefox and your upcoming network conditions PR, so we avoid conflicts on those PRs.

Androbin:
@kblok Ready when you are

Androbin:
@kblok Ready when you are

PR #1647, opened 2021-03-16
feat: add page.emulateNetworkConditions

Androbin:
See https://github.com/hardkoded/puppeteer-sharp/pull/1645#issuecomment-799692242

### DIFF
--- a/lib/PuppeteerSharp/NetworkManager.cs
+++ b/lib/PuppeteerSharp/NetworkManager.cs
@@ -97,10 +97,10 @@ namespace PuppeteerSharp
         internal async Task SetOfflineModeAsync(bool value)
         {
             _emulatedNetworkConditions.Offline = value;
-            await _updateNetworkConditions().ConfigureAwait(false);
+            await UpdateNetworkConditionsAsync().ConfigureAwait(false);
         }
 
-        internal async Task EmulateNetworkConditions(NetworkConditions networkConditions)
+        internal async Task EmulateNetworkConditionsAsync(NetworkConditions networkConditions)
         {
             _emulatedNetworkConditions.Upload = networkConditions != null
               ? networkConditions.Upload
@@ -111,19 +111,17 @@ namespace PuppeteerSharp
             _emulatedNetworkConditions.Latency = networkConditions != null
               ? networkConditions.Latency
               : 0;
-            await _updateNetworkConditions().ConfigureAwait(false);
+            await UpdateNetworkConditionsAsync().ConfigureAwait(false);
         }
 
-        internal async Task _updateNetworkConditions()
-        {
-            await _client.SendAsync("Network.emulateNetworkConditions", new NetworkEmulateNetworkConditionsRequest
+        private Task UpdateNetworkConditionsAsync()
+            => _client.SendAsync("Network.emulateNetworkConditions", new NetworkEmulateNetworkConditionsRequest
             {
                 Offline = _emulatedNetworkConditions.Offline,
                 Latency = _emulatedNetworkConditions.Latency,
                 UploadThroughput = _emulatedNetworkConditions.Upload,
                 DownloadThroughput = _emulatedNetworkConditions.Download,
-            }).ConfigureAwait(false);
-        }
+            });
 
         internal Task SetUserAgentAsync(string userAgent)
             => _client.SendAsync("Network.setUserAgentOverride", new NetworkSetUserAgentOverrideRequest
--- a/lib/PuppeteerSharp/NetworkManager.cs
+++ b/lib/PuppeteerSharp/NetworkManager.cs
@@ -102,15 +102,9 @@ namespace PuppeteerSharp
 
         internal async Task EmulateNetworkConditionsAsync(NetworkConditions networkConditions)
         {
-            _emulatedNetworkConditions.Upload = networkConditions != null
-              ? networkConditions.Upload
-              : -1;
-            _emulatedNetworkConditions.Download = networkConditions != null
-              ? networkConditions.Download
-              : -1;
-            _emulatedNetworkConditions.Latency = networkConditions != null
-              ? networkConditions.Latency
-              : 0;
+            _emulatedNetworkConditions.Upload = networkConditions?.Upload ?? -1;
+            _emulatedNetworkConditions.Download = networkConditions?.Download ?? -1;
+            _emulatedNetworkConditions.Latency = networkConditions?.Latency ?? 0;
             await UpdateNetworkConditionsAsync().ConfigureAwait(false);
         }
 
--- a/lib/PuppeteerSharp/NetworkManager.cs
+++ b/lib/PuppeteerSharp/NetworkManager.cs
@@ -22,13 +22,19 @@ namespace PuppeteerSharp
         private readonly ConcurrentDictionary<string, string> _requestIdToInterceptionId = new ConcurrentDictionary<string, string>();
         private readonly ILogger _logger;
         private Dictionary<string, string> _extraHTTPHeaders;
-        private bool _offine;
         private Credentials _credentials;
         private readonly List<string> _attemptedAuthentications = new List<string>();
         private bool _userRequestInterceptionEnabled;
         private bool _protocolRequestInterceptionEnabled;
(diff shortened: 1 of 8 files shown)


lib/PuppeteerSharp/NetworkManager.cs
```diff
@@ -90,18 +96,33 @@ internal Task SetExtraHTTPHeadersAsync(Dictionary<string, string> extraHTTPHeade
...
-                    Latency = 0,
-                    DownloadThroughput = -1,
-                    UploadThroughput = -1
-                }).ConfigureAwait(false);
-            }
+        internal async Task EmulateNetworkConditions(NetworkConditions networkConditions)
+        {
+            _emulatedNetworkConditions.Upload = networkConditions != null
+              ? networkConditions.Upload
+              : -1;
```
kblok, line 107: Can't we do something like ` networkConditions?.Upload ?? -1`

lib/PuppeteerSharp/NetworkManager.cs
```diff
@@ -90,18 +96,33 @@ internal Task SetExtraHTTPHeadersAsync(Dictionary<string, string> extraHTTPHeade
...
+        internal async Task _updateNetworkConditions()
+        {
+            await _client.SendAsync("Network.emulateNetworkConditions", new NetworkEmulateNetworkConditionsRequest
+            {
+                Offline = _emulatedNetworkConditions.Offline,
+                Latency = _emulatedNetworkConditions.Latency,
+                UploadThroughput = _emulatedNetworkConditions.Upload,
+                DownloadThroughput = _emulatedNetworkConditions.Download,
+            }).ConfigureAwait(false);
         }
```
kblok, line 126: Let's use an arrow function and return the task instead of making the method async

lib/PuppeteerSharp/NetworkManager.cs
```diff
@@ -90,18 +96,33 @@ internal Task SetExtraHTTPHeadersAsync(Dictionary<string, string> extraHTTPHeade
...
+            _emulatedNetworkConditions.Download = networkConditions != null
+              ? networkConditions.Download
+              : -1;
+            _emulatedNetworkConditions.Latency = networkConditions != null
+              ? networkConditions.Latency
+              : 0;
+            await _updateNetworkConditions().ConfigureAwait(false);
+        }
+
+        internal async Task _updateNetworkConditions()
```
kblok, line 117: Rename to `UpdateNetworkConditionsAsync`

lib/PuppeteerSharp/Page.cs
```diff
@@ -616,6 +616,16 @@ public Task SetRequestInterceptionAsync(bool value)
...
 
+        /// <summary>
+        /// Emulates network conditions
+        /// </summary>
+        /// <param name="networkConditions">Passing <c>null</c> disables network condition emulation.</param>
+        /// <returns>Result task</returns>
+        /// <remarks>
+        /// **NOTE** This does not affect WebSockets and WebRTC PeerConnections (see https://crbug.com/563644)
+        /// </remarks>
+        public Task EmulateNetworkConditions(NetworkConditions networkConditions) => FrameManager.NetworkManager.EmulateNetworkConditions(networkConditions);
```
kblok, line 627: Add Async suffix

lib/PuppeteerSharp/NetworkManager.cs
```diff
@@ -90,18 +96,33 @@ internal Task SetExtraHTTPHeadersAsync(Dictionary<string, string> extraHTTPHeade
...
 
-                await _client.SendAsync("Network.emulateNetworkConditions", new NetworkEmulateNetworkConditionsRequest
-                {
-                    Offline = value,
-                    Latency = 0,
-                    DownloadThroughput = -1,
-                    UploadThroughput = -1
-                }).ConfigureAwait(false);
-            }
+        internal async Task EmulateNetworkConditions(NetworkConditions networkConditions)
```
kblok, line 103: Add Async suffix.

kblok (changes requested):

lib/PuppeteerSharp/NetworkManager.cs
Androbin: I had the same thought but I tried to be faithful to the reference implementation

kblok: No need here :)

lib/PuppeteerSharp/NetworkManager.cs
Androbin: Okay

lib/PuppeteerSharp/NetworkManager.cs
Androbin: Okay

lib/PuppeteerSharp/Page.cs
Androbin: Okay

lib/PuppeteerSharp/NetworkManager.cs
Androbin: Okay

lib/PuppeteerSharp/NetworkManager.cs
Androbin: If we change it like that, we could also make `NetworkConditions.Upload` nullable.
In the reference implementation, null values are prohibited, so we would break with it.

kblok: It shouldn't be nullable because we would default it to -1

Androbin: How about we declare it as `public double Upload { get; set; } = -1;` , so it has a more reasonable default value in case the user doesn't assign one.

kblok: agreed.

lib/PuppeteerSharp/PredefinedNetworkConditions.cs
```diff
@@ -0,0 +1,49 @@
+using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+
+namespace PuppeteerSharp
+{
+    /// <summary>
+    /// Predefined network conditions name.
+    /// </summary>
+    public enum PredefinedNetworkConditionsName
```
kblok, line 10: Let's call this `NetworkCondition`

kblok (changes requested):
This is perfect, just one tiny change

lib/PuppeteerSharp/PredefinedNetworkConditions.cs
Androbin: There was an upstream debate about the singular/plural naming:
https://github.com/puppeteer/puppeteer/pull/6759#discussion_r560055871

kblok:
We have a real fail in appveyor

lib/PuppeteerSharp/PredefinedNetworkConditions.cs
kblok: Good point. It's not about the plural. I don't like much the suffix `Name`what do you think?

Androbin: How about we drop the enum and declare string constants in the `NetworkConditions` class?

kblok: Love it!

Androbin: I took inspiration from [`DeviceDescriptorName.cs`](https://github.com/hardkoded/puppeteer-sharp/blob/master/lib/PuppeteerSharp/Mobile/DeviceDescriptorName.cs)

kblok: Those could be strings with the same values as upstream, what do you think?

Androbin: Sounds good to me

kblok: Let's do it

kblok (approved):

Androbin:
Tests passing

PR #1650, opened 2021-03-16
Fix BoundingBox equality

### LINKED ISSUE #1626: BoundingBox Equals has a null bug
```
        public override bool Equals(object obj)
        {
            if (obj == null && GetType() != obj.GetType())
            {
                return false;
            }

            return Equals((BoundingBox)obj);
        }
```

if `obj == null` is true, `obj.GetType()` will throw

Androbin:
Fixes #1626

### DIFF
--- a/lib/PuppeteerSharp/BoundingBox.cs
+++ b/lib/PuppeteerSharp/BoundingBox.cs
@@ -66,7 +66,7 @@ namespace PuppeteerSharp
         /// <inheritdoc/>
         public override bool Equals(object obj)
         {
-            if (obj == null && GetType() != obj.GetType())
+            if (obj == null || GetType() != obj.GetType())
             {
                 return false;
             }


lib/PuppeteerSharp/BoundingBox.cs
```diff
@@ -66,7 +66,7 @@ internal Clip ToClip()
         /// <inheritdoc/>
         public override bool Equals(object obj)
         {
-            if (obj == null && GetType() != obj.GetType())
+            if (obj == null || GetType() != obj.GetType())
```
kblok, line 69: :O

kblok (approved):

Androbin:
I can't reproduce this issue locally, seems to be a flake, an interesting one though:
https://ci.appveyor.com/project/kblok/puppeteer-sharp/builds/38258335/job/74y3bkkee23q45ri#L86

PR #1676, opened 2021-03-29
Make Clip.Scale public

campersau:
Was there a specific reason to not make it public?


lib/PuppeteerSharp/Page.cs
```diff
@@ -2033,7 +2033,7 @@ private Clip ProcessClip(Clip clip)
                 Y = y,
                 Width = Math.Round(clip.Width + clip.X - x, MidpointRounding.AwayFromZero),
                 Height = Math.Round(clip.Height + clip.Y - y, MidpointRounding.AwayFromZero),
-                Scale = 1
+                Scale = Math.Max(1, clip.Scale)
```
kblok, line 2036: I would just assign `clip.Scale`

lib/PuppeteerSharp/Media/Clip.cs
```diff
@@ -31,6 +31,6 @@ public class Clip
         /// Scale of the webpage rendering. Defaults to 1.
         /// </summary>
         /// <value>The scale.</value>
-        public int Scale { get; internal set; }
+        public int Scale { get; set; }
```
kblok, line 34: Let's default this to 1

kblok (changes requested):
Good catch!

kblok (approved):

Issue #1670, opened 2021-03-27
Object reference not set to an instance of an object


Androbin:
```
PuppeteerSharp.NavigationException: Navigation failed because browser has disconnected! (Connection failed to process Runtime.executionContextCreated. Object reference not set to an instance of an object..
    at PuppeteerSharp.DOMWorld.SetContext(ExecutionContext context) in C:\projects\puppeteer-sharp\lib\PuppeteerSharp\DOMWorld.cs:line 37
    at PuppeteerSharp.FrameManager.OnExecutionContextCreatedAsync(ContextPayload contextPayload) in C:\projects\puppeteer-sharp\lib\PuppeteerSharp\FrameManager.cs:line 291
    at PuppeteerSharp.FrameManager.Client_MessageReceived(Object sender, MessageEventArgs e) in C:\projects\puppeteer-sharp\lib\PuppeteerSharp\FrameManager.cs:line 218)
 ---> PuppeteerSharp.TargetClosedException: Navigation failed because browser has disconnected! (Connection failed to process Runtime.executionContextCreated. Object reference not set to an instance of an object..
    at PuppeteerSharp.DOMWorld.SetContext(ExecutionContext context) in C:\projects\puppeteer-sharp\lib\PuppeteerSharp\DOMWorld.cs:line 37
    at PuppeteerSharp.FrameManager.OnExecutionContextCreatedAsync(ContextPayload contextPayload) in C:\projects\puppeteer-sharp\lib\PuppeteerSharp\FrameManager.cs:line 291
    at PuppeteerSharp.FrameManager.Client_MessageReceived(Object sender, MessageEventArgs e) in C:\projects\puppeteer-sharp\lib\PuppeteerSharp\FrameManager.cs:line 218)
   at PuppeteerSharp.Helpers.TaskHelper.WithTimeout(Task task, Func`1 timeoutAction, TimeSpan timeout, CancellationToken cancellationToken) in C:\projects\puppeteer-sharp\lib\PuppeteerSharp\Helpers\TaskHelper.cs:line 82
   at PuppeteerSharp.FrameManager.NavigateFrameAsync(Frame frame, String url, NavigationOptions options) in C:\projects\puppeteer-sharp\lib\PuppeteerSharp\FrameManager.cs:line 130
   --- End of inner exception stack trace ---
```

```
[...]

PR #1677, opened 2021-03-29
Avoid leaking reference to partially constructed Frame instance

Androbin:
The upstream counterpart for reference:
https://github.com/puppeteer/puppeteer/blob/b349c91e7df76630b7411d6645e649945c4609bd/src/common/FrameManager.ts#L593

See #1670 for context

### DIFF
--- a/lib/PuppeteerSharp/Frame.cs
+++ b/lib/PuppeteerSharp/Frame.cs
@@ -61,16 +61,16 @@ namespace PuppeteerSharp
             ParentFrame = parentFrame;
             Id = frameId;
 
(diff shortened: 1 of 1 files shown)


kblok (approved):

kblok:
Will this close #1677 @Androbin ?

Androbin:
There is a chance it will take care of the issue.
It does not explain what happened in #1670.
I would like to keep that issue open for now.

Androbin:
Anyway, it can only make things better, it's good practise regardless and also closer to upstream

Issue #1680, opened 2021-03-29
Concurrently modified lists not thread-safe


Androbin:
https://github.com/hardkoded/puppeteer-sharp/blob/984fdc06a564e8ded3071a703f83d5dce1627696/lib/PuppeteerSharp/DOMWorld.cs#L19
https://github.com/hardkoded/puppeteer-sharp/blob/984fdc06a564e8ded3071a703f83d5dce1627696/lib/PuppeteerSharp/Frame.cs#L49
https://github.com/hardkoded/puppeteer-sharp/blob/984fdc06a564e8ded3071a703f83d5dce1627696/lib/PuppeteerSharp/Frame.cs#L80
https://github.com/hardkoded/puppeteer-sharp/blob/984fdc06a564e8ded3071a703f83d5dce1627696/lib/PuppeteerSharp/NetworkManager.cs#L26
https://github.com/hardkoded/puppeteer-sharp/blob/984fdc06a564e8ded3071a703f83d5dce1627696/lib/PuppeteerSharp/Helpers/MultiMap.cs#L9

Androbin:
@kblok What's the "`ConcurrentList`" implementation of your choice?

Androbin:
If order is not important, we can use `ConcurrentBag<T>`

kblok:
Do we have any reported bug around this?

kblok:
Do you think all these should be thread-safe?

Androbin:
I have looked at each and every list in the library and assessed whether they might be accessed/modified concurrently and whether that may be a problem. See https://github.com/hardkoded/puppeteer-sharp/issues/1681#issuecomment-813072742 for the kinds of problems this may cause.

Androbin:
In fact `WaitTasks` could explain #1670.

Androbin:
`_attemptedAuthentications` is modified concurrently for numerous network requests and you shouldn't be surprised to see it break. It may randomly cancel requests and I have observed errors that might have caused by it but it's difficult to confirm.

Androbin:
`MultiMap` is used in `AsyncDictionaryHelper`, which is specifically designed to be thread-safe, but it may break in unexpected ways, due to the unsynchronized lists within the otherwise synchronized dictionary.

Androbin:
`LifecycleEvents` is not very likely to cause problems, but it may prevent a page from reporting the `load` event.

Androbin:
[...]

PR #1686, opened 2021-04-05
Fix NetworkManager._attemptedAuthentications concurrency issues

Androbin:
I have prepared a test project to demonstrate the errors that can occur with the current version: [Test.zip](https://github.com/hardkoded/puppeteer-sharp/files/6256308/Test.zip)

- `ContainsKey` returns wrong value

See #1680 for context

### DIFF
--- a/lib/PuppeteerSharp/NetworkManager.cs
+++ b/lib/PuppeteerSharp/NetworkManager.cs
@@ -23,7 +23,7 @@ namespace PuppeteerSharp
         private readonly ILogger _logger;
         private Dictionary<string, string> _extraHTTPHeaders;
         private Credentials _credentials;
-        private readonly List<string> _attemptedAuthentications = new List<string>();
+        private readonly ConcurrentDictionary<string, bool> _attemptedAuthentications = new ConcurrentDictionary<string, bool>();
         private bool _userRequestInterceptionEnabled;
         private bool _protocolRequestInterceptionEnabled;
         private readonly bool _ignoreHTTPSErrors;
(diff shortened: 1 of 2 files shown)


Androbin:
CI fails due to Firefox binary server issues unrelated

Androbin:
@kblok PTAL

lib/PuppeteerSharp/NetworkManager.cs
```diff
@@ -23,7 +23,7 @@ internal class NetworkManager
         private readonly ILogger _logger;
         private Dictionary<string, string> _extraHTTPHeaders;
         private Credentials _credentials;
-        private readonly List<string> _attemptedAuthentications = new List<string>();
+        private readonly ConcurrentSet<string> _attemptedAuthentications = new ConcurrentSet<string>();
```
kblok, line 26: `ConcurrentBag` won't help?

Androbin: `Contains` for `List` and `ConcurrentBag` is O(n) while for `ConcurrentDictionary` it's O(1).
It would work too though. Would you prefer to declare it as `ConcurrentDictionary` directly?
I see no good reason to use `ConcurrentBag` over `ConcurrentDictionary`.

kblok:
Gave it a read. How about copying this code https://github.com/dotnet/roslyn/blob/6da1274c9d24c2f90a48290394a951b23617f2a3/src/Compilers/Core/Portable/InternalUtilities/ConcurrentSet.cs#L16 ?
I would even add the reference to that link in the source code.

Just to keep track and follow some `dotnet` code.

Androbin:
@kblok Done

kblok:
Great job @Androbin, this is awesome!

PR #1687, opened 2021-04-05
Fix MultiMap concurrency issues

Androbin:
I have prepared a test project to demonstrate the errors that can occur with the current version: [Test.zip](https://github.com/hardkoded/puppeteer-sharp/files/6256236/Test.zip)
- `Has` returns wrong value
- (Iterator skips values)

See #1680 for context


Androbin:
@kblok PTAL

kblok:
@Androbin trying to get firefox tests back. I'll get back here.

kblok:
Same questions here. I don't want to re-invent the wheel.

Androbin:
`ConcurrentBag` doesn't allow you to remove specific elements.

Androbin:
We could make a `ConcurrentDictionary` of locks and do the synchronization in `MultiMapUnsafe` directly.

kblok:
Why didn't we use the same approach as the other PR?

Androbin:
@kblok Done

Androbin:
Fluke? https://ci.appveyor.com/project/kblok/puppeteer-sharp/builds/38604027/job/y0x62ev4jhgqr20k/tests

PR #1693, opened 2021-04-07
Sign PuppeteerSharp.dll with private key

leonardo-fernandes:
Fixes https://github.com/hardkoded/puppeteer-sharp/issues/1691


kblok:
@leonardo-fernandes do you think this should be considered a breaking change (a.k.a v4)?

leonardo-fernandes:
From my tests, this doesn't break binary compatibility with 3.0.0. The test I executed is:
1. Built a simple console application against PuppeteerSharp 3.0.0 directly from NuGet.
2. Executed the console application.
3. Deleted the PuppeteerSharp.dll from the build folder, and replaced it with the assembly signed with the private key.
4. Executed the console application again (without rebuilding it). The output was exactly the same as before.

So this doesn't need a major version, and could be released in 3.0.1.


A few other notes for your consideration:
* I believe the public.snk file is no longer needed, since keypair.snk has both the private and public key.
* The public.snk file is included as an item in the solution. I didn't change that, but may be a good idea. It has no effect in the build, though.
* It would be interesting to have the assembly version be consistent with the package version. Right now, the PuppeteerSharp assembly has the default version 1.0.0.0.

kblok:
Do you want to clean up the `public.snk` stuff?
If you want to upgrade the Assembly at least to 3.0.0 that would be awesome :)

leonardo-fernandes:
Maybe one last thing I have noticed... the PuppeteerSharp.TestServer project includes a non-existing file `testCert.cer`. I am not sure if you forgot to check it in, or if you forgot to remove it from the solution.

Androbin:
@leonardo-fernandes See [CONTRIBUTING.md #getting-setup](https://github.com/hardkoded/puppeteer-sharp/blob/master/CONTRIBUTING.md#getting-setup)

leonardo-fernandes:
I have made another test of binary compatibility, and it looks like changing the assembly version breaks binary compatibility. So it's better to delay at least that change until 4.0.0.

How do you want to proceed? Revert the assembly version? Keep this PR open until version 4?

kblok:
Let's set the version for the assembly and the nugget to v4. That will be our next release.

PR #1697, opened 2021-04-10
Fix Frame.ChildFrames concurrency issues

Androbin:
I have prepared a test project to demonstrate the errors that can occur with the current version:  [Test.zip](https://github.com/hardkoded/puppeteer-sharp/files/6289342/Test.zip)

- `Contains` returns wrong value

See #1680

### DIFF
--- a/lib/PuppeteerSharp/Frame.cs
+++ b/lib/PuppeteerSharp/Frame.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 using Newtonsoft.Json.Linq;
 using PuppeteerSharp.Input;
@@ -39,6 +40,7 @@ namespace PuppeteerSharp
     public class Frame
     {
         private readonly CDPSession _client;
+        private readonly List<Frame> _childFrames = new List<Frame>();
 
         internal List<WaitTask> WaitTasks { get; }
 
@@ -69,7 +71,7 @@ namespace PuppeteerSharp
 
             if (parentFrame != null)
             {
-                ParentFrame.ChildFrames.Add(this);
+                ParentFrame.AddChildFrame(this);
             }
         }
 
@@ -77,7 +79,16 @@ namespace PuppeteerSharp
         /// <summary>
         /// Gets the child frames of the this frame
         /// </summary>
-        public List<Frame> ChildFrames { get; } = new List<Frame>();
+        public List<Frame> ChildFrames
+        {
+            get
+            {
+                lock (_childFrames)
+                {
+                    return _childFrames.ToList();
+                }
+            }
+        }
 
         /// <summary>
         /// Gets the frame's name attribute as specified in the tag
@@ -498,6 +509,22 @@ namespace PuppeteerSharp
         public Task TypeAsync(string selector, string text, TypeOptions options = null)
              => SecondaryWorld.TypeAsync(selector, text, options);
 
+        internal void AddChildFrame(Frame frame)
+        {
+            lock (_childFrames)
+            {
+                _childFrames.Add(frame);
+            }
+        }
+
+        internal void RemoveChildFrame(Frame frame)
+        {
+            lock (_childFrames)
+            {
+                _childFrames.Remove(frame);
+            }
+        }
+
         internal void OnLoadingStopped()
         {
             LifecycleEvents.Add("DOMContentLoaded");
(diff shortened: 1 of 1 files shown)


lib/PuppeteerSharp/Frame.cs
```diff
@@ -74,10 +76,16 @@ internal Frame(FrameManager frameManager, CDPSession client, Frame parentFrame,
...
+
         /// <summary>
         /// Gets the child frames of the this frame
         /// </summary>
-        public List<Frame> ChildFrames { get; } = new List<Frame>();
+        public List<Frame> ChildFrames {
+            get {
+                return _childFrames.OrderBy(frame => frame.Id).ToList();
+            }
+        }
```
kblok, line 88: Do we need this?

lib/PuppeteerSharp/Frame.cs
```diff
@@ -74,10 +76,16 @@ internal Frame(FrameManager frameManager, CDPSession client, Frame parentFrame,
...
         #region Properties
+        internal ICollection<Frame> _childFrames { get; } = new ConcurrentSet<Frame>();
+
         /// <summary>
         /// Gets the child frames of the this frame
         /// </summary>
-        public List<Frame> ChildFrames { get; } = new List<Frame>();
+        public List<Frame> ChildFrames {
+            get {
+                return _childFrames.OrderBy(frame => frame.Id).ToList();
```
kblok, line 86: I think this won't work because. `ParentFrame.ChildFrames.Add(this);` will add child to a copy.

lib/PuppeteerSharp/Frame.cs
Androbin: This is an attempt at backwards-compatibility and deterministic order. It is not strictly needed.

lib/PuppeteerSharp/Frame.cs
Androbin: Touche, I forgot to change that line

lib/PuppeteerSharp/Frame.cs
Androbin: For example, I think that `FrameManager` expects the order to be deterministic.

Androbin: Specifically `RemoveFramesRecursively` and possibly user code as well.

kblok:
We are having some fails in CI

Androbin:
Turns out the tests expect a deterministic order, so my attempt made sense, it just wasn't enough to preserve the original behavior.

Androbin:
We could use `Interlocked.Increment` to generate indices. That would preserve the original behavior in the absence of race conditions.

kblok:
Is your test still failing after the other changes we merged?
I'm wondering if this needs to be thread-safe.

Androbin:
If two or more calls to either `Page.frameAttached` or `Page.frameDetached` overlap, then `ChildFrames` can drop entries. The other changes don't alleviate that.

Androbin:
We could just use a simple and straight-forward `lock` to synchronize the read and write access.

lib/PuppeteerSharp/Frame.cs
```diff
@@ -69,15 +70,25 @@ internal Frame(FrameManager frameManager, CDPSession client, Frame parentFrame,
...
-                ParentFrame.ChildFrames.Add(this);
+                lock (ParentFrame._childFrames) {
+                    ParentFrame._childFrames.Add(this);
+                }
             }
         }
 
         #region Properties
+        internal List<Frame> _childFrames { get; } = new List<Frame>();
+
```
kblok, line 81: We could leave only one public property and add an internal `AddChildFrame` method. There we can do the lock.

Androbin: We need the lock for both reading and writing though.

kblok: Sure. But I don't like that internal prop

Androbin: We need a backing collection declared somewhere and `ChildFrames` is a public api, so we need both.

Androbin: Even changing the declaration of `ChildFrames` from `List<Frame>` to `IList<Frame>` would be a breaking change.

Androbin: We could factor this into out into a `ConcurrentList` but I remember you saying you would want to avoid that.

kblok: We could have that but as internal variable instead a property. And let the child frames interact with that `_childFrames` variable through an internal method.

kblok:
Great job!

PR #1698, opened 2021-04-10
Fix WaitTasks concurrency issues

Androbin:
I have prepared a test project to demonstrate the errors that can occur with the current version:  [Test.zip](https://github.com/hardkoded/puppeteer-sharp/files/6289374/Test.zip)

- `Index was outside the bounds of the array.`
- `Contains` returns wrong value

See #1680


kblok:
If you want to create a PR with your test project under "demos" that would be great.
Concurrency Demo?

PR #1784, opened 2021-07-15
Add iphone family devices

carmonac:
- Fix iPhone X user agents
- Add iPhone devices from XR version to 11 Pro Max

Data about these devices have been extracted from Puppeteer JavaScript project.


kblok:
Thank you!

kblok (approved):

PR #1798, opened 2021-09-07
Add missing set request interception tests

kblok:
closes #1735

PR #1800, opened 2021-09-08
Fix NetworkManager.OnRequestAsync

Androbin:
Fixes the issues seen in #1798 @kblok


kblok:
Thank you @Androbin !

Androbin:
The condition was newly introduced in #720 and probably for a reason, so we should see if we can find a replacement.

kblok:
@Androbin I think we should be good. Let's see what tests say.

PR #1803, opened 2021-09-14
Simplify NuGet dependencies

pirrmann:
Referencing PuppeteerSharp pulls lots of packages transitively, and it looks like these come mostly from the System.Net.Http package, which apparently is not needed in order to build PuppeteerSharp, but only the Tests project.
Removing this dependency from the published NuGet package would make things simpler, so I've moved the reference to the test project.


pirrmann:
Failing tests here don't seem related to my changes. Let me know if there is anything I should change.

kblok (approved):

kblok:
Yeah, it looks good. Rebuilding the CI

Issue #1440, opened 2020-02-29
Enable CA1724 rule


kblok:
Change the action for CA1724 on the [ruleset file](https://github.com/kblok/puppeteer-sharp/blob/master/lib/PuppeteerSharp.ruleset) to Error, and make the solution build

kgar:
Looks like this issue can be closed.

PR #1807, opened 2021-09-23
Activated CA1724 and resolved errors

kgar:
Set the CA1724 rule to "Error" and resolved build errors related to the rule.

Renamed Tracing, TracingOptions, TracingTests, and Extensions to satisfy the rule.

For issue #1440

### DIFF
--- a/lib/PuppeteerSharp.Tests/DevToolsTracingTests/DevToolsTracingTests.cs
+++ b/lib/PuppeteerSharp.Tests/DevToolsTracingTests/DevToolsTracingTests.cs
@@ -53,7 +53,7 @@ namespace PuppeteerSharp.Tests.TracingTests
             }
         }
 
-        [PuppeteerTest("tracing.spec.ts", "Tracing", "should output a trace")]
+        [PuppeteerTest("devtoolstracing.spec.ts", "Tracing", "should output a trace")]
         [SkipBrowserFact(skipFirefox: true)]
         public async Task ShouldOutputATrace()
         {
@@ -68,7 +68,7 @@ namespace PuppeteerSharp.Tests.TracingTests
             Assert.True(File.Exists(_file));
         }
 
-        [PuppeteerTest("tracing.spec.ts", "Tracing", "should run with custom categories if provided")]
+        [PuppeteerTest("devtoolstracing.spec.ts", "Tracing", "should run with custom categories if provided")]
         [SkipBrowserFact(skipFirefox: true)]
         public async Task ShouldRunWithCustomCategoriesProvided()
         {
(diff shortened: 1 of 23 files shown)


lib/PuppeteerSharp.Tests/DevToolsTracingTests/DevToolsTracingTests.cs
```diff
@@ -53,11 +53,11 @@ public override async Task DisposeAsync()
             }
         }
 
-        [PuppeteerTest("tracing.spec.ts", "Tracing", "should output a trace")]
+        [PuppeteerTest("devtoolstracing.spec.ts", "Tracing", "should output a trace")]
```
kblok, line 56: This `spec.ts` should match the location of the test in puppeteer. I think we shouldn't change this.

lib/PuppeteerSharp.Tests/DevToolsTracingTests/DevToolsTracingTests.cs
```diff
@@ -14,11 +14,11 @@
 namespace PuppeteerSharp.Tests.TracingTests
 {
     [Collection(TestConstants.TestFixtureCollectionName)]
-    public class TracingTests : PuppeteerPageBaseTest
+    public class DevToolsTracingTests : PuppeteerPageBaseTest
```
kblok, line 17: Why the rule didn't like this name?

lib/PuppeteerSharp/DevToolsTracing.cs
```diff
@@ -23,7 +23,7 @@ namespace PuppeteerSharp
     /// await Page.Tracing.StopAsync();
     /// </code>
     /// </example>
-    public class Tracing
+    public class DevToolsTracing
```
kblok, line 26: This is a breaking change. Do we need to do this?

kblok (changes requested):

lib/PuppeteerSharp.Tests/DevToolsTracingTests/DevToolsTracingTests.cs
kgar: The analyzer was ok with the original name on this file. I had originally renamed it to match the rename of the class under test. I'll get this changed back.

lib/PuppeteerSharp/DevToolsTracing.cs
kgar: The CA1724 rule was emitting an error for the class name Tracing. However, it looks like keeping parity with puppeteer class names is important, so I think the best approach may be to suppress the error for this class, rather than renaming it. I should have those changes for review soon.

lib/PuppeteerSharp.Tests/TracingTests/TracingTests.cs
```diff
@@ -14,7 +14,8 @@
 namespace PuppeteerSharp.Tests.TracingTests
 {
     [Collection(TestConstants.TestFixtureCollectionName)]
-    public class TracingTests : PuppeteerPageBaseTest
+    public class TracingTests
+        : PuppeteerPageBaseTest
```
kblok, line 18: Do we need this change?

lib/PuppeteerSharp/Tracing.cs
```diff
@@ -9,6 +9,7 @@
 
 namespace PuppeteerSharp
 {
+#pragma warning disable CA1724
```
kblok, line 12: Don't we need a `pragma restore` after the code we want to bypass?

kblok (approved):

PR #1822, opened 2021-09-29
Set CA2008 to "Error" and resolved compiler errors

### LINKED ISSUE #1431: Enable CA2008 rule
Change the action for CA2008 on the [ruleset file](https://github.com/kblok/puppeteer-sharp/blob/master/lib/PuppeteerSharp.ruleset) to Error, and make the solution build

kgar:
closes #1431 

Set CA2008 to "Error" and resolved compiler errors.

When prompted for a TaskScheduler, set the value to the Default TaskScheduler.


kblok:
Thank you!

PR #1823, opened 2021-09-29
Set CA2237 to "Error" and resolved compiler errors.

kgar:
RE: #1429

Marked ISerializable types with SerializableAttribute and added serialization constructor delegating to args the base constructor.

### DIFF
--- a/lib/PuppeteerSharp/ProcessException.cs
+++ b/lib/PuppeteerSharp/ProcessException.cs
@@ -6,7 +6,7 @@ namespace PuppeteerSharp
     /// <summary>
     /// process exception thrown by <see cref="Launcher"/>.
     /// </summary>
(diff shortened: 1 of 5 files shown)


lib/PuppeteerSharp/ProcessException.cs
```diff
@@ -1,11 +1,13 @@
 using System;
+using System.Runtime.Serialization;
 
 namespace PuppeteerSharp
 {
     /// <summary>
     /// process exception thrown by <see cref="Launcher"/>.
     /// </summary>
-    #pragma warning disable 612, 618
+#pragma warning disable 612, 618
```
kblok, line 9:
```suggestion
    #pragma warning disable 612, 618
```

kblok:
Thank you!

PR #1826, opened 2021-09-29
Set CA1001 to "Error" and resolved issues

### LINKED ISSUE #1423: Enable CA1001 rule
Change the action for CA1001 on the [ruleset file](https://github.com/kblok/puppeteer-sharp/blob/master/lib/PuppeteerSharp.ruleset) to Error, and make the solution build

kgar:
closes #1423 

For classes with disposable members, made those classes implement IDisposable, using the recommended Dispose() methods.

This created a ripple effect where the newly disposable classes needed to be disposed of wherever used.

Updated documentation, examples, sample code, and demos where `BrowserFetcher` is used and could be disposed.

@kblok sorry, this one has got a lot of changes. Documentation aside, the rest was necessary to satisfy the analyzer.

### DIFF
--- a/demos/PuppeteerSharpPdfDemo/PuppeteerSharpPdfDemo-Local.csproj
+++ b/demos/PuppeteerSharpPdfDemo/PuppeteerSharpPdfDemo-Local.csproj
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFrameworks>netcoreapp3.1;net471</TargetFrameworks>
-    <LangVersion>7.3</LangVersion>
+    <LangVersion>8.0</LangVersion>
   </PropertyGroup>
   <PropertyGroup Condition="'$(TargetFramework)'=='net471'">
     <RuntimeIdentifier>win7-x64</RuntimeIdentifier>
--- a/lib/PuppeteerSharp/BrowserFetcher.cs
+++ b/lib/PuppeteerSharp/BrowserFetcher.cs
@@ -30,7 +30,7 @@ namespace PuppeteerSharp
     /// var browser = await await Puppeteer.LaunchAsync(new LaunchOptions { ExecutablePath = revisionInfo.ExecutablePath});
     /// </code>
     /// </example>
-    public class BrowserFetcher
+    public class BrowserFetcher : IDisposable
     {
         private static readonly Dictionary<Product, string> _hosts = new Dictionary<Product, string>
         {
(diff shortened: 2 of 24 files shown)


demos/PuppeteerSharpPdfDemo/PuppeteerSharpPdfDemo-Local.csproj
```diff
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFrameworks>netcoreapp3.1;net471</TargetFrameworks>
-    <LangVersion>7.3</LangVersion>
+    <LangVersion>8.0</LangVersion>
```
kblok, line 6: <3

lib/PuppeteerSharp/BrowserFetcher.cs
```diff
@@ -30,7 +30,7 @@ namespace PuppeteerSharp
     /// var browser = await await Puppeteer.LaunchAsync(new LaunchOptions { ExecutablePath = revisionInfo.ExecutablePath});
     /// </code>
     /// </example>
-    public class BrowserFetcher
+    public class BrowserFetcher : IDisposable
```
kblok, line 33: This will mean a breaking change and a major version change.

kblok (approved):
Great job!
Quite painful, but needed.

lib/PuppeteerSharp/BrowserFetcher.cs
kgar: I was getting nervous, the more files that changed 😱

demos/PuppeteerSharpPdfDemo/PuppeteerSharpPdfDemo-Local.csproj
kgar: Those `using` statements are the spice of life

kblok:
It seems the build broke badly https://ci.appveyor.com/project/kblok/puppeteer-sharp/builds/40962024

kgar:
I see the problem. I'm checking a few things over on MSDN, but I should have the tests passing again.

kgar:
The issue was I disposed of the timeout task before cancellation had put the task into a finalized state. I think we're ok not to dispose of the timeout task, as the cancellation token source will ensure that the task gets sorted out.

kblok:
It seems netcore builds are getting stuck on `ShouldSupportCustomTransport`

kgar:
I'm looking forward to figuring this one out 👀

kblok (approved):

PR #1829, opened 2021-09-30
Tests - Add PuppeteerFact to allow for greater control over tests

amaitland:
When debugging the `xUnit` tests the default timeouts are too low. This `PR` changes the timeouts only when the debugger is attached.

- Add a new `PuppateerFact` as common for all `Facts`
- Current tests with `[Fact]` were replaced with `[PuppeteerFact(Timeout = -1)]` to ensure all tests were 
- Add new TestConstants.DebuggerAttachedTestTimeout constant with 5min timeout
- Set Test timeout to TestConstants.DebuggerAttachedTestTimeout when debugger attached
- Set Page.DefaultTimeout to TestConstants.DebuggerAttachedTestTimeout when debugger attached

No changes were made to the `[Theory]` tests, though that would be simply enough to add another attribute if required.

Unfortunately replacing the attributes in so many classes creates a little bit of noise, so I appreciate it looks like a lot of files to review. The other simpler option would be to change the timeout with an `#if DEBUG` to change the timeouts for debug builds generally. Overall I think this is preferable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ await using var page = await browser.NewPageAsync();
 await page.GoToAsync("http://www.google.com");
 await page.PdfAsync(outputFile);
 ```
-<sup><a href='/lib/PuppeteerSharp.Tests/PageTests/PdfTests.cs#L28-L37' title='Snippet source file'>snippet source</a> | <a href='#snippet-pdfasync' title='Start of snippet'>anchor</a></sup>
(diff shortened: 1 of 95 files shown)


kblok (approved):
I like that :)

amaitland:
Thoughts on a `PuppeteerTheory` attribute so all the tests have a consistent timeout? Let me know and I'll great a `PR` if required.

PR #1831, opened 2021-09-30
Allow ResponseData.Headers to support multiple values for the same key

amaitland:
- Without breaking the current API this change allows for arrays of values
  (Anything that implements ICollection and can be iterated over).
- The Headers with null values will be ignored, previously an exception
  would have occurred with the Value.ToString() call so this is unlikely
  a breaking change in behaviour.
- New test case added

Resolves https://github.com/hardkoded/puppeteer-sharp/issues/1542
Replaces #1607

### DIFF
--- a/lib/PuppeteerSharp.Tests/RequestInterceptionTests/RequestRespondTests.cs
+++ b/lib/PuppeteerSharp.Tests/RequestInterceptionTests/RequestRespondTests.cs
@@ -145,5 +145,35 @@ namespace PuppeteerSharp.Tests.RequestInterceptionTests
             Assert.Equal("True", response.Headers["foo"]);
             Assert.Equal("Yo, page!", await Page.EvaluateExpressionAsync<string>("document.body.textContent"));
         }
+
+        [SkipBrowserFact(skipFirefox: true)]
+        public async Task ShouldAllowMultipleInterceptedRequestResponseHeaders()
+        {
+            await Page.SetRequestInterceptionAsync(true);
+            Page.Request += async (_, e) =>
+            {
+                await e.Request.RespondAsync(new ResponseData
+                {
+                    Status = HttpStatusCode.OK,
+                    Headers = new Dictionary<string, object>
+                    {
+                        ["foo"] = new bool[] { true, false },
+                        ["Set-Cookie"] = new string[] { "sessionId=abcdef", "specialId=123456" }
+                    },
+                    Body = "Yo, page!"
+                });
+            };
+
+            var response = await Page.GoToAsync(TestConstants.EmptyPage);
+            var cookies = await Page.GetCookiesAsync(TestConstants.EmptyPage);
+
+            Assert.Equal(HttpStatusCode.OK, response.Status);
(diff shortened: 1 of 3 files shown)


amaitland:
Failing Linux test case looks unrelated. Please let me know if you think it's related and I'll look into it further.

lib/PuppeteerSharp.Tests/RequestInterceptionTests/RequestRespondTests.cs
```diff
@@ -145,5 +145,36 @@ public async Task ShouldStringifyInterceptedRequestResponseHeaders()
             Assert.Equal("True", response.Headers["foo"]);
             Assert.Equal("Yo, page!", await Page.EvaluateExpressionAsync<string>("document.body.textContent"));
         }
+
+        [PuppeteerTest("requestinterception.spec.ts", "Request.respond", "should allow for multiple headers to be sent for an intercepted request response ")]
```
kblok, line 149: I don't see this test [here](https://github.com/puppeteer/puppeteer/blob/main/test/requestinterception.spec.ts). Am I missing something?

amaitland: Thanks for the review,  I was meaning to follow up on this. 

I ran out of time to dig into in the purpose of PuppeteerTestAttribute, so i just did a copy paste with the intention of following up. Should I simply remove the attribute? 

Perhaps PuppeteerTestAttribute xml doc summary could be expanded a little?

kblok: We use `PuppeteerTest` to report how many missing tests we have compared with the upstream puppeteer.

amaitland: Thanks for the clarification 👍  Attribute has been removed.

lib/PuppeteerSharp/Request.cs
```diff
@@ -231,24 +232,41 @@ public async Task RespondAsync(ResponseData response)
...
-            var responseHeaders = new Dictionary<string, string>();
+            var responseHeaders = new List<Header>();
 
             if (response.Headers != null)
             {
-                foreach (var keyValue in response.Headers)
+                foreach (var keyValuePair in response.Headers)
                 {
-                    responseHeaders[keyValue.Key] = keyValue.Value.ToString();
+                    // Skip headers with null values
```
kblok, line 241:
```suggestion
```

lib/PuppeteerSharp/Request.cs
```diff
@@ -231,24 +232,41 @@ public async Task RespondAsync(ResponseData response)
...
+                    if (keyValuePair.Value == null)
+                    {
+                        continue;
+                    }
+
+                    // Use ICollection instead of IEnumerable as string implements IEnumerable
```
kblok, line 247:
```suggestion
```

PR #1849, opened 2021-10-15
Add an option to include/specify a custom delegate to download the browser.

mckabue:
To Support using a cache browser location.

Scenario: Azure function runs for a limited amount of time (10 mins  - Consumption plan), and one has to download the browser each time the app starts!

> Azure Functions/Apps with *Headless Chromium*
>  -  [Headless Chromium ONLY works in the Linux Consumption plan](https://anthonychu.ca/post/azure-functions-headless-chromium-puppeteer-playwright/)
>  -  [Linux consumption apps ONLY Run-From-Package.](https://stackoverflow.com/a/56035713/3563013)
>  -  [You can ONLY use "shared mounted storage" or `Path.GetTempPath()`.](https://anthonychu.ca/post/azure-functions-puppeteer-pdf-razor-template/)

https://github.com/hardkoded/puppeteer-sharp/discussions/1845

### DIFF
--- a/lib/PuppeteerSharp/BrowserFetcher.cs
+++ b/lib/PuppeteerSharp/BrowserFetcher.cs
@@ -52,6 +52,7 @@ namespace PuppeteerSharp
         };
 
         private readonly WebClient _webClient = new WebClient();
+        private readonly CustomFileDownloadAction _customFileDownload;
         private bool _isDisposed;
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


gregoriomartin:
hi @McKabue, do you have any example on how to use this? I need to avoid downloading the browser on Function startup.

McKabue:
Download the browser and put it in a folder inside your solution, eg: puppeteer.

Then copy the browser in the build directory: eg:
`<Content Include="puppeteer/**" CopyToPublishDirectory="Always" CopyToOutputDirectory="Always" />`

When creating a browser fetcher options, do:
```
DirectoryInfo directoryInfo = new(Path.Combine(Directory.GetCurrentDirectory(), "puppeteer"));
BrowserFetcherOptions browserFetcherOptions = new BrowserFetcherOptions
{
    ....other options removed for brevity...

    CustomFileDownload = (string address, string fileName) =>
    {
        FileInfo fileInfo = new(Path.Combine(directoryInfo.FullName, new UriBuilder(address).Uri.Segments.Last()));

        if (fileInfo.Exists)
        {
            fileInfo.CopyTo(fileName);
        }

        return Task.CompletedTask;
    }
}
```

PR #1868, opened 2021-11-10
Make Page.EmulateCPUThrottlingAsync public

amaitland:
Currently it's internal and only accessible via the Test project


kblok:
Thank you!

kblok (approved):
🚀

PR #1871, opened 2021-11-16
Fix for #1866

### LINKED ISSUE #1866: TargetCreated CDP event received before subscription can be made
### Description

When connecting to a running browser instance using `Puppeteer.ConnectAsync` the library does multiple things in one step and returns the Browser object. Once it's returned I can subscribe to the TargetCreated event, however before this is done the library already receives `Target.targetCreated` events from the browser (because there is 1 opened tab at least) which won't be handled by my event handler.

```csharp
_browser = await Puppeteer.ConnectAsync(new ConnectOptions { BrowserWSEndpoint = WebSocketEndpoint }, GetLoggerFactory());

// At this point a Target.targetCreated event received because ConnectAsync calls Target.setDiscoverTargets, so the following event handler will not execute
_browser.TargetCreated += OnTargetCreatedOrChanged;
```

### Expected behavior:

I would like to be able to use the event handler for all received events.
Maybe make the setDiscoverTargets call optional? Or provide a way to set event handlers before this call is made?

### Versions

PuppeteerSharp 5.0.0
.NET Core 3.1

brnbs:
Added InitAction to fix #1866 
Fixed serialization issues for logging
Added example to console logging


docfx_project/examples/LogCDPCommunication.md
```diff
@@ -0,0 +1,66 @@
+# How to log messages between Puppeteer-Sharp and Chromium
+
+_Contributors: [Darío Kondratiuk](https://www.hardkoded.com/)_
```
kblok, line 3: You can add yourself to the contributors list :)

docfx_project/examples/LogCDPCommunication.md
```diff
@@ -0,0 +1,66 @@
+# How to log messages between Puppeteer-Sharp and Chromium
```
kblok, line 1: We need to update the toc https://github.com/hardkoded/puppeteer-sharp/blob/master/docfx_project/examples/toc.yml#L20

lib/PuppeteerSharp/Browser.cs
```diff
@@ -520,10 +520,17 @@ private async Task CreateTargetAsync(TargetCreatedResponse e)
...
+            LauncherBase launcher,
+            Func<TargetInfo, bool> targetFilter,
+            Action<Browser> initAction = null)
         {
             var browser = new Browser(connection, contextIds, ignoreHTTPSErrors, defaultViewPort, launcher, targetFilter);
+
+            if (initAction != null)
+            {
+                initAction(browser);
+            }
```
kblok, line 532:
```suggestion
            initAction?.Invoke(browser);
```

kblok (changes requested):
Thank you! Great job!

lib/PuppeteerSharp.Tests/LauncherTests/PuppeteerConnectTests.cs
```diff
@@ -120,6 +120,26 @@ await using (var originalBrowser = await Puppeteer.LaunchAsync(TestConstants.Def
             }
         }
 
+        [PuppeteerTest("launcher.spec.ts", "Puppeteer.connect", "should be able to set browser properties using connect options")]
```
kblok, line 123: This is used to map with pptr tests. we don't need it 
```suggestion
        [PuppeteerTest("launcher.spec.ts", "Puppeteer.connect", "should be able to set browser properties using connect options")]
```

kblok (approved):
Thank you!

PR #1886, opened 2021-12-20
Fix some UnobservedTaskException errors

### LINKED ISSUE #1870: UnobservedTaskException: System.InvalidOperationException: Cannot enterFrom in state Disposed
### Description

The following Exception is catched by the TPL UnobservedTaskException-handler some time after the browser has been disposed. The actual exception is triggered by the finalizer call of ~LauncherBase().

**System.InvalidOperationException: Cannot enterFrom in state Disposed**

The call sequence is: 

- LauncherBase ~LauncherBase() -> Dispose(false) => _stateManager.CurrentState.Dispose(this);
- CurrentState is Exited (which does not overwrite Dispose) => State.Dipose(LauncherBase) is called
- State.Dipose(LauncherBase) calls => StateManager.Disposed.EnterFromAsync(p, this, TimeSpan.Zero)

- EnterFromAsync with 3 Parameters is not overwritten in Dispose State, so implementation in State is called:
EnterFromAsync(LauncherBase p, State fromState, TimeSpan timeout) => Task.FromException(InvalidOperation("enterFrom"));
=> Exception is thrown.

```csharp
// Arrange
      TaskScheduler.UnobservedTaskException += (object sender, UnobservedTaskExceptionEventArgs eventArgs) =>
      {
          ((AggregateException)eventArgs.Exception).Handle(ex =>
          {
              Console.WriteLine("Exception type: {0}", ex.GetType());
              return true;
          });
      };

      var browserContext = await browser.CreateIncognitoBrowserContextAsync();
      var page = await browserContext.NewPageAsync();
                
      await page.GoToAsync("http://www.google.com", WaitUntilNavigation.Networkidle0);                
      await page.ScreenshotAsync("C:\\Result.png");
                
      await page.CloseAsync();
      await page.DisposeAsync();

      await browserContext.CloseAsync();

// Act
      await browser.DisposeAsync();
      Thread.Sleep(5000); //helps to get the error more reliable

// Assert
=> No UnobservedTaskException expected.
```

### Expected behavior:

No UnobservedTaskException should be detected.

### Actual behavior:

UnobservedTaskException is detected.

### Versions

* Which version of PuppeteerSharp are you using?
5.1.0

* Which .NET runtime and version are you targeting? E.g. .NET framework 4.6.1 or .NET Core 2.0.
.NET framework 4.8

### Additional Information
It seems the bug sneaked in with commit "Experimental Firefox Support (#1640)

A possible fix could be to implement this method in DisposedState:
```csharp
        public override Task EnterFromAsync(LauncherBase p, State fromState, TimeSpan timeSpan)
        {
            if (fromState == StateManager.Exited)
            {
                return null;
            }

            Kill(p);

            p.ExitCompletionSource.TrySetException(new ObjectDisposedException(p.ToString()));
            p.TempUserDataDir?.Dispose();

            return null;
        }
```

Chris-Dev-GH:
closes #1870
Removed method which was not called anymore. Added override method which handles the StateMaanger.Exited use case explicitly.

### DIFF
--- a/lib/PuppeteerSharp/States/DisposedState.cs
+++ b/lib/PuppeteerSharp/States/DisposedState.cs
@@ -9,21 +9,19 @@ namespace PuppeteerSharp.States
         {
         }
 
-        public void EnterFrom(LauncherBase p, State fromState)
+        public override Task EnterFromAsync(LauncherBase p, State fromState, TimeSpan timeSpan)
         {
-            if (!StateManager.TryEnter(p, fromState, this))
+            if (fromState == StateManager.Exited)
             {
-                // Delegate Dispose to current state, because it has already changed since
-                // transition to this state was initiated.
-                StateManager.CurrentState.Dispose(p);
+                return null;
             }
-            else if (fromState != StateManager.Exited)
-            {
-                Kill(p);
 
-                p.ExitCompletionSource.TrySetException(new ObjectDisposedException(p.ToString()));
-                p.TempUserDataDir?.Dispose();
-            }
+            Kill(p);
+
+            p.ExitCompletionSource.TrySetException(new ObjectDisposedException(p.ToString()));
+            p.TempUserDataDir?.Dispose();
+
+            return null;
         }
 
         public override Task StartAsync(LauncherBase p) => throw new ObjectDisposedException(p.ToString());


kblok:
Thank you!

PR #1893, opened 2022-01-14
Migrate from AppVeyor to Github Actions #1836

### LINKED ISSUE #1836: Migrate from AppVeyor to Github actions
It would be great if we could get out of appveyor.

venky8951:
closes #1836


kblok:
hmmm I can't see how I can approve these new actions

.github/workflows/demo.yml
```diff
@@ -0,0 +1,41 @@
+name: Demo Project
+
+on:
+  push:
+  pull_request:
+    branches: [ main ]
```
kblok, line 6: We don't have a main branch yet :)

venky8951: Hi, When we set branches to main it takes the default branch which is master, we can see Actions run here -> https://github.com/venky8951/puppeteer-sharp/actions

kblok:
Thank you @venky8951!
I've made some edits. We do need to point to `master`.
I will need some follow-up PR but this is a great first step!

PR #1899, opened 2022-01-20
Switch browser contexts from Dictionary to ConcurrentDictionary

thomaschanneladvisor:
Switch browser contexts from Dictionary to ConcurrentDictionary due to rare race condition throwing non-concurrent error


kblok (approved):
🚀

thomaschanneladvisor:
Is that a flaky test or something i should look at ?

kblok:
rebuilt :)

PR #1910, opened 2022-02-14
Up docs

lofcz:
This gave me a real headache. On some html documents / pages pdf would be generated prematurely with no text.
Relates to https://github.com/puppeteer/puppeteer/issues/422

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,8 +89,8 @@ using var browserFetcher = new BrowserFetcher();
 await browserFetcher.DownloadAsync();
 await using var browser = await Puppeteer.LaunchAsync(new LaunchOptions {Headless = true});
 await using var page = await browser.NewPageAsync();
-await page.GoToAsync("http://www.google.com");
-await page.EvaluateExpressionHandleAsync("document.fonts.ready");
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -90,6 +90,7 @@ await browserFetcher.DownloadAsync();
 await using var browser = await Puppeteer.LaunchAsync(new LaunchOptions {Headless = true});
 await using var page = await browser.NewPageAsync();
 await page.GoToAsync("http://www.google.com");
+await page.EvaluateExpressionHandleAsync("document.fonts.ready");
```
kblok, line 93: Can we add a comment on top of this line?
```

```suggestion
// You might also need to wait for the fonts to be fully loaded.
await page.EvaluateExpressionHandleAsync("document.fonts.ready");
```

lofcz: Updated. I think the major point here is to outline the possibility of fonts not being loaded and offer some immediate fixes. Might not work in all cases but at least gives a starting point for research.
_(I was thinking for an hour this is due to chromium processess not being disposed of or something due to the bug hitting only on 2nd+ calls)._

PR #1915, opened 2022-02-23
bumped the default chromium revision to 974265

litan1106:


kblok:
@litan1106 it seems all accessibility tests started to fail with this change. Can you check how that was solved on puppeteer?

kblok:
Closing in favor of https://github.com/hardkoded/puppeteer-sharp/pull/1923

PR #1920, opened 2022-03-15
Add Page.WaitForNetworkIdleAsync

### LINKED ISSUE #1911: Add support for page.waitForNetworkIdle
It seems that Puppeteer has something called [page.waitForNetworkIdle](https://github.com/puppeteer/puppeteer/blob/main/docs/api.md#pagewaitfornetworkidleoptions). I can't find that in PuppeteerSharp. Could support for that be added?

I'm guessing a workaround is to use `NavigationOptions(WaitUntil = [|WaitUntilNavigation.Networkidle0|])` when navigating or loading content.

(Using F# syntax above, in case it looks weird to anyone.)

amaitland:
Adds new `Page.WaitForNetworkIdleAsync` method ported from [page.waitForNetworkIdle](https://github.com/puppeteer/puppeteer/blob/8ff9d598bf4afd10cbc61ca9579b7bd38edb8026/src/common/Page.ts#L1942)

The current method signature looks like the following:

```c#
public async Task WaitForNetworkIdleAsync(int? timeout = null, int idleTime = 500)
```

The other `WaitFor` methods use `WaitForOptions` which only has a `Timeout` property. Happy to add a new `WaitForNetworkIdleOptions` class if required, though seems overkill for two optional params.

I've expanded the `ShouldWorkWithNoTimeout` test a little, I would have liked to see the responses validated to confirm that not only did the requests complete, that the expected number completed. So I've done that in `ShouldWorkWithNoTimeout`

Tests passing locally, be great to have a second pair of eyes on this.

Resolves #1911

### DIFF
--- a/lib/PuppeteerSharp/NetworkManager.cs
+++ b/lib/PuppeteerSharp/NetworkManager.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Globalization;
+using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using PuppeteerSharp.Helpers;
@@ -223,6 +224,11 @@ namespace PuppeteerSharp
                     Request = request
                 });
             }
+        }
+
+        internal int NumRequestsInProgress()
+        {
+            return _requestIdToRequest.Count(x => x.Value.Response == null);
         }
 
         private void ForgetRequest(Request request, bool events)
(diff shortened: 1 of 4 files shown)


lib/PuppeteerSharp/NetworkManager.cs
```diff
@@ -225,6 +226,11 @@ private void OnLoadingFinished(LoadingFinishedResponse e)
             }
         }
 
+        internal int NumRequestsInProgress()
+        {
+            return _requestIdToRequest.Count(x => x.Value.Response == null);
+        }
```
kblok, line 232: Let's turn this into a read only property.

kblok (changes requested):
Thank you!
Just a few comments

kblok (approved):
🚀

Issue #1388, opened 2020-02-29
Enable SA1201 rule


kblok:
Change the action for SA1201 on the [ruleset file](https://github.com/kblok/puppeteer-sharp/blob/master/lib/PuppeteerSharp.ruleset) to Error, and make the solution build

BL323:
Are you accepting PRs to resolve these analyzer issues? I'd like to attempt this one if possible?

kblok:
@BL323 go for it!

BL323:
I’ve raised a PR to enable the analyzer, some reordering of members was required but nothing too invasive. Let me know if there is anything I’ve missed, thanks.

akshay-zz:
@kblok Can we close these issues and similar issues? As the code has been merged, it will become easier to look into `good first issue`.

kblok:
@akshay-zz yes!

Issue #1395, opened 2020-02-29
Enable SA1124 rule


kblok:
Change the action for SA1124 on the [ruleset file](https://github.com/kblok/puppeteer-sharp/blob/master/lib/PuppeteerSharp.ruleset) to Error, and make the solution build

PR #1937, opened 2022-04-18
Configure SA1201 action as error

BL323:
The intention of this change is to enable the analyzer rule [SA1201: ElementsMustAppearInTheCorrectOrder](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1201.md) as described in the issue #1388.

Please let me know if there is anything I've missed while raising the PR.

### DIFF
--- a/lib/PuppeteerSharp/Browser.cs
+++ b/lib/PuppeteerSharp/Browser.cs
@@ -42,6 +42,15 @@ namespace PuppeteerSharp
         /// </summary>
         private const int CloseTimeout = 5000;
 
+        #region Private members
+
+        private readonly ConcurrentDictionary<string, BrowserContext> _contexts;
+        private readonly ILogger<Browser> _logger;
+        private readonly Func<TargetInfo, bool> _targetFilterCallback;
+        private Task _closeTask;
+
+        #endregion
+
         internal Browser(
             Connection connection,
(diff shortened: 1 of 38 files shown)


lib/PuppeteerSharp/Browser.cs
```diff
@@ -42,6 +42,15 @@ public class Browser : IDisposable, IAsyncDisposable
         /// </summary>
         private const int CloseTimeout = 5000;
 
+        #region Private members
+
```
kblok, line 46: Can you remove the regions?
I'm taking the chance to them every time I see one :)

kblok (approved):
Thank you!

lib/PuppeteerSharp/Browser.cs
BL323: All done :) I've now removed them in few files that were in the vicinity of my previous changes.

I could look at picking up the issue #1395 to enable SA1124 which would remove all the other regions. Although I didn't want to make the scope of this PR too big.

BL323:
The failing test `NetworkRedirectsShouldReportSecurityDetails` seems to be a little flaky, Sometimes this passes, sometimes fails when running from my local machine. 

I don't think any of the from this PR would have effected it's behavior?

PR #1941, opened 2022-04-20
Configure SA1124 action as error, remove regions

BL323:
The intention of this change is to enable the analyzer rule [SA1124: DoNotUseRegions](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1124.md) as described in the issue #1395.

Please let me know if there is anything I've missed while raising the PR.

There are some additional regions in the tests, should these be removed as well?


kblok (approved):
Looks good!

PR #1948, opened 2022-04-28
Frame - Fix xml doc example code

amaitland:
- Add CDATA formatting to all blocks
- Replace const with var
- Fix some other mistakes


kblok (approved):
Thank you!

PR #1963, opened 2022-05-15
ElementHandle - Improve Debugger Display

amaitland:
- Add RemoteObject.ClassName and RemoteObject.Description
- Add ElementHandle.DebuggerDisplay attribute to use newly added ClassName and Description

Doesn't change any runtime functionality, just improves the debugging output

Viewing an `ElementHandle` in the debugger will now show for example `HtmlButtonElement@button`.


kblok:
Thank you!

PR #1967, opened 2022-05-24
change SA1600 rulset to Error

MajeQafouri:
Reference Issue: https://github.com/hardkoded/puppeteer-sharp/issues/1384

### DIFF
--- a/lib/PuppeteerSharp.ruleset
+++ b/lib/PuppeteerSharp.ruleset
@@ -10,7 +10,7 @@
 
     <Rule Id="SA1624" Action="Error" /> <!-- Error SA1624: Because the property only contains a visible get accessor, the documentation summary text should begin with 'Gets'. (SA1624)-->
     <Rule Id="SA1028" Action="Error" /> <!-- Error SA1028: Code should not contain trailing whitespace (SA1028)-->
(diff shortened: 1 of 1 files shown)


kblok:
@MajeQafouri did you miss any commit here?

kblok:
@MajeQafouri these issues are about enabling that rule and fixing the errors that come from enabling it.

MajeQafouri:
> @MajeQafouri these issues are about enabling that rule and fixing the errors that come from enabling it.

Ok, I will check it again

MajeQafouri:
@kblok 
During the build process, StyleCop.Analyzers package does not allow to build the solution, because of the lack of documentation. do you have any suggestions?

kblok:
If SA1600 wasn't set as an error before it's because we have missing documentation :)
I see that much of the missing documentation is from internal elements. I would try disabling that using [this](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/Configuration.md#documentation-requirements).
After that, we will need to add all the missing documentation. Constructors tend to have the same style, the same for exceptions. [pptr.dev](https://pptr.dev/) is a good source of content.

MajeQafouri:
>

Issue #1945, opened 2022-04-26
ContentFrameAsync() exception occurred in iframe elementHandle


namilkimfree:
## Looking for some help? Get priority support by [sponsoring the project](https://github.com/sponsors/hardkoded)/

**Before you file a bug, have you:**

* Tried upgrading to newest version of PuppeteerSharp, to see if your issue has already been resolved and released?
* Checked existing open *and* closed [issues](https://github.com/kblok/puppeteer-sharp/issues?utf8=%E2%9C%93&q=is%3Aissue), to see if the issue has already been reported?
* Tried reproducing your problem in a new isolated project?
* Considered if this is a general question and not a bug?. For general questions please use [StackOverflow](https://stackoverflow.com/search?q=puppeteer-sharp).

### Description

[Description of the issue]
iframeElementHandle.ContentFrameAsync()  throw exception 

Frame 8B6EB85116EFEB0ABBEFC65625234E35 not found

### Complete minimal example reproducing the issue

```csharp
// Arrange
var options = new LaunchOptions { /*  */ };
var chromiumRevision = BrowserFetcher.DefaultRevision;
var browser = await Puppeteer.LaunchAsync(options, chromiumRevision);
var page = browser.NewPageAsync();
   await        page.GoToAsync("https://forum.netmarble.com/enn_ko/view/33/132235",
                new NavigationOptions() {WaitUntil = new[] {WaitUntilNavigation.Load, WaitUntilNavigation.Networkidle0}});



// Act
...

var elementHandle = await page.QuerySelectorAsync("#contentsDetail iframe");

var frame = await elementHandle.ContentFrameAsync(); 

// Assert

--> throw exception  Frame 8B6EB85116EFEB0ABBEFC65625234E35 not found
...
```

### Expected behavior:

[What you expect to happen]
not throw exception

### Actual behavior:

[What actually happens]
but throw exception

### Versions

* Which version of PuppeteerSharp are you using? 7.0.0
* Which .NET runtime and version are you targeting? E.g. .NET 5

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

Yaagod:
Confirming, `ContentFrameAsync()` is not working on any website in current version (7.0). 6.x works fine,

kblok:
Can you try v7.1 @Yaagod ?

PR #1972, opened 2022-06-10
Add failing ContentFrame test

Androbin:
See issue #1945


lib/PuppeteerSharp.Tests/ElementHandleTests/ContentFrameTests.cs
```diff
@@ -24,5 +28,17 @@ public async Task ShouldWork()
             var frame = await elementHandle.ContentFrameAsync();
             Assert.Equal(Page.FirstChildFrame(), frame);
         }
+
+        [PuppeteerTest("elementhandle.spec.ts", "ElementHandle.contentFrame", "should work headful")]
```
kblok, line 32: We don't need this if it's not puppeteer.

```suggestion
```

Androbin:
It looks like the frame gets detached for no reason

kblok:
Can you see if you get the same in puppeteer in node?

Androbin:
Node works like a charm:
```js
require('puppeteer').launch({ headless: false }).then(async browser => {
  const page = await browser.newPage()
  await page.goto('http://localhost:8081/frame-example.html')
  const elementHandle = await page.$('iframe')
  const frame = await elementHandle.contentFrame()
  const content = await frame.content()
  console.log(content)
  await browser.close()
})
```

kblok:
Good catch @Androbin

kblok:
Fixed on https://github.com/hardkoded/puppeteer-sharp/pull/1979

PR #1979, opened 2022-06-28
Add missing oopif tests

kblok:
# New Features
 * ElementHandle.ClickablePointAsync is now public
 * New Page.WaitForFrameAsync


lib/PuppeteerSharp/Page.cs
```diff
@@ -1830,6 +1831,76 @@ void requestEventListener(object sender, RequestEventArgs e)
             return await requestTcs.Task.ConfigureAwait(false);
         }
 
+        /// <summary>
+        /// Waits for a frame.
+        /// </summary>
+        /// <example>
+        /// <code>
+        /// <![CDATA[
+        /// var frame = await page.WaitForFrameAsync(frane=> request.Name == "Test");
```
amaitland, line 1840: Looks like there's a mistake here. There's no request variable.

lib/PuppeteerSharp/Page.cs
```diff
@@ -1830,6 +1831,76 @@ void requestEventListener(object sender, RequestEventArgs e)
...
+        {
+            var timeout = options?.Timeout ?? DefaultTimeout;
+            var frameTcs = new TaskCompletionSource<Frame>(TaskCreationOptions.RunContinuationsAsynchronously);
+
+            void frameEventListener(object sender, FrameEventArgs e)
+            {
+                if (predicate(e.Frame))
+                {
+                    frameTcs.TrySetResult(e.Frame);
+                    FrameManager.FrameAttached -= frameEventListener;
```
amaitland, line 1873: Unsubscribe from both events?

lib/PuppeteerSharp/LifecycleWatcher.cs
```diff
@@ -163,10 +182,12 @@ public void Dispose()
         {
             _frameManager.LifecycleEvent -= FrameManager_LifecycleEvent;
             _frameManager.FrameNavigatedWithinDocument -= NavigatedWithinDocument;
+            _frameManager.FrameNavigated -= Navigated;
```
amaitland, line 185: Unsubscribe from FrameSwapped also?

lib/PuppeteerSharp/ElementHandle.cs
```diff
@@ -629,20 +633,26 @@ public async Task DragAndDropAsync(ElementHandle target, int delay = 0)
...
         }
 
-        private async Task<(decimal X, decimal Y)> ClickablePointAsync()
+        /// <summary>
+        /// Returns the middle point within an element unless a specific offset is provided.
+        /// </summary>
+        /// <param name="offset">Optional offset</param>
+        /// <exception cref="PuppeteerException">When the node is not visible or not an HTMLElement</exception>
+        /// <returns>A <see cref="Task"/> that resolves to the clickable point</returns>
+        public async Task<BoxModelPoint> ClickablePointAsync(BoxModelPoint? offset = null)
```
amaitland, line 647: Rather than using Boxmodelpoint for the offset I'd like to propose a new Offset struct be created/used. Matching the current puppeteer implementation. The Offset class in puppeteer has specific descriptions for the x/y properties. 

https://pptr.dev/api/puppeteer.offset

The struct is largely self documenting then.

kblok: Disagree :)

amaitland: Puppeteer has an offset object so puppeteer Sharp should have one too 

> x-offset for the clickable point relative to the top-left corder of the border box.

The properties are documented differently to Boxmodelpoint

Androbin:
@kblok When can we expect v7.1.0?

lib/PuppeteerSharp/DOMWorld.cs
```diff
@@ -26,6 +32,7 @@ public DOMWorld(FrameManager frameManager, Frame frame, TimeoutSettings timeoutS
 
             WaitTasks = new ConcurrentSet<WaitTask>();
             _detached = false;
+            _client.MessageReceived += Client_MessageReceived;
```
amaitland, line 35: Unsubscribe when the DOM world has detached?

kblok:
@Androbin I can cook a version :)

PR #1984, opened 2022-07-03
ElementHandle.ScrollIntoViewIfNeededAsync public

Mephisto090:
So that the user can choose a variable point where to press in the button

https://github.com/hardkoded/puppeteer-sharp/pull/1985


kblok (approved):
It's private on puppeteer. But I agree on exposing this.

PR #1985, opened 2022-07-03
Make ElementHandle.ClickablePointAsync public

Mephisto090:
So that the user can choose a variable point where to press in the button

https://github.com/hardkoded/puppeteer-sharp/pull/1984

I don't want to open a ticket now. Wouldn't it make sense to write your own function so that it is variable for the user and not fixed in the middle? Or Optional variables in the ClickAsync function? If you want it, I'll be happy to open an extra ticket for it.


kblok:
@Mephisto090 you PR is not building

Mephisto090:
why? :(

Mephisto090:
I capitalized public, not lowercase :(

PR #1993, opened 2022-07-15
DoubleClick Function

Mephisto090:
Add a DoubleClick Function


kblok:
There is no `DoubleClick` in puppeteer. It's performed passing `ClickCount = 2`

Issue #1397, opened 2020-02-29
Enable SA1413 rule


kblok:
Change the action for SA1413 on the [ruleset file](https://github.com/kblok/puppeteer-sharp/blob/master/lib/PuppeteerSharp.ruleset) to Error, and make the solution build

MursalovAltun:
@kblok Hey, can you please review the PR that enables this rule and fixes all the errors?

PR #2013, opened 2022-09-13
Enabled and fixed issues for the rule SA1413.

MursalovAltun:
#1397 - Set the action to error and fixed all errors.


kblok (approved):
Thank you!
Sorry for the delay. I took a few days off :)

PR #2028, opened 2022-10-17
Enable SA1204 rule

pikami:
### Summary

Change the action for SA1204 to Error, and update solution files.

### Ticket Link

Fixes https://github.com/hardkoded/puppeteer-sharp/issues/1392


kblok (approved):
Thank you!

PR #2038, opened 2022-11-08
Allow for passing IJSHandle as argument to EvaluateFunctionAsync/EvaluateFunctionHandleAsync

amaitland:
Motivation for this change is to allow for [PuppeteerSharp.Dom](https://github.com/ChromiumDotNet/PuppeteerSharp.Dom) to pass it's custom `DOM` object instances as `IJSHandle` args to EvaluateFunctionAsync/EvaluateFunctionHandleAsync. 

Currently you have to cast to `JSHandle`. e.g. https://github.com/ChromiumDotNet/PuppeteerSharp.Dom/blob/b1f862ee24dcc2a025dfea43d198597dbdedf17c/PuppeteerSharp.Dom.Tests/ElementHandleTests/BoundingBoxTests.cs#L61

### DIFF
--- a/lib/PuppeteerSharp/Helpers/RemoteObjectHelper.cs
+++ b/lib/PuppeteerSharp/Helpers/RemoteObjectHelper.cs
@@ -8,8 +8,37 @@ using PuppeteerSharp.Messaging;
 
 namespace PuppeteerSharp.Helpers
 {
-    internal class RemoteObjectHelper
+    internal static class RemoteObjectHelper
     {
+        internal static object FormatArgument(this IJSHandle jSHandle, ExecutionContext context)
+        {
+            if (jSHandle.Disposed)
+            {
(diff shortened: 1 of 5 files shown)


lib/PuppeteerSharp/Helpers/RemoteObjectHelper.cs
```diff
@@ -8,8 +8,37 @@
 
 namespace PuppeteerSharp.Helpers
 {
-    internal class RemoteObjectHelper
+    internal static class RemoteObjectHelper
     {
+        internal static object FormatArgument(this IJSHandle jSHandle, ExecutionContext context)
```
kblok, line 13: How about moving this to `PuppeteerHandleExtensions`?

amaitland: Done

PR #2058, opened 2022-12-27
Enable rule SA1600

akshay-zz:
Fix #1967


kblok:
Can you add some explanation here?

akshay-zz:
> Can you add some explanation here?

The PR was not ready yet. But even though just wanted to confirm we are not making it mandatory for the `internal` to be documented. But forcing other things to be documented?

akshay-zz:
https://github.com/hardkoded/puppeteer-sharp/pull/1967#issuecomment-1141068698

> If SA1600 wasn't set as an error before it's because we have missing documentation :) I see that much of the missing documentation is from internal elements. I would try disabling that using [this](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/Configuration.md#documentation-requirements). After that, we will need to add all the missing documentation. Constructors tend to have the same style, the same for exceptions. [pptr.dev](https://pptr.dev/) is a good source of content.

@kblok does the above comment by me hold true for this statement?

kblok:
That comment still makes sense. We don't want to comment internal methods.

akshay-zz:
@kblok Can you please review the PR?

kblok (approved):
Nice!

PR #2088, opened 2023-02-13
Fix single file publish

### LINKED ISSUE #2073: Publish using "Produce single file" - Unhandled exception. System.ArgumentException: The path is empty. (Parameter 'path')
### Description

PuppeteerSharp 9.0.0 introduced a bug that will cause to fail the parameterless ctor of BrowserFetcher when using in some published exe with "Produce single file" enabled.

- Get any example code that uses the parameterless ctor of BrowserFetcher and put it into a console application (.net6 and .net7 are known to cause the issue)
- Publish project with option "Produce single file" enabled.
- Run the published exe
- [Minimal repro](https://github.com/stesee/PuppeteerSharpMinimalRepro)

### Expected behavior:

Running without exception, like PuppeteerSharp 8.

### Actual behavior:

```
Unhandled exception. System.ArgumentException: The path is empty. (Parameter 'path')
   at System.IO.Path.GetFullPath(String path)
   at System.IO.FileInfo..ctor(String originalPath, String fullPath, String fileName, Boolean isNormalized)
   at System.IO.FileInfo..ctor(String fileName)
   at PuppeteerSharp.BrowserFetcher.GetExecutablePath() in C:\projects\puppeteer-sharp\lib\PuppeteerSharp\BrowserFetcher.cs:line 354
   at PuppeteerSharp.BrowserFetcher..ctor() in C:\projects\puppeteer-sharp\lib\PuppeteerSharp\BrowserFetcher.cs:line 53
   at Program.<Main>$(String[] args) in C:\source\repos\PuppeteerSharpMinimalRepro\PuppeteerSharpMinimalRepro\Program.cs:line 5
   at Program.<Main>(String[] args)
```

### Versions

* Which version of PuppeteerSharp are you using? 9.0.0 and 9.0.1
* Which .NET runtime and version are you targeting? .net6 and .net7

### Additional Information - Workarround - Use overloads of the BrowserFetcher ctor.

```
var options = new BrowserFetcherOptions();
options.Path = Path.GetTempPath();
using var browserFetcher = new BrowserFetcher(options);
var revisionInfo = await browserFetcher.DownloadAsync();
await using var browser = await Puppeteer.LaunchAsync(new LaunchOptions { Headless = true, ExecutablePath = revisionInfo.ExecutablePath });
await page.ScreenshotAsync("google.jpg");
```

stesee:
closes #2073


lib/PuppeteerSharp.SingleFileDeployment/Program.cs
```diff
@@ -0,0 +1,9 @@
+using PuppeteerSharp;
```
kblok, line 1: Can we call this project `PuppeteerSharp.Tests.SingleFileDeployment`?

lib/PuppeteerSharp.TestServer/PuppeteerSharp.TestServer.csproj
```diff
@@ -14,4 +14,7 @@
     <PackageReference Include="Microsoft.AspNetCore" Version="2.2.0" />
     <PackageReference Include="Microsoft.AspNetCore.StaticFiles" Version="2.2.0" />
   </ItemGroup>
+  <Target Name="testCertCheck" BeforeTargets="BeforeBuild" Condition="!Exists('testCert.cer')" >
```
kblok, line 17: Nice!

lib/PuppeteerSharp.Tests/SingleFileDeployment/PublishedSelfContainedBinaryShouldWork.cs
```diff
@@ -0,0 +1,94 @@
+using System.Diagnostics;
+using System.IO;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace PuppeteerSharp.Tests.SingleFileDeployment
+{
+    public class PublishedSelfContainedBinary
```
kblok, line 8: Let's name this file `PublishedSelfContainedBinary.cs`

kblok (commented):
Looking good!

lib/PuppeteerSharp.Tests/SingleFileDeployment/PublishedSelfContainedBinaryShouldWork.cs
amaitland: SelfContained != SingleFile 

> Publishing your app as self-contained produces an application that includes the .NET runtime and libraries, and your application and its dependencies. Users of the application can run it on a machine that doesn't have the .NET runtime installed.

SelfContained refers to the `runtime` being included with the application. 

> Single-file deployment is available for both the [framework-dependent deployment model](https://learn.microsoft.com/en-us/dotnet/core/deploying/#publish-framework-dependent) and [self-contained applications](https://learn.microsoft.com/en-us/dotnet/core/deploying/#publish-self-contained).

For reference.

If the intention is to test `Single File` build then one of the following is probably more descript:

- `PublishSingleFileTests`
- `PublishSingleFileDeploymentTests`

lib/PuppeteerSharp.Tests/SingleFileDeployment/PublishedSelfContainedBinary.cs
```diff
@@ -0,0 +1,95 @@
...
+            var expectedBinaryPath = Path.Combine(absolutePath, $"bin/Release/net6.0/publish/{projectName}.exe");
+
+            DeleteIfExists(expectedBinaryPath);
+
+            using var process = new Process
+            {
+                StartInfo = new ProcessStartInfo
+                {
+                    FileName = "dotnet",
+                    Arguments = $"publish {absolutePath} --configuration Release -f net6.0 -p:PublishProfile=FolderProfile",
```
amaitland, line 66: I don't see a [FolderProfile.pubxml](https://github.com/stesee/PuppeteerSharpMinimalRepro/blob/master/PuppeteerSharpMinimalRepro/Properties/PublishProfiles/FolderProfile.pubxml) file in this `PR`. Am I missing something?

For simplicity is it better to just include the args here directly?

```
-r win-x64 -p:PublishSingleFile=true --self-contained false
```

lib/PuppeteerSharp.Tests/SingleFileDeployment/FactRunableOnWindowsAttribute.cs
```diff
@@ -0,0 +1,18 @@
+using System.Runtime.InteropServices;
+using Xunit;
+
+namespace PuppeteerSharp.Tests.SingleFileDeployment
+{
+    internal class FactRunableOnWindowsAttribute : FactAttribute
```
amaitland, line 6: There's other `xunit` test attributes in https://github.com/hardkoded/puppeteer-sharp/tree/master/lib/PuppeteerSharp.Tests/Attributes

Should this live in there?

lib/PuppeteerSharp/BrowserFetcher.cs
```diff
@@ -348,25 +354,29 @@ internal static Platform GetCurrentPlatform()
         internal static string GetExecutablePath()
         {
             DirectoryInfo assemblyDirectory = new(AppContext.BaseDirectory);
-            if (!assemblyDirectory.Exists || !File.Exists(Path.Combine(assemblyDirectory.FullName, "PuppeteerSharp.dll")))
+            if (!assemblyDirectory.Exists || !File.Exists(Path.Combine(assemblyDirectory.FullName, nameof(Puppeteer) + ".dll")))
```
amaitland, line 357: Can you double check that `nameof(Puppeteer)` is equal to `PuppeteerSharp`? I think this check will always fail with this change.

Getting the name of the assembly from the actual assembly would be better.  Possibly something like the following?

```c#
var assembly = typeof(Puppeteer).Assembly;
var assemblyName = assembly.GetName().Name;
```

lib/PuppeteerSharp.Tests/SingleFileDeployment/PublishedSelfContainedBinary.cs
amaitland: `--use-current-runtime` instead of specifying a `RID` would actually be preferable.

Make it easer to take this test cross platform

lib/PuppeteerSharp.Tests/SingleFileDeployment/PublishedSelfContainedBinaryShouldWork.cs
stesee: I choosed 'PublishSingleFileTests'

lib/PuppeteerSharp.Tests/PublishSingleFileTests/PublishSingleFileTests.cs
```diff
@@ -0,0 +1,96 @@
...
+            }
+
+            Assert.True(isExited);
+
+            Assert.True(File.Exists(actualFilePath), $"StdOut: {outputResult}\nStdErr: {errorResult}\n");
+        }
+
+        private static string GetStreamOutput(StreamReader stream)
+        {
+            var outputReadTask = Task.Run(() => stream.ReadToEnd());
```
kblok, line 50: Why do we need this `Task.Run`?

lib/PuppeteerSharp.Tests/PublishSingleFileTests/PublishSingleFileTests.cs
```diff
@@ -0,0 +1,96 @@
...
+
+            return outputReadTask.Result;
+        }
+
+        private static string DotnetPublishSingleFile(string projectName)
+        {
+            var absolutePath = Path.GetFullPath("../../../../" + projectName);
```
kblok, line 57: Let's use `Path.Join` here.

lib/PuppeteerSharp.Tests/PublishSingleFileTests/PublishSingleFileTests.cs
```diff
@@ -0,0 +1,96 @@
...
+            var expectedBinaryPath = Path.Combine(absolutePath, $"bin/Release/net6.0/win-x64/publish/{projectName}.exe");
```
kblok, line 58: Can we add a comment making sure that  `net6.0` is in sync with the project file?

kblok:
@amaitland as you stepped into this, I'll wait for your approval.

lib/PuppeteerSharp.Tests/PublishSingleFileTests/PublishSingleFileTests.cs
stesee: I followed https://stackoverflow.com/questions/7160187/standardoutput-readtoend-hangs

lib/PuppeteerSharp.Tests/PublishSingleFileTests/PublishSingleFileTests.cs
stesee: There is no Path.Join in net48. Used Path.Combine.

lib/PuppeteerSharp.Tests/PublishSingleFileTests/PublishSingleFileTests.cs
stesee: Solved that using `publish -o ./publish`

lib/PuppeteerSharp.Tests/PublishSingleFileTests/PublishSingleFileTests.cs
amaitland: Does using [ReadToEndAsync](https://learn.microsoft.com/en-us/dotnet/api/system.io.streamreader.readtoendasync?view=net-7.0) work around the issue?

stesee: I will give it a try... I am not sure optimizing testcode is worth it... 😌

lib/PuppeteerSharp/BrowserFetcher.cs
```diff
@@ -347,26 +353,29 @@ internal static Platform GetCurrentPlatform()
...
-                {
-                    assemblyLocation = codeBase.LocalPath;
-                }
-                else
+                var assemblyLocation = assembly.Location;
+
+                if (string.IsNullOrEmpty(assemblyLocation))
                 {
-                    assemblyLocation = assembly.Location;
+                    var singleFilePublishFilePathForBrowserExecutables = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), nameof(Puppeteer));
```
amaitland, line 366: Should the folder name be `PuppeteerSharp`? 

Thoughts?

lib/PuppeteerSharp.Tests/PublishSingleFileTests/PublishSingleFileTests.cs
```diff
@@ -0,0 +1,96 @@
...
+            var expectedBinaryPath = Path.Combine(absolutePath, $"publish/{projectName}.exe");
+
+            DeleteIfExists(expectedBinaryPath);
+
+            using var process = new Process
+            {
+                StartInfo = new ProcessStartInfo
+                {
+                    FileName = "dotnet",
+                    Arguments = $"publish {absolutePath} --configuration Release -r win-x64 -p:PublishSingleFile=true --self-contained false --use-current-runtime -o ./publish",
```
amaitland, line 67: If you remove `-r win-x64` then hopefully `--use-current-runtime` should pickup the current `RID`.

https://github.com/dotnet/sdk/issues/10449

lib/PuppeteerSharp.Tests/PublishSingleFileTests/PublishSingleFileTests.cs
amaitland: I'll defer to @kblok on this one, what's your thoughts?

lib/PuppeteerSharp/BrowserFetcher.cs
stesee: There are a bunch of aspects:
- it is common to use the vendor name for the folder - techies will know what to expect in `PuppeteerSharp`
- shared use of downloaded chrome in multiple puppeteer-sharp consumers - for good or bad
- length vs max path length
- complexity to do magic to get the name out of consuming applications

I would like to keep it simple.

lib/PuppeteerSharp.Tests/PublishSingleFileTests/PublishSingleFileTests.cs
stesee: As far as I remember reading StdOut and StdErr at the same time is causing deadlocks and using [ReadToEndAsync](https://learn.microsoft.com/en-us/dotnet/api/system.io.streamreader.readtoendasync?view=net-7.0) wont fix that. The output of them isnt used to assert the expectation and could be dropped entirely but if the assertion will fail it is the information you will need to find out what went wrong.

lib/PuppeteerSharp/BrowserFetcher.cs
amaitland: Unless I'm mistaken here `nameof(Puppeteer)` = `Puppeteer` 

I think the folder should be named `PuppeteerSharp`.

> I would like to keep it simple.

You've gone way more complex than I was actually meaning 😄

lib/PuppeteerSharp.Tests/PublishSingleFileTests/PublishSingleFileTests.cs
amaitland: I've seen the following pattern used a few times, I've not used it myself yet.

https://stackoverflow.com/a/53504707/4583726

amaitland:
@kblok Just ran the latest changes locally, happy to approve for merge 👍

PR #2091, opened 2023-02-24
Prefer IsEmpty over Count/Any

jnyrup:
[Roslyn Aanalyzers](https://github.com/dotnet/roslyn-analyzers) reports [CA1836: Prefer IsEmpty over Count when available](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1836).

E.g. [`ConcurrentDictionary.IsEmpty`](https://source.dot.net/#System.Collections.Concurrent/System/Collections/Concurrent/ConcurrentDictionary.cs,1479) has a lock-free fast path, whereas [`ConcurrentDictionary.Count`](https://source.dot.net/#System.Collections.Concurrent/System/Collections/Concurrent/ConcurrentDictionary.cs,1116) always acquires all locks.

### DIFF
--- a/lib/PuppeteerSharp/DOMWorld.cs
+++ b/lib/PuppeteerSharp/DOMWorld.cs
@@ -48,7 +48,7 @@ namespace PuppeteerSharp
             _logger = _client.Connection.LoggerFactory.CreateLogger<DOMWorld>();
         }
 
-        internal ICollection<WaitTask> WaitTasks { get; set; }
+        internal ConcurrentSet<WaitTask> WaitTasks { get; set; }
 
         internal Frame Frame { get; }
 
@@ -143,7 +143,7 @@ namespace PuppeteerSharp
         internal void Detach()
         {
             _detached = true;
-            while (WaitTasks.Count > 0)
+            while (!WaitTasks.IsEmpty)
             {
                 WaitTasks.First().Terminate(new Exception("waitForFunction failed: frame got detached."));
             }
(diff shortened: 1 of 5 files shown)


lib/PuppeteerSharp/DOMWorld.cs
```diff
@@ -48,7 +48,7 @@ public DOMWorld(CDPSession client, FrameManager frameManager, Frame frame, Timeo
             _logger = _client.Connection.LoggerFactory.CreateLogger<DOMWorld>();
         }
 
-        internal ICollection<WaitTask> WaitTasks { get; set; }
+        internal ConcurrentSet<WaitTask> WaitTasks { get; set; }
```
jnyrup, line 51: Changing the type of this internal property was necessary to get access to `IsEmpty` at line 146.

kblok:
Can you add that to the [ruleset](https://github.com/hardkoded/puppeteer-sharp/blob/master/lib/PuppeteerSharp.ruleset)?

jnyrup:
> Can you add that to the [ruleset](https://github.com/hardkoded/puppeteer-sharp/blob/master/lib/PuppeteerSharp.ruleset)?

Sure!
CA1836 was added in a newer version of `Microsoft.NetCore.Analyzers` than what is currently referenced.
Upgrading the nuget package triggers several new analyzers, so I've left out upgrading the package from this PR to keep it focused.

kblok:
> Upgrading the nuget package triggers several new analyzers, so I've left out upgrading the package from this PR to keep it focused.

A PR with that upgrade is welcome! We can mark all failing rules as `None`, and I can create issues for those.

kblok:
Thank you!

PR #2092, opened 2023-02-24
Fix duplicated test

jnyrup:
[`ShouldWorkWithFakeClickingTargetBlankAndReNoopener`](https://github.com/hardkoded/puppeteer-sharp/blob/c18812b66ae07e1c0ecda5af5573a2d12accad8f/lib/PuppeteerSharp.Tests/PageTests/PageEventsPopupTests.cs#L64-L80) and [`ShouldWorkWithFakeClickingTargetBlankAndRelNoopener`](https://github.com/hardkoded/puppeteer-sharp/blob/c18812b66ae07e1c0ecda5af5573a2d12accad8f/lib/PuppeteerSharp.Tests/PageTests/PageEventsPopupTests.cs#L82-L98) are identical.
The first one should use `rel=opener`

Cross-checked with [puppeteer source](https://github.com/puppeteer/puppeteer/blob/415da9230066e5b399f60963b639bfbafdc8ec57/test/src/page.spec.ts#L296-L346)

### DIFF
--- a/lib/PuppeteerSharp.Tests/PageTests/PageEventsPopupTests.cs
+++ b/lib/PuppeteerSharp.Tests/PageTests/PageEventsPopupTests.cs
@@ -63,10 +63,10 @@ namespace PuppeteerSharp.Tests.PageTests
 
         [PuppeteerTest("page.spec.ts", "Page.Events.Popup", "should work with clicking target=_blank and with rel=opener")]
         [SkipBrowserFact(skipFirefox: true)]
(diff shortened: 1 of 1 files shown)


kblok:
Nice! Thank you!

kblok:
It seems that the test didn't pass :(

jnyrup:
I missed that the assertions also changed in the corresponding puppeteer test.
https://github.com/puppeteer/puppeteer/blob/415da9230066e5b399f60963b639bfbafdc8ec57/test/src/page.spec.ts#L294

PR #2093, opened 2023-02-24
Analyzers

jnyrup:
Builds on top of #2091

New ignored analyzers
* [CA1805: Do not initialize unnecessarily.](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1805)
* [CA2012: Use ValueTasks correctly](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2012)
* [IL3000: Avoid accessing Assembly file path when publishing as a single file](https://learn.microsoft.com/en-us/dotnet/core/deploying/single-file/warnings/il3000)


kblok:
Could you fix the conflict?

lib/PuppeteerSharp/.editorconfig
```diff
@@ -0,0 +1,173 @@
+[*.cs]
```
kblok, line 1: I recall having issues with `.editorconfig` in other environments. Does it work on CI/CD? Can we force an error to see if that works?

jnyrup: `.editorconfig` works with CI, see now failing pipelines.

Maybe the bad experience you previously had was CI pipelines with older build tools, where `.editorconfig` wasn't yet supported?
E.g. Visual Studio has [supported `.editorconfig`](https://learn.microsoft.com/en-us/visualstudio/code-quality/analyzers-faq?view=vs-2022) since v16.3 released in [September 2019](https://learn.microsoft.com/en-us/visualstudio/releases/2019/release-notes-v16.3#16.3.0)

kblok:
CI is not building here.

kblok:
> I added a second commit to trigger CA1836 and break the pipelines.
> Are you referring to anything other than that?

I just tested this on VS for Mac, and it worked. I think editorconfig didn't work there before.
So let's roll back the breaking commit.

I also got this warning. `Could not find rule set file "../PuppeteerSharp.ruleset"`.

kblok:
> I added a second commit to trigger CA1836 and break the pipelines.
> Are you referring to anything other than that?

I just tested this on VS for Mac, and it worked. I think editorconfig didn't work there before.
So let's roll back the breaking commit.

I also got this warning. `Could not find rule set file "../PuppeteerSharp.ruleset"`.

jnyrup:
> So let's roll back the breaking commit.

done!

> I also got this warning. `Could not find rule set file "../PuppeteerSharp.ruleset"`.

I missed that `PuppeteerSharp.ruleset` was also referenced from `PuppeteerSharp.Tooling`.

kblok:
@jnyrup I would rather move the config to `lib/.editorconfig` if it works :)

jnyrup:
> @jnyrup I would rather move the config to `lib/.editorconfig` if it works :)

Test projects then also picks up all rules. How would you like to have that handled? That's the three cases I tried to outline above.

kblok:
> Test projects then also picks up all rules. How would you like to have that handled? That's the three cases I tried to outline above.

What do you think about adding an editorconfig in the tests, relaxing all rules?

jnyrup:
> > Test projects then also picks up all rules. How would you like to have that handled? That's the three cases I tried to outline above.
> 
> What do you think about adding an editorconfig in the tests, relaxing all rules?

Makes sense, after having another look, the vast majority were tests not using `ConfigureAwait`.

kblok (approved):
Looks good!

PR #2096, opened 2023-03-01
Test - Use Assert.ThrowsAsync for exceptions instead of ContinueWith

amaitland:
After reviewing `ContinueWith` usage in the tests there are a couple of cases which can be improved by changing to Assert.ThrowsAsync.


kblok:
Makes sense. Let's see if they CI likes it.

PR #2107, opened 2023-03-20
Improve timing related tests

jnyrup:
This PR consists of two commits related to timing:

1) JavaScript's [`Date.now()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now) is documented as 
> The `Date.now()` static method returns the number of milliseconds elapsed since the [epoch](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#the_ecmascript_epoch_and_timestamps), which is defined as the midnight at the beginning of January 1, 1970, UTC.

C#[ `DateTime.Now`](https://learn.microsoft.com/en-us/dotnet/api/system.datetime.now?view=net-8.0) on the other hand is documented as:  
> Gets a [DateTime](https://learn.microsoft.com/en-us/dotnet/api/system.datetime?view=net-8.0) object that is set to the current date and time on this computer, expressed as the local time.

If running a test while changing from/to DST, the difference between two calls to `DateTime.Now` is no longer only an approximation of the time to run some code, but also the difference in UTC offsets.
By changing usages of `DateTime.Now` to `DateTime.UtcNow` we remove that "race-condition" and brings the intention closer to the puppeteer tests.

The second commit includes idea from https://github.com/hardkoded/puppeteer-sharp/pull/2090#issuecomment-1445203137 to avoid silently ignoring exceptions in some tests.
@amaitland I did see your wish for a more elegant solution, but do you agree that including the less pretty solution for now is better than keeping status quo?

### DIFF
--- a/lib/PuppeteerSharp.Tests/PageTests/WaitForNetworkIdleTests.cs
+++ b/lib/PuppeteerSharp.Tests/PageTests/WaitForNetworkIdleTests.cs
@@ -22,7 +22,12 @@ namespace PuppeteerSharp.Tests.PageTests
             var t2 = DateTime.UtcNow;
 
             await Page.GoToAsync(TestConstants.EmptyPage);
-            var task = Page.WaitForNetworkIdleAsync().ContinueWith(x => t1 = DateTime.UtcNow);
+            var task = Page.WaitForNetworkIdleAsync()
+                .ContinueWith(x =>
+                {
+                    if (x.IsFaulted) throw x.Exception;
+                    return t1 = DateTime.UtcNow;
+                });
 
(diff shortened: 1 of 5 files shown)


kblok (approved):

amaitland:
> but do you agree that including the less pretty solution for now is better than keeping status quo?

@jnyrup Absolutely 👍 I've rewritten a couple of cases (https://github.com/hardkoded/puppeteer-sharp/pull/2096) where `ContinueWith` wasn't necessary, that's as far as I've got so far.

PR #2118, opened 2023-03-28
TaskQueue - Make class sealed and simplify Dispose

amaitland:
The class is now **sealed** which means we can remove the `protected` methods

### DIFF
--- a/lib/PuppeteerSharp/Helpers/TaskQueue.cs
+++ b/lib/PuppeteerSharp/Helpers/TaskQueue.cs
@@ -62,7 +62,7 @@ namespace PuppeteerSharp.Helpers
 
             if (dispose)
             {
-                _ = Task.Run(() => DisposeAsync());
+                _ = Task.Run(() => DisposeAsyncCore());
             }
 
             _isDisposed = true;
--- a/lib/PuppeteerSharp/Helpers/TaskQueue.cs
+++ b/lib/PuppeteerSharp/Helpers/TaskQueue.cs
@@ -5,7 +5,7 @@ using System.Threading.Tasks;
 
 namespace PuppeteerSharp.Helpers
 {
-    internal class TaskQueue : IDisposable, IAsyncDisposable
+    internal sealed class TaskQueue : IDisposable, IAsyncDisposable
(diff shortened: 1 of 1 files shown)


amaitland:
@kblok What do you think about making `TaskQueue` sealed? The `Dispose/DisposeAsync` code could be greatly simplified.

amaitland:
I've updated the code to make the class `sealed`. I haven't run the tests locally yet.

kblok (approved):

jnyrup:
Do any of you know why we we call `_ = Task.Run(() => DisposeAsync());` instead of `_semaphore.Wait();`?
I can't find anything about it in #1826.


Also I guess we should also check `_isDisposed` in `DisposeAsync()`

amaitland:
> Do any of you know why we we call `_ = Task.Run(() => DisposeAsync());` instead of `_semaphore.Wait();`?

I wasn't involved in the original implementation. 

Changing to `Wait()` would mean `Dispose` would then block, I'm not sure if that would have any side effects, so I left it alone for now.

> Also I guess we should also check `_isDisposed` in `DisposeAsync()`

I was planning on revisiting this one, the question I have is do we need a `Thread Safe` guard or not? From a quick look at the usage it's probably fine for just a bool check, I'm still somewhat leaning towards using `Interlocked.Exchange`.

```c#
private int _disposed;

{
    {
        return;
    }

    try
    {
        await _semaphore.WaitAsync().ConfigureAwait(false);
    }
    catch (ObjectDisposedException)
    {
        // Ignore
    }

    _semaphore.Dispose();
}
```

amaitland:
> I'm still somewhat leaning towards using `Interlocked.Exchange`.

Promised change as a commit https://github.com/amaitland/puppeteer-sharp/commit/2ca28aa75cb1d59d5e481d881d82afece9e6927c

kblok:
@amaitland @jnyrup Honestly, I never find issues around de disposable path as you are working on.
I did work on many things around page creation, requests lists, etc, etc.
So, as you can see :),  I'm open to suggestions and ideas here.

PR #2119, opened 2023-04-01
Update samples to use PuppeteerSharp v9.0.2

cestrand:
Updated PuppeteerSharp to 9.0.2 in the samples.

There is no constant called `BrowserFetcher.DefaultRevision` since version 7.0.0. Changed it to `BrowserFetcher.DefaultChromiumRevision`.

Fixed broken CSS selectors in samples. All samples working correctly now.


kblok (approved):
Thank you!

PR #2123, opened 2023-04-07
TaskQueue - ThreadSafe Dispose(Async) guard

amaitland:
Adds a double dispose guard to `DisposeAsync`. As proposed in https://github.com/hardkoded/puppeteer-sharp/pull/2118#issuecomment-1489622471 

Same pattern is used within the `dotnet runtime` itself. 

https://github.com/dotnet/runtime/blob/38b81ba83ab2b27a730a9524401d74043d172049/src/libraries/System.Net.Quic/src/System/Net/Quic/QuicListener.cs#L356

### DIFF
--- a/lib/PuppeteerSharp/Helpers/TaskQueue.cs
+++ b/lib/PuppeteerSharp/Helpers/TaskQueue.cs
@@ -10,32 +10,29 @@ namespace PuppeteerSharp.Helpers
         [SuppressMessage("Usage", "CA2213: Disposable fields should be disposed", Justification = "The disposable field is being disposed asynchronously.")]
         [SuppressMessage("CodeQuality", "IDE0079:Remove unnecessary suppression", Justification = "The CA2213 suppression is actually necessary.")]
         private readonly SemaphoreSlim _semaphore;
-        private bool _isDisposed;
+        private int _disposed;
 
         internal TaskQueue() => _semaphore = new SemaphoreSlim(1);
 
         public void Dispose()
         {
-            if (_isDisposed)
+            if (Interlocked.Exchange(ref _disposed, 1) != 0)
             {
                 return;
             }
 
-            _ = Task.Run(() => DisposeAsync());
-
-            _isDisposed = true;
+            _semaphore.Wait();
+            _semaphore.Dispose();
         }
 
         public async ValueTask DisposeAsync()
         {
-            try
+            if (Interlocked.Exchange(ref _disposed, 1) != 0)
             {
-                await _semaphore.WaitAsync().ConfigureAwait(false);
+                return;
             }
-            catch (ObjectDisposedException)
-            {
-                // Ignore
-            }
+
+            await _semaphore.WaitAsync().ConfigureAwait(false);
 
             _semaphore.Dispose();
         }


lib/PuppeteerSharp/Helpers/TaskQueue.cs
```diff
@@ -10,24 +10,27 @@ internal sealed class TaskQueue : IDisposable, IAsyncDisposable
...
 
         public void Dispose()
         {
-            if (_isDisposed)
+            if (Interlocked.Exchange(ref _disposed, 1) != 0)
             {
                 return;
             }
 
             _ = Task.Run(() => DisposeAsync());
```
jnyrup, line 24: I haven't had a chance to test this out, but won't we we immediately return from `DisposeAsync` as `_disposed` has already been changed?

jnyrup: I had some time to test, and `_semaphore` is not disposed when calling `Dispose` for the reason described above.

If we decouple `Dispose` and `DisposeAsync` completely and call `SemaphoreSlim.Wait()` in `Dispose`, here are some unit tests to verify it.
With the thread safe guards, we should in addition be able to remove the try/catch of `ObjectDisposedException`.

```cs
public void Dispose()
{
    {
        return;
    }

    _semaphore.Wait();
    _semaphore.Dispose();
}

{
    {
        return;
    }

    await _semaphore.WaitAsync().ConfigureAwait(false);
    _semaphore.Dispose();
}
```

```cs
public class TaskQueueTests
{
    [Fact]
    public void ShouldDisposeSemaphoreWhenDisposing()
    {
        var taskQueue = new TaskQueue();
        taskQueue.Dispose();

        var semaphore = GetSemaphore(taskQueue);
        Assert.Throws<ObjectDisposedException>(() => semaphore.AvailableWaitHandle);
    }

    [Fact]
    public void ShouldNotThrowWhenDisposingMultipleTimes()
    {
        var taskQueue = new TaskQueue();
        taskQueue.Dispose();

        // Can be disposed a second time without throwing
        taskQueue.Dispose();
    }

    [Fact]
    public async Task ShouldDisposeSemaphoreWhenDisposingAsync()
    {
        var taskQueue = new TaskQueue();
        await taskQueue.DisposeAsync().ConfigureAwait(false);

        var semaphore = GetSemaphore(taskQueue);
        Assert.Throws<ObjectDisposedException>(() => semaphore.AvailableWaitHandle);
    }

    [Fact]
    public async Task ShouldNotThrowWhenDisposingMultipleTimesAsync()
    {
        var taskQueue = new TaskQueue();
        await taskQueue.DisposeAsync().ConfigureAwait(false);

        // Can be disposed a second time without throwing
        await taskQueue.DisposeAsync().ConfigureAwait(false);
    }

    private static SemaphoreSlim GetSemaphore(TaskQueue queue) =>
        (SemaphoreSlim)typeof(TaskQueue).GetField("_semaphore", BindingFlags.Instance | BindingFlags.NonPublic).GetValue(queue);
}
```

kblok:
I'll wait for your approval here @jnyrup

lib/PuppeteerSharp/Helpers/TaskQueue.cs
amaitland: > I haven't had a chance to test this out, but won't we we immediately return from DisposeAsync as _disposed has already been changed?

You are absolutely correct! My mistake!

jnyrup (approved):

PR #2128, opened 2023-04-08
Rework some disposing

jnyrup:
Found some more places related to disposing to rework.
I've split the changes into three separate commits to ease reviewing.

### DIFF
--- a/lib/PuppeteerSharp/Browser.cs
+++ b/lib/PuppeteerSharp/Browser.cs
@@ -295,10 +295,18 @@ namespace PuppeteerSharp
                 targetFilter,
                 isPageTargetCallback);
 
-            initAction?.Invoke(browser);
+            try
+            {
+                initAction?.Invoke(browser);
 
-            await browser.AttachAsync().ConfigureAwait(false);
-            return browser;
+                await browser.AttachAsync().ConfigureAwait(false);
+                return browser;
+            }
+            catch
+            {
+                await browser.DisposeAsync().ConfigureAwait(false);
+                throw;
+            }
         }
 
         internal IEnumerable<string> GetCustomQueryHandlerNames()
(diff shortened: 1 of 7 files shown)


lib/PuppeteerSharp/Browser.cs
```diff
@@ -295,10 +295,18 @@ public async ValueTask DisposeAsync()
...
+                initAction?.Invoke(browser);
 
-            await browser.AttachAsync().ConfigureAwait(false);
-            return browser;
+                await browser.AttachAsync().ConfigureAwait(false);
+                return browser;
+            }
+            catch
+            {
+                await browser.DisposeAsync().ConfigureAwait(false);
```
jnyrup, line 307: if `initAction` or `AttachAsync` throws an exception, we now explicitly dispose `browser`.

lib/PuppeteerSharp/WaitTask.cs
```diff
@@ -171,8 +171,14 @@ internal class WaitTask : IDisposable
 
         public void Dispose()
         {
-            Dispose(true);
-            GC.SuppressFinalize(this);
```
jnyrup, line 175: Since the class is now `sealed` and there is not finalizer, there's no longer a need for `GC.SuppressFinalize(this)`.

lib/PuppeteerSharp/Helpers/TempDirectory.cs
```diff
@@ -31,31 +31,20 @@ public TempDirectory(string path)
...
-        protected virtual void Dispose(bool disposing)
-        {
-            if (_deleteTask == null)
-            {
-                _ = DeleteAsync();
-            }
-        }
-
-        private static async Task DeleteAsync(string path, CancellationToken cancellationToken = default)
+        private static async Task DeleteAsync(string path)
```
jnyrup, line 47: `DeleteAsync(CancellationToken cancellationToken = default)` has never been used outside this class since it was introduced in #585, so `cancellationToken` is always `CancellationToken.None`.

I don't have the full overview of what #585 does, so I refrained from removing the fire-and-forget invocation of of `DeleteAsync`.

lib/PuppeteerSharp/Browser.cs
kblok: Good point.

kblok (approved):
Great job going through this!

kblok (approved):

PR #2131, opened 2023-04-08
TaskHelper improvements

jnyrup:
dc2e1137e888cacfc3598c41941098053d4e2020
When `task` times out and `timeoutAction` does not throw an exception we resumed awaiting `task` instead of returning early like `public static async Task<T> WithTimeout<T>(this Task<T> task, Action timeoutAction, TimeSpan timeout)` does.

71a9c95f7251dda0ed9f654178f4c956783ef007
When the `cancellationToken` is canceled we don't stop the `TimeoutTask`.

### DIFF
--- a/lib/PuppeteerSharp.Tests/TaskHelperTests.cs
+++ b/lib/PuppeteerSharp.Tests/TaskHelperTests.cs
@@ -72,5 +72,17 @@ namespace PuppeteerSharp.Tests
 
             await task.WithTimeout(() => Task.FromException(new TimeoutException()), TimeSpan.FromTicks(1), token);
         }
+
+        [Fact]
+        public async Task ShouldStopExecutionWhenTokenIsCanceled()
+        {
+            using var tokenSource = new CancellationTokenSource();
+            var tcs = new TaskCompletionSource<bool>();
+
+            var task = tcs.Task.WithTimeout(() => Task.FromException(new TimeoutException()), TimeSpan.FromHours(42), tokenSource.Token);
+            tokenSource.Cancel();
+
+            await task;
+        }
     }
 }
--- a/lib/PuppeteerSharp.Tests/TaskHelperTests.cs
+++ b/lib/PuppeteerSharp.Tests/TaskHelperTests.cs
@@ -1,0 +1,76 @@
+using System;
+using System.Threading;
+using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


lib/PuppeteerSharp.Tests/TaskHelperTests.cs
kblok, line 1: We could create a new Namespace/Directory for tests on utility classes.

lib/PuppeteerSharp/Helpers/TaskHelper.cs
```diff
@@ -187,14 +192,32 @@ private static async Task<bool> TimeoutTask(Task task, TimeSpan timeout)
...
+
+            var tcs = new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously);
+            using (var linkedCancellationToken = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken))
+            {
+                linkedCancellationToken.CancelAfter(timeout);
+                using (linkedCancellationToken.Token.Register(s => ((TaskCompletionSource<bool>)s).TrySetResult(true), tcs))
+                {
+                    return tcs.Task == await Task.WhenAny(task, tcs.Task).ConfigureAwait(false);
+                }
+            }
```
kblok, line 220: Can we use [using statements](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/using)?

jnyrup: I've changed the outermost using statements to using declarations.

For the return value of `Register` I've kept the using statements, since
```cs
using var _ = linkedCancellationToken.Token.Register(s => ((TaskCompletionSource<bool>)s).TrySetResult(true), tcs);
```
triggers SA1312

See https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3057#issuecomment-552490781

kblok (approved):

lib/PuppeteerSharp.Tests/TaskHelperTests.cs
jnyrup: Can I postpone that to another PR?
Then I'll move both this and `TaskQueueTests` in there

kblok: Sure!

PR #2135, opened 2023-04-08
Use IAsyncLifetime for async initialization/disposal

jnyrup:
Removes some blocking calls around xunit initialization/disposal
Found on https://jimmybogard.com/integration-testing-with-xunit/#sharedfixturedesign

https://github.com/xunit/xunit/blob/1892224c60f8499b42ee1d8a521d65995ef32aea/src/xunit.execution/Sdk/Frameworks/Runners/XunitTestCollectionRunner.cs#L107-L121

https://github.com/xunit/xunit/blob/1892224c60f8499b42ee1d8a521d65995ef32aea/src/xunit.execution/Sdk/Frameworks/Runners/XunitTestCollectionRunner.cs#L56-L64


kblok (approved):
Good one. Although we should move to NUnit eventually.

PR #2136, opened 2023-04-08
Cleanups

jnyrup:
Some minor cleanups that have no behavioral changes.
When relevant I've included the context I could find in the commit message.

### DIFF
--- a/lib/PuppeteerSharp/AriaQueryHandlerFactory.cs
+++ b/lib/PuppeteerSharp/AriaQueryHandlerFactory.cs
@@ -43,7 +43,7 @@ namespace PuppeteerSharp
 
             Func<IElementHandle, string, WaitForSelectorOptions, Task<IElementHandle>> waitFor = async (IElementHandle root, string selector, WaitForSelectorOptions options) =>
             {
-                var frame = (root as ElementHandle).Frame;
+                var frame = ((ElementHandle)root).Frame;
                 var element = await frame.SecondaryWorld.AdoptHandleAsync(root).ConfigureAwait(false);
 
                 Func<string, Task<IElementHandle>> func = (string selector) => queryOne(element, selector);
@@ -63,7 +63,7 @@ namespace PuppeteerSharp
 
             Func<IElementHandle, string, Task<IElementHandle[]>> queryAll = async (IElementHandle element, string selector) =>
             {
-                var exeCtx = element.ExecutionContext as ExecutionContext;
+                var exeCtx = (ExecutionContext)element.ExecutionContext;
                 var world = exeCtx.World;
                 var ariaSelector = ParseAriaSelector(selector);
                 var res = await QueryAXTreeAsync(exeCtx.Client, element, ariaSelector.Name, ariaSelector.Role).ConfigureAwait(false);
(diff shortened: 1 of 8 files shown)


lib/PuppeteerSharp/AriaQueryHandlerFactory.cs
```diff
@@ -26,10 +26,6 @@ internal static InternalQueryHandler Create()
             {
                 var queryOptions = ParseAriaSelector(selector);
                 var res = await QueryAXTreeAsync(((ElementHandle)element).Client, element, queryOptions.Name, queryOptions.Role).ConfigureAwait(false);
-                if (res.FirstOrDefault().BackendDOMNodeId == null)
-                {
-                    return null;
-                }
```
kblok, line 32: Why do you think we need to remove this?

jnyrup: We don't _need_ to remove it, but it's evaluating the lazy `IEnumerable<AXTreeNode>` returned by `QueryAXTreeAsync` without needing to.

If `QueryAXTreeAsync` returns an empty sequence, `FirstOrDefault()` returns null and `.BackendDOMNodeId` throws a `NullReferenceException`.

If `QueryAXTreeAsync` returns a non-empty sequence we can just return `res.First().BackendDOMNodeId` directly instead of calling `First`/`FirstOrDefault` twice.

Here's a reduced [example](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEUCuA7AHwAEAmARgFgAoIgBgAIiyUBuauxsgOgBkBLPAEc2VdmQCcACgCKOGFACeAQQAaAFVgwlAZwV4wkgJRcAYnyjaMAeSgARGADMAhjgA2GI4ZFMps+cvVNHT0DYzMLD0MvahiaMgBmAB4BDAA+ej9FVQ0YLV19I2oAb2p6MvoUivoAXnpaEXL6AHcACz5XGHpJDFwYQ1LykqpGxp9JABIAIgAHKAgAExwwAQBzeiK+AF9J6OGRsqYyRgB2CoBqM4byzepNoA).

kblok: I think the logic was wrong there. We should return null if `backendDOMNodeId` is `0` or `null`.

jnyrup: If I understand https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer-core/src/common/AriaQueryHandler.ts correctly, `queryAXTree` can return an empty sequence which is handled in `queryOne` by
```ts
(await AsyncIterableUtil.first(this.queryAll(element, selector))) ?? null
```

We could then return `return res.FirstOrDefault()?.BackendDOMNodeId;`

kblok: Technically we should return `null` if the value is `0`, right?

jnyrup: I don't see anywhere in [AriaQueryHandler.ts](https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer-core/src/common/AriaQueryHandler.ts) code where they have special handling of `0`.

`return res.FirstOrDefault()?.BackendDOMNodeId;` will return `null` if `res` is empty and otherwise `BackendDOMNodeId` of the first item.

PR #2190, opened 2023-04-21
Waittask refactor

kblok:
This will match the upstream change [here](https://github.com/puppeteer/puppeteer/pull/8943/files#diff-b5735653f04183106c7769b4c707f0a9d7742f9367334fc9f38f79bb9a596786).

### DIFF
--- a/lib/PuppeteerSharp/CustomQueriesManager.cs
+++ b/lib/PuppeteerSharp/CustomQueriesManager.cs
@@ -154,9 +154,9 @@ namespace PuppeteerSharp
                 internalHandler.WaitFor = async (IElementHandle root, string selector, WaitForSelectorOptions options) =>
                 {
                     var frame = (root as ElementHandle).Frame;
-                    var element = await frame.PuppeteerWorld.AdoptHandleAsync(root).ConfigureAwait(false);
+                    var element = await frame.PuppeteerWorld.AdoptHandleAsync(root).ConfigureAwait(false) as IElementHandle;
 
-                    return await frame.PuppeteerWorld.WaitForSelectorInPageAsync(handler.QueryOne, selector, options).ConfigureAwait(false);
+                    return await frame.PuppeteerWorld.WaitForSelectorInPageAsync(handler.QueryOne, element, selector, options).ConfigureAwait(false);
                 };
             }
 
--- a/lib/PuppeteerSharp/TaskManager.cs
+++ b/lib/PuppeteerSharp/TaskManager.cs
@@ -6,7 +6,7 @@ namespace PuppeteerSharp
 {
     internal class TaskManager
     {
-        internal ConcurrentSet<WaitTask> WaitTasks { get; set; } = new ConcurrentSet<WaitTask>();
+        private ConcurrentSet<WaitTask> WaitTasks { get; } = new ConcurrentSet<WaitTask>();
 
         internal void Add(WaitTask waitTask) => WaitTasks.Add(waitTask);
 
--- a/lib/PuppeteerSharp/TaskManager.cs
+++ b/lib/PuppeteerSharp/TaskManager.cs
@@ -1,0 +1,31 @@
+﻿using System;
+using System.Linq;
+using PuppeteerSharp.Helpers;
+
+namespace PuppeteerSharp
+{
+    internal class TaskManager
+    {
+        internal ConcurrentSet<WaitTask> WaitTasks { get; set; } = new ConcurrentSet<WaitTask>();
+
+        internal void Add(WaitTask waitTask) => WaitTasks.Add(waitTask);
+
+        internal void Delete(WaitTask waitTask) => WaitTasks.Remove(waitTask);
+
+        internal void RerunAll()
+        {
+            foreach (var waitTask in WaitTasks)
+            {
+                _ = waitTask.Rerun();
+            }
+        }
+
+        internal void TerminateAll(Exception exception)
+        {
+            while (!WaitTasks.IsEmpty)
+            {
+                _ = WaitTasks.First().TerminateAsync(exception);
+            }
+        }
+    }
+}
--- a/lib/PuppeteerSharp/WaitTask.cs
+++ b/lib/PuppeteerSharp/WaitTask.cs
@@ -89,8 +89,11 @@ namespace PuppeteerSharp
         {
             try
             {
-                var context = await _isolatedWorld.GetExecutionContextAsync().ConfigureAwait(false);
-                await System.Threading.Tasks.Task.WhenAll(_bindings.Select(binding => _isolatedWorld.AddBindingToContextAsync(context, binding.Name))).ConfigureAwait(false);
+                if (_bindings.Length > 0)
+                {
+                    var context = await _isolatedWorld.GetExecutionContextAsync().ConfigureAwait(false);
+                    await System.Threading.Tasks.Task.WhenAll(_bindings.Select(binding => _isolatedWorld.AddBindingToContextAsync(context, binding.Name))).ConfigureAwait(false);
+                }
 
                 if (_pollingInterval.HasValue)
                 {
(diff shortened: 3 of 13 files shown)


lib/PuppeteerSharp/WaitTask.cs
```diff
@@ -183,84 +87,127 @@ public void Dispose()
...
 
-        internal void Terminate(Exception exception)
+        private Exception GetBadException(Exception exception)
         {
-            _terminated = true;
-            _taskCompletion.TrySetException(exception);
-            Cleanup();
+            if (exception.Message.Contains("Execution context is not available in detached frame"))
+            {
+                return new PuppeteerException("Waiting failed: Frame detached");
```
jnyrup, line 201: Should `exception` be set as the inner exception of the constructed `PuppeteerException`?

lib/PuppeteerSharp/WaitTask.cs
```diff
@@ -183,84 +87,127 @@ public void Dispose()
...
                 return;
             }
 
-            if (exception?.Message.Contains("Cannot find context with specified id") == true)
-            {
-                return;
-            }
+            _terminated = true;
+            _isolatedWorld.TaskManager.Delete(this);
+            Cleanup(); // This matches the clearTimeout upstream
```
jnyrup, line 172: `_isolatedWorld.TaskManager.Delete(this);` is also called inside `Cleanup`

lib/PuppeteerSharp/WaitTask.cs
```diff
@@ -183,84 +87,127 @@ public void Dispose()
...
-            Cleanup();
+                    await _poller.DisposeAsync().ConfigureAwait(false);
+                    _poller = null;
+                }
+                catch (Exception)
+                {
+                    // swallow error.
+                }
+            }
         }
```
jnyrup, line 195: I'm not into the execution flow of when `Rerun` is executed, but I just wanted to point out a _potential_ race-condition.
If one thread running `Rerun` assigns a new value to `_poller` while another thread is in here, we might invoke `EvaluateFunctionAsync` and `DisposeAsync` on different instances.

So my suggestion is to save `_poller` to a local variable and act on that instead of the mutable field.

E.g. 
```cs
if (_poller is { } poller)
{
    try
    {
        await poller.EvaluateFunctionAsync(@"async poller => {
            await poller.stop();
        }").ConfigureAwait(false);

        await poller.DisposeAsync().ConfigureAwait(false);
        _poller = null;
    }
    catch (Exception)
    {
        // swallow error.
    }
}
```

amaitland (commented):
A few comments inline.

I haven't run the code or stepped through with a debugger, so some comments maybe just need simple clarifications the new behaviour is correct.

lib/PuppeteerSharp/WaitTask.cs
jnyrup: To put our suggestions into code
```cs
if (_poller is { } poller)
{
    _poller = null;

    await using (poller.ConfigureAwait(false))
    {
        try
        {
            await poller.EvaluateFunctionAsync(@"async poller => {
                await poller.stop();
            }").ConfigureAwait(false);
        }
        catch (Exception)
        {
            // swallow error.
        }
    }
}
```

lib/PuppeteerSharp/TaskManager.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System;
+using System.Linq;
+using PuppeteerSharp.Helpers;
+
+namespace PuppeteerSharp
+{
+    internal class TaskManager
+    {
+        internal ConcurrentSet<WaitTask> WaitTasks { get; set; } = new ConcurrentSet<WaitTask>();
```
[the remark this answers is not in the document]
jnyrup: It seems it can also be read-only

lib/PuppeteerSharp/WaitTask.cs
```diff
@@ -183,84 +87,127 @@ public void Dispose()
...
 
-            if (exception == null &&
-                await _isolatedWorld.EvaluateFunctionAsync<bool>("s => !s", success)
-                    .ContinueWith(
-                        task => task.IsFaulted || task.Result,
-                        TaskScheduler.Default)
-                    .ConfigureAwait(false))
+                var success = await _poller.EvaluateFunctionHandleAsync("poller => poller.result()").ConfigureAwait(false);
+                _result.TrySetResult(success);
+                await TerminateAsync().ConfigureAwait(false);
```
jnyrup, line 150: If we dispose `_poller` here anyways, it seems we could keep `_poller` a local variable?
Less mutable state just makes thread-safety easier.

It could look something like this

```cs
internal async Task Rerun()
{
    IJSHandle poller = null;

    try
    {
        var context = await _isolatedWorld.GetExecutionContextAsync().ConfigureAwait(false);
        await System.Threading.Tasks.Task.WhenAll(_bindings.Select(binding => _isolatedWorld.AddBindingToContextAsync(context, binding.Name))).ConfigureAwait(false);

        if (_pollingInterval.HasValue)
        {
            poller = [...]
        }
        else if (_polling == WaitForFunctionPollingOption.Raf)
        {
            poller = [...]
        }
        else
        {
            poller = [...]
        }

        await poller.EvaluateFunctionAsync("poller => poller.start()").ConfigureAwait(false);

        var success = await poller.EvaluateFunctionHandleAsync("poller => poller.result()").ConfigureAwait(false);
        _result.TrySetResult(success);
        await Terminate(poller).ConfigureAwait(false);
        Terminate();
    }
    catch (Exception ex)
    {
        if (poller is not null)
        {
            await Terminate(poller).ConfigureAwait(false);
        }

        var exception = GetBadException(ex);
        if (exception != null)
        {
            Terminate(exception);
        }
    }
}

internal void Terminate(Exception exception = null)
{
    // The timeout timer might call this method on cleanup
    if (_terminated)
    {
        return;
    }

    _terminated = true;
    _isolatedWorld.TaskManager.Delete(this);
    Cleanup(); // This matches the clearTimeout upstream

    if (exception != null)
    {
        _result.TrySetException(exception);
    }
}

private async Task Terminate(IJSHandle poller)
{
    try
    {
        await using (poller.ConfigureAwait(false))
        {
            await poller.EvaluateFunctionAsync(@"async poller => {
                await poller.stop();
            }").ConfigureAwait(false);
        }
    }
    catch (Exception)
    {
        // swallow error.
    }
}
```

jnyrup: Wait... I just saw the case when `_poller` has not completed yet, but `TerminateAsync` should terminate it anyways.

I wanted to guard against repeated calls to `Rerun` overwriting `_poller` as that could mean that only the _last_ poller is disposed. 

It's tricky 😃

lib/PuppeteerSharp/WaitTask.cs
kblok: Yes!

lib/PuppeteerSharp/WaitTask.cs
kblok: Good catch. Removing it from `Cleanup`.

lib/PuppeteerSharp/WaitTask.cs
kblok: That's interesting. It's quite an edge scenario the best thread-safe we're, the better. 
Thank you for this idea!

lib/PuppeteerSharp/TaskManager.cs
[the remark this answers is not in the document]
kblok: Done!

lib/PuppeteerSharp/CustomQueriesManager.cs
```diff
@@ -154,9 +154,9 @@ private static InternalQueryHandler MakeQueryHandler(CustomQueryHandler handler)
                 internalHandler.WaitFor = async (IElementHandle root, string selector, WaitForSelectorOptions options) =>
                 {
                     var frame = (root as ElementHandle).Frame;
-                    var element = await frame.PuppeteerWorld.AdoptHandleAsync(root).ConfigureAwait(false);
+                    var element = await frame.PuppeteerWorld.AdoptHandleAsync(root).ConfigureAwait(false) as IElementHandle;
```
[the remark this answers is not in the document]
kblok: `AdoptHandleAsync` expects and returns an `IJSHandle`, we should honor that.

lib/PuppeteerSharp/WaitTask.cs
```diff
@@ -183,84 +87,127 @@ public void Dispose()
...
-                        _pollingInterval ?? (object)_polling,
-                        _timeout,
-                    }.Concat(_args).ToArray()).ConfigureAwait(false);
-            }
-            catch (Exception ex)
-            {
-                exception = ex;
-            }
+                var context = await _isolatedWorld.GetExecutionContextAsync().ConfigureAwait(false);
+                await System.Threading.Tasks.Task.WhenAll(_bindings.Select(binding => _isolatedWorld.AddBindingToContextAsync(context, binding.Name))).ConfigureAwait(false);
```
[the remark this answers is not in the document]
kblok: Good idea.

lib/PuppeteerSharp/WaitTask.cs
```diff
@@ -183,84 +87,127 @@ public void Dispose()
...
 
-            if (_terminated || runCount != _runCount)
-            {
-                if (success != null)
+                if (_pollingInterval.HasValue)
```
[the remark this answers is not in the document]
kblok: The problem we have is that in puppeteer the type is `'raf' | 'mutation' | number`. In PuppeteerSharp we solved that with two props. `PollingInterval` of type `int?` and `Polling` of type `WaitForFunctionPollingOption`.

lib/PuppeteerSharp/WaitTask.cs
```diff
@@ -183,84 +87,127 @@ public void Dispose()
...
+                }
+                catch (Exception)
+                {
+                    // swallow error.
+                }
+            }
         }
 
-        internal void Terminate(Exception exception)
+        private Exception GetBadException(Exception exception)
```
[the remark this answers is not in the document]
kblok: Sounds fair. Sometimes I don't know whether we should copy comments or not.

kblok:
@amaitland @jnyrup, I think I went through all your comments. I appreciate it a lot!!! Thank you!
Some tests are still not passing, so there is still some work to do. But feel free to keep pushing :)

PR #2192, opened 2023-04-25
WaitTask Refactor - Fix Evaluation failed: TypeError: fun is not a function

amaitland:
This should hopefully fix the `Evaluation failed: TypeError: fun is not a function` error. At least it does when I run it locally.

It looks like the whitespace is causing the `new Function` call to eval to null.

Lots of different places the whitespace could be trimmed, Could just as easily be done in `.Net`.

Reformatting the `const string` is also an option, my concern would be that formatting changes could break this again.


kblok:
Would you mind trimming the source? I think this change can easily get lost with an upstream change.

kblok:
Thank you!

PR #2206, opened 2023-05-09
Switch to System.Text.Json

### LINKED ISSUE #2022: Upgrade to System.Text.Json from Newtonsoft.Json
Thank you for this wonderful library. We have observed that this library still uses newtonsoft json, is it possible to upgrade to a better and faster json serializer library like system.text.json?

image

akshay-zz:
Fix #2022


akshay-zz:
Hi, I was earlier digging into it. Didn’t able to test it by running it. Got busy and didn’t able to look into it. We can abounded this PR as currently I have no plan to continue on this.

PR #2214, opened 2023-05-26
Fix race condition that caused LaunchAsync to never resolve for chrome

### LINKED ISSUE #2202: `Puppeteer.LaunchAsync` hangs every once in a while waiting for a page to open
### Description

`Puppeteer.LaunchAsync` hangs occasionally - perhaps once every thousand calls (not sure of the exact frequency).

I could try writing sample code to launch the browser repeatedly, haven't done that yet.

### Expected behavior:

`Puppeteer.LaunchAsync` should return (or at least throw an exception).

### Actual behavior:

`Puppeteer.LaunchAsync` never returns. It appears from examining the process tree that Chrome launches successfully, but PuppeteerSharp never completes the connection process. Downloads are not a factor here since we're launching a preinstalled binary.

### Versions

* Which version of PuppeteerSharp are you using? 9.1.0
* Which .NET runtime and version are you targeting? 7.0

### Additional Information

I've experienced this issue only in an Ubuntu container, but since I haven't tried running PuppeteerSharp heavily on Windows, I'm not sure whether the OS is a factor.

If I `exec` into the container and kill the Chrome process, Puppeteer throws the following exception:
```
PuppeteerSharp.ProcessException: Failed to create connection  ---> System.TimeoutException: Timeout of 30000 ms exceeded
    at PuppeteerSharp.Helpers.TaskHelper.WithTimeout[T](Task`1 task, TimeSpan timeout, Func`2 exceptionFactory) in /home/runner/work/puppeteer-sharp/puppeteer-sharp/lib/PuppeteerSharp/Helpers/TaskHelper.cs:line 180
    at PuppeteerSharp.Browser.WaitForTargetAsync(Func`2 predicate, WaitForOptions options) in /home/runner/work/puppeteer-sharp/puppeteer-sharp/lib/PuppeteerSharp/Browser.cs:line 235
    at PuppeteerSharp.Launcher.LaunchAsync(LaunchOptions options) in /home/runner/work/puppeteer-sharp/puppeteer-sharp/lib/PuppeteerSharp/Launcher.cs:line 83
    --- End of inner exception stack trace ---
    at PuppeteerSharp.Launcher.LaunchAsync(LaunchOptions options) in /home/runner/work/puppeteer-sharp/puppeteer-sharp/lib/PuppeteerSharp/Launcher.cs:line 89
    at PuppeteerSharp.Launcher.LaunchAsync(LaunchOptions options) in /home/runner/work/puppeteer-sharp/puppeteer-sharp/lib/PuppeteerSharp/Launcher.cs:line 95
```

So it seems that PuppeteerSharp is hanging waiting for a page to open, and the timeout it sets isn't actually being triggered. If it's possible to fix this so that the timeout actually throws, that would be a great start, since it's not a big deal to me if PuppeteerSharp fails to launch every once in a while.

Just reviewing `Browser.WaitForTargetAsync` without knowing much about the PuppeteerSharp internals it seems like there is the potential for a race condition if the `TargetCreated` event is fired just after `Targets()` is called on line 224 and just before the event handler is attached on line 232. However, I'm not sure why the timeout isn't eventually throwing in that case. If that is the issue, it seems like it would be easy to reproduce by inserting a `Task.Delay` between lines 224 and 232.

leonardo-fernandes:
Fixes #2202

Only process received messages after the discovered targets have been stored.


leonardo-fernandes:
From my tests, FirefoxTargetManager doesn't seem affected by the same race condition.

leonardo-fernandes:
I spent some hours trying to understand why some unit tests were failing with my previous commit.

It looks like the await inside of the `OnMessageReceived` was continuing the flow of execution in `ProcessIncomingMessage` and eventually releasing the semaphore from the TaskQueue which would continue to process the next messages, even though the previous message was still not processed.

I think this is caused by the `OnMessageReceived` method returning void instead of `Task`, so if it encounters an asynchronous call the continuation is queued but not returned to the caller, and the caller cannot await from it. Sorry if I can't explain it any better than this :)

I tried making `OnMessageReceived` synchronous but hit a deadlock. The `OnAttachedToTarget` terminates with an `await Task.WhenAll` and two `SendAsync` tasks. These tasks will only terminate when the return messages are processed. But the thread executing `OnAttachedToTarget` is waiting for the `SendAsync` inside of a semaphore, which prevents the processing of more messages.

In my opinion, the code in `OnAttachedToTarget` shouldn't await for these, and could simply run the tasks with Task.Run. There's nothing being gained from the await because (due to `OnMessageReceived` returning void) the caller will resume the control flow. There's a few more logic using await inside `OnAttachedToTarget` which I wouldn't dare to touch, so this is where I stopped the analysis. If we can make `OnAttachedToTarget` synchronous, then the incoming messages will be processed in sequence and would greatly reduce the chances of race conditions.

kblok:
@leonardo-fernandes Event handling is tricky here. I wanted to get rid of them and use delegates (with tasks). But having multiple listeners is challenging. 
That's why we have tools like `AsyncDictionaryHelper`.

I'm looking at your findings in #2202. But honestly. This looks more like a hard patch than a solution. Let's see what we can find :)

Thank you for your time!

lib/PuppeteerSharp/ChromeTargetManager.cs
```diff
@@ -132,6 +140,7 @@ private async void OnMessageReceived(object sender, MessageEventArgs e)
                 switch (e.MessageID)
                 {
                     case "Target.attachedToTarget":
+                        await _targetDiscoveryCompletionSource.Task.ConfigureAwait(false);
```
kblok, line 143: Let's move this inside `OnAttachedToTarget`.

lib/PuppeteerSharp/ChromeTargetManager.cs
```diff
@@ -23,6 +23,7 @@ internal class ChromeTargetManager : ITargetManager
         private readonly ConcurrentDictionary<string, TargetInfo> _discoveredTargetsByTargetId = new();
         private readonly ConcurrentDictionary<ICDPConnection, List<TargetInterceptor>> _targetInterceptors = new();
         private readonly List<string> _targetsIdsForInit = new();
+        private readonly TaskCompletionSource<bool> _targetDiscoveryCompletionSource = new();
```
kblok, line 26: Let's add a comment mentioning that this is .NET and we need this to avoid race conditions.

leonardo-fernandes:
> The problem here is that I think it's not the only place where we send a message inside an event. Event handling is tricky in Puppeteer-Sharp. I wanted to get rid of them and use delegates (with tasks). But having multiple listeners is challenging. That's why we need to find alternatives like the `AsyncDictionaryHelper`. It's pretty much like you did here. Find the race and try to solve it.

I know it's tricky, I tried changing the code to be synchronous and hit a few deadlocks. But today I was able to change `OnAttachedToTarget` to synchronous. I'll commit it to another branch, and I am keen on your feedback because it made all tests pass reliably (except `PublishSingleFileTests`, for some reason it has never passed in my machine).

lib/PuppeteerSharp/ChromeTargetManager.cs
kblok: Good question. As opposed to "uptream" (the original puppeteer implementation we follow).
It's important for us to follow which changes are not in upstream, because we don't want to remove our custom code.

leonardo-fernandes:
Ok, I compared the code with the upstream puppeteer, and I believe the upstream code does not await on the `OnAttachedToTarget`. See https://github.com/puppeteer/puppeteer/blob/f342c26d953e51bf262123e2e97276eff8ee2841/packages/puppeteer-core/src/common/ChromeTargetManager.ts#L174

The call to `#onAttachedToTarget` (which is an async function) does not await on it.

I also have a clearer understanding of why this might not be an issue in upstream. There are some difference in semantics between async in js and .net. One key difference is that js is single-threaded, so it's not possible in js to have a thread switch between the execution of `_targetsIdsForInit.Remove(target.TargetId);` and `FinishInitializationIfReady();`. But in .NET, it's possible that a second thread adds a new element into `_targetsIdsForInit` after the last element had been removed, and so the check inside `FinishInitializationIfReady` would fail.

As I'm feeling more confident, and seeing that it improves the test results, I've included the changes for not awaiting `OnAttachedToTarget` for your consideration. I've also included a missing return, see https://github.com/puppeteer/puppeteer/blob/f342c26d953e51bf262123e2e97276eff8ee2841/packages/puppeteer-core/src/common/ChromeTargetManager.ts#L335, and without which some tests would fail randomly.

kblok:
I'm not being able to get a full green in AppVeyor :(

leonardo-fernandes:
I know. I'm trying to reproduce the failures in my local machine, but so far haven't been able to.
I'll need more time then to understand what's happening.

leonardo-fernandes:
I've been running test cases over and over to detect race conditions / deadlocks. I'll document what I find here. Note that I've added logging in multiple places, and line numbers might not match exactly what's committed.

Today I found the following exception led to a deadlock when executing PuppeteerSharp.Tests.LauncherTests.PuppeteerConnectTests.ShouldBeAbleToCloseRemoteBrowser:

```
Value cannot be null.
Parameter name: logger
   at Microsoft.Extensions.Logging.LoggerExtensions.Log(ILogger logger, LogLevel logLevel, EventId eventId, Exception exception, String message, Object[] args)
   at Microsoft.Extensions.Logging.LoggerExtensions.LogTrace(ILogger logger, String message, Object[] args)
   at PuppeteerSharp.Browser.CreateTarget(TargetInfo targetInfo, CDPSession session) in C:\src\puppeteer-sharp\lib\PuppeteerSharp\Browser.cs:line 516
   at PuppeteerSharp.ChromeTargetManager.OnTargetCreated(TargetCreatedResponse e) in C:\src\puppeteer-sharp\lib\PuppeteerSharp\ChromeTargetManager.cs:line 195
   at PuppeteerSharp.ChromeTargetManager.OnMessageReceived(Object sender, MessageEventArgs e) in C:\src\puppeteer-sharp\lib\PuppeteerSharp\ChromeTargetManager.cs:line 156
```

This is what seems to have happened.

1. `ChromeTargetManager` is created in Browser.cs before the `Browser._logger` field is populated.
2. The `ChromeTargetManager` constructor attaches the `OnMessageReceived` listener (again, this is before `Browser._logger` is populated).
3. If a message is received in the connection, there's a chance that it is processed before the `ChromeTargetManager` constructor finishes executing (and before `Browser._logger` is populated).
4. If that happens, accessing `Browser._logger` will throw an exception. I had instrumented `Browser.CreateTarget` with logging, which caused an exception.
5. Exception is caught in `OnMessageReceived` and it calls `_connection.Close()`. This eventually calls `Connection.Dispose()` which calls `TaskQueue.Dispose()`, which waits for the semaphore.
6. However, the semaphore is held by the same thread (which was processing a message), which causes the deadlock.

Stack of the deadlocked thread (TaskQueue appearing twice):

```
 	PuppeteerSharp.dll!PuppeteerSharp.Helpers.TaskQueue.Dispose() Line 21	C#
 	PuppeteerSharp.dll!PuppeteerSharp.Connection.Dispose(bool disposing) Line 240	C#
 	PuppeteerSharp.dll!PuppeteerSharp.Connection.Dispose() Line 102	C#
 	PuppeteerSharp.dll!PuppeteerSharp.Browser.Disconnect() Line 191	C#
 	PuppeteerSharp.dll!PuppeteerSharp.Browser.CloseCoreAsync() Line 457	C#
 	PuppeteerSharp.dll!PuppeteerSharp.Browser.CloseAsync() Line 196	C#
 	PuppeteerSharp.dll!PuppeteerSharp.Browser.Connection_Disconnected(object sender, System.EventArgs e) Line 477	C#
 	PuppeteerSharp.dll!PuppeteerSharp.Connection.Close(string closeReason) Line 200	C#
 	PuppeteerSharp.dll!PuppeteerSharp.ChromeTargetManager.OnMessageReceived(object sender, PuppeteerSharp.MessageEventArgs e) Line 172	C#
 	PuppeteerSharp.dll!PuppeteerSharp.Connection.ProcessIncomingMessage(PuppeteerSharp.Messaging.ConnectionResponse obj) Line 323	C#
 	PuppeteerSharp.dll!PuppeteerSharp.Connection.ProcessMessage(PuppeteerSharp.Transport.MessageReceivedEventArgs e) Line 269	C#
 	PuppeteerSharp.dll!PuppeteerSharp.Connection.Transport_MessageReceived.AnonymousMethod__0() Line 244	C#
 	PuppeteerSharp.dll!PuppeteerSharp.Helpers.TaskQueue.Enqueue(System.Func<System.Threading.Tasks.Task> taskGenerator) Line 55	C#
 	PuppeteerSharp.dll!PuppeteerSharp.Connection.Transport_MessageReceived(object sender, PuppeteerSharp.Transport.MessageReceivedEventArgs e) Line 244	C#
 	PuppeteerSharp.dll!PuppeteerSharp.Transport.WebSocketTransport.GetResponseAsync(System.Threading.CancellationToken cancellationToken) Line 190	C#
```


My thoughts on this:
* Logger could be initialized sooner in Browser.cs. This seems like a low-risk change.
* Sending the message "Target.setDiscoverTargets" before the `Browser` and `ChromeTargetManager` objects are fully constructed is risky. Maybe consider moving the send to a separate method, that must be called after the constructor?
* Any exception caught by OnMessageReceived may lead to this deadlock. How can we avoid the deadlock in this case? Maybe use a ThreadLocal to signal that the semaphore is held on the same thread?

leonardo-fernandes:
Another race condition.

1. The code in `WaitForTargetAsync` and `TargetManager_TargetAvailable` can run in parallel in two distinct threads.
2. The thread executing `WaitForTargetAsync` checks for the target in `Targets()`. If it's not found, then a `targetCompletionSource` is created to await for the creation of the target.
3. The code checks again in `Targets()` for some reason.
4. At this point (after the second check against `Targets()`, there's a chance that a thread switch continues executing `TargetManager_TargetAvailable`.
5. The code in `TargetManager_TargetAvailable` calls the `TargetCreated` event. However, this call happens before the event handlers in the first thread had been added.
6. When the thread switches back to `WaitForTargetAsync`, the `TargetCreated` event handler is added, but it's too late and it won't be invoked. The `targetCompletionSource` will never resolve, and the code will terminate with a timeout.

Thoughts:
* I think we should re-check `Targets()` after attaching the event handlers. If the `TargetCreated` event is invoked before the event handlers are attached, we will find them by re-checking `Targets()`. If the event is invoked after, then the `targetCompletionSource` will do its job.

kblok:
> Do you mind if I handle all those cases in a separate PR?

Sure!

jnyrup:
> As I'm feeling more confident, and seeing that it improves the test results, I've included the changes for not awaiting OnAttachedToTarget for your consideration. I've also included a missing return, see [puppeteer/puppeteer@f342c26/packages/puppeteer-core/src/common/ChromeTargetManager.ts#L335](https://github.com/puppeteer/puppeteer/blob/f342c26d953e51bf262123e2e97276eff8ee2841/packages/puppeteer-core/src/common/ChromeTargetManager.ts#L335), and without which some tests would fail randomly.

Great finding 🚀  

Consider extracting this a separate PR.
That'll make it easier to track how AppVeyor behaves.

lib/PuppeteerSharp/Browser.cs
```diff
@@ -220,17 +222,15 @@ void TargetHandler(object sender, TargetChangedArgs e)
...
-                }
-
                 TargetCreated += TargetHandler;
                 TargetChanged += TargetHandler;
 
+                existingTarget = Targets().FirstOrDefault(predicate);
+                if (existingTarget != null)
+                {
+                    return existingTarget;
+                }
```
jnyrup, line 232: This [commit](https://github.com/hardkoded/puppeteer-sharp/pull/2051/commits/42cdc6550658dda8b807bdab89b72da36c60cb79) added enumeration of `Targets()` below attaching the event handlers - not sure if that was intended since the commit message is "Update chromium version".

[Puppeteer](https://github.com/puppeteer/puppeteer/blob/9abd48a062a4a30fb93d0b555f2fa03d3dc410f3/packages/puppeteer-core/src/common/Browser.ts#L499) only enumerates `targets` after attaching the event handlers.

leonardo-fernandes: The first check for `Targets()` on line 207 is indeed unnecessary. I'll remove that.

lib/PuppeteerSharp/Browser.cs
```diff
@@ -62,6 +62,8 @@ public class Browser : IBrowser
                 contextId => contextId,
                 contextId => new BrowserContext(Connection, this, contextId)));
 
+            _logger = Connection.LoggerFactory.CreateLogger<Browser>();
```
jnyrup, line 65: FYI, I moved this to line 49 in #2218

kblok:
@jnyrup I'll wait for your approval here. AppVeyor seems to go in the green direction.

lib/PuppeteerSharp/ChromeTargetManager.cs
```diff
@@ -35,8 +38,21 @@ internal class ChromeTargetManager : ITargetManager
...
+
+        public event EventHandler<TargetChangedArgs> TargetChanged;
 
+        public event EventHandler<TargetChangedArgs> TargetDiscovered;
+
+        public ConcurrentDictionary<string, Target> GetAvailableTargets() => _attachedTargetsByTargetId;
+
+        public async Task InitializeAsync()
+        {
             _ = _connection.SendAsync("Target.setDiscoverTargets", new TargetSetDiscoverTargetsRequest
```
jnyrup, line 56: Is there a benefit of moving the call to `Target.setDiscoverTargets` in `InitializeAsync` instead of having it in the constructor as it is in [puppeteer](https://github.com/puppeteer/puppeteer/blob/793a37183750b85ee71c8a6ac9d24f8e4ac0dff7/packages/puppeteer-core/src/common/ChromeTargetManager.ts#L103-L109)?

jnyrup (commented):
The changes look well-argued to me.
I just have some questions to ensure your fixes don't regress anything else.

lib/PuppeteerSharp/Browser.cs
```diff
@@ -220,17 +214,15 @@ void TargetHandler(object sender, TargetChangedArgs e)
...
-                }
-
                 TargetCreated += TargetHandler;
                 TargetChanged += TargetHandler;
 
+                var existingTarget = Targets().FirstOrDefault(predicate);
+                if (existingTarget != null)
+                {
+                    return existingTarget;
+                }
```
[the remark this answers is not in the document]
leonardo-fernandes: The issue is that in .NET two threads could execute code in parallel. If we check for `Targets()` before adding the event handlers, there's a possibility that it returns empty and then, in another thread, a `Target` is created, even before the event handlers are attached in the original thread. In this scenario, the creation won't be observed by the event listeners, and `targetCompletionSource` will never resolve. It's an improbable race condition, which is why it is not easy to reproduce and was undetected so far, but if you run enough tests it eventually happens.

In JavaScript this scenario doesn't happen, because it is single-threaded, so there's a guarantee that `Targets()` won't change between the enumeration and the event handlers being added.

lib/PuppeteerSharp/ChromeTargetManager.cs
```diff
@@ -125,14 +137,16 @@ private void StoreExistingTargetsForInit()
...
+        private void OnMessageReceived(object sender, MessageEventArgs e)
         {
             try
             {
                 switch (e.MessageID)
                 {
                     case "Target.attachedToTarget":
-                        await OnAttachedToTarget(sender, e.MessageData.ToObject<TargetAttachedToTargetResponse>(true)).ConfigureAwait(false);
+#pragma warning disable CS4014
+                        OnAttachedToTarget(sender, e.MessageData.ToObject<TargetAttachedToTargetResponse>(true)).ConfigureAwait(false);
```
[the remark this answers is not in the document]
leonardo-fernandes: It doesn't change the handling of exceptions before the first `await` in `OnAttachedToTarget`. After that point, any exceptions would be unhandled and wouldn't be logged, and the connection wouldn't be closed.

I will run some tests forcing exceptions in that code, and improve this handling.

lib/PuppeteerSharp/ChromeTargetManager.cs
leonardo-fernandes: I didn't fully understand this, so I'll have to run more tests to get a better grip at it. But the test PuppeteerConnectTests.ShouldSupportTargetFilter was failing randomly. This test connects to a browser that already has 3 tabs open.

What I found was that the messages of the targets were being processed concurrently, even before the event handlers in `AttachAsync` were added. Again, ultimately this is due to the multi-threaded .NET vs single-threaded JS difference, as in JS the constructor and the event handlers would be added without any concurrent code.

I will get back to you with a more thorough explanation.

lib/PuppeteerSharp/Browser.cs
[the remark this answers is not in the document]
jnyrup: Roger, I hadn't fully understood your first explanation.

lib/PuppeteerSharp/ChromeTargetManager.cs
kblok, line 148: Can we do `_ = OnAttachedToTarget` and remove the `#pragma`?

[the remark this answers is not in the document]
kblok: If we don't await `OnAttachedToTarget` we must wrap all the code inside `OnAttachedToTarget` in a try-catch and log the error.

lib/PuppeteerSharp/ChromeTargetManager.cs
jnyrup: Another idea, if we move `Target.setDiscoverTargets` to `InitializeAsync` anyways, can we use modern async/await syntax and avoid `_targetDiscoveryCompletionSource` or I'm a missing some concurrency again?

```cs
public async Task InitializeAsync()
{
    var setDiscoverTargetsTask = _connection.SendAsync("Target.setDiscoverTargets", new TargetSetDiscoverTargetsRequest
    {
        Discover = true,
        Filter = new[]
        {
            new TargetSetDiscoverTargetsRequest.DiscoverFilter()
            {
                Type = "tab",
                Exclude = true,
            },
            new TargetSetDiscoverTargetsRequest.DiscoverFilter(),
        },
    });

    await _connection.SendAsync("Target.setAutoAttach", new TargetSetAutoAttachRequest()
    {
        WaitForDebuggerOnStart = true,
        Flatten = true,
        AutoAttach = true,
    }).ConfigureAwait(false);

    try
    {
        await setDiscoverTargetsTask.ConfigureAwait(false);
        StoreExistingTargetsForInit();
    }
    catch (Exception ex)
    {
        _logger.LogError(ex, "Target.setDiscoverTargets failed");
    }

    FinishInitializationIfReady();

    await _initializeCompletionSource.Task.ConfigureAwait(false);
}
```

lib/PuppeteerSharp/ChromeTargetManager.cs
```diff
@@ -264,6 +280,7 @@ private async Task OnAttachedToTarget(object sender, TargetAttachedToTargetRespo
                 _ignoredTargets.Add(targetInfo.TargetId);
                 FinishInitializationIfReady(targetInfo.TargetId);
                 await silentDetach().ConfigureAwait(false);
+                return;
```
jnyrup, line 283: This seems to fix `PuppeteerConnectTests.ShouldSupportTargetFilter` when running all `PuppeteerConnectTests` in parallel 🚀

lib/PuppeteerSharp/ChromeTargetManager.cs
leonardo-fernandes: @jnyrup I have reverted these changes and couldn't reproduce any failure in 100 executions of PuppeteerConnectTests.ShouldSupportTargetFilter. Let's see how it behaves in AppVeyor.

lib/PuppeteerSharp/ChromeTargetManager.cs
```diff
@@ -137,29 +137,12 @@ private void StoreExistingTargetsForInit()
...
-                            try
-                            {
-                                await OnAttachedToTarget(sender, e.MessageData.ToObject<TargetAttachedToTargetResponse>(true)).ConfigureAwait(false);
-                            }
-                            catch (Exception ex)
-                            {
-                                handleException(ex);
-                            }
-                        });
+                        _ = OnAttachedToTargetHandlingExceptions(sender, e.MessageID, e.MessageData.ToObject<TargetAttachedToTargetResponse>(true)).ConfigureAwait(false);
```
kblok, line 145: You don't need `ConfigureAwait(false);` if you are dropping the `Task`.

```suggestion
                        _ = OnAttachedToTargetHandlingExceptions(sender, e.MessageID, e.MessageData.ToObject<TargetAttachedToTargetResponse>(true));
```

lib/PuppeteerSharp/ChromeTargetManager.cs
```diff
@@ -354,6 +337,25 @@ private async Task OnAttachedToTarget(object sender, TargetAttachedToTargetRespo
             }
         }
 
+        private async Task OnAttachedToTargetHandlingExceptions(object sender, string messageId, TargetAttachedToTargetResponse e)
```
kblok, line 340:
```suggestion
        private async Task OnAttachedToTargetHandlingExceptionsAsync(object sender, string messageId, TargetAttachedToTargetResponse e)
```

lib/PuppeteerSharp/ChromeTargetManager.cs
leonardo-fernandes: Turns out awaiting for `_targetDiscoveryCompletionSource` is needed before calling `FinishInitializationIfReady()`, otherwise `FinishInitializationIfReady` could execute before any targets have been added to the `_targetsIdsForInit` collection and the initialization would be considered completed.

leonardo-fernandes:
Anyone has any idea why FrameWaitForFunctionTests.ShouldPollOnMutation is failing in ubuntu? I can't find any explanation for that failure, unless something is really broken with `Page.WaitForFunctionAsync()` and it's not waiting for the poller to return a result.

I tried reproducing it on an ubuntu VM and was not able to get it to fail.

lib/PuppeteerSharp/Page.cs
```diff
@@ -667,25 +667,29 @@ public async Task<byte[]> ScreenshotDataAsync(ScreenshotOptions options)
...
+                await Client.Connection.SendAsync("Target.closeTarget", new TargetCloseTargetRequest
                 {
                     TargetId = Target.TargetId,
-                }).ContinueWith(task => ((Target)Target).CloseTask, TaskScheduler.Default);
-            }
+                }).ConfigureAwait(false);
 
-            _logger.LogWarning("Protocol error: Connection closed. Most likely the page has been closed.");
-            return Task.CompletedTask;
+                await ((Target)Target).CloseTask.ConfigureAwait(false);
```
jnyrup, line 691: I'm noticing that we now only execute `((Target)Target).CloseTask` if `Client.Connection.SendAsync("Target.closeTarget"` completes _successfully_.
Should we still execute `((Target)Target).CloseTask` if `Client.Connection.SendAsync("Target.closeTarget"` throws an exception?

lib/PuppeteerSharp/ChromeTargetManager.cs
jnyrup: I'm fine with how it's done now 👍 

... but if we moved `_connection.SendAsync("Target.setDiscoverTargets"` to `InitializeAsync` again and awaited it, like in my example above, I'm failing to see how we can reach `FinishInitializationIfReady` before `setDiscoverTargetsTask` has completed, which should remove the need for `_targetDiscoveryCompletionSource`.

jnyrup (approved):

lib/PuppeteerSharp/ChromeTargetManager.cs
```diff
@@ -182,9 +194,10 @@ private void OnTargetCreated(TargetCreatedResponse e)
             }
         }
 
-        private void OnTargetDestroyed(TargetDestroyedResponse e)
+        private async void OnTargetDestroyed(TargetDestroyedResponse e)
         {
             _discoveredTargetsByTargetId.TryRemove(e.TargetId, out var targetInfo);
+            await _targetDiscoveryCompletionSource.Task.ConfigureAwait(false);
```
kblok, line 200: Although we are trying to enforce this. What do you think about using `WithTimeout`, and throwing an exception? If we fail to complete this task somehow, the code will get stuck here.

lib/PuppeteerSharp/ChromeTargetManager.cs
```diff
@@ -245,6 +258,7 @@ private async Task OnAttachedToTarget(object sender, TargetAttachedToTargetRespo
             if (targetInfo.Type == TargetType.ServiceWorker &&
                 _connection.IsAutoAttached(targetInfo.TargetId))
             {
+                await _targetDiscoveryCompletionSource.Task.ConfigureAwait(false);
```
kblok, line 261: Same here.

lib/PuppeteerSharp/ChromeTargetManager.cs
```diff
@@ -261,8 +275,10 @@ private async Task OnAttachedToTarget(object sender, TargetAttachedToTargetRespo
             if (_targetFilterFunc?.Invoke(targetInfo) == false)
             {
                 _ignoredTargets.Add(targetInfo.TargetId);
+                await _targetDiscoveryCompletionSource.Task.ConfigureAwait(false);
```
kblok, line 278: and here :)

lib/PuppeteerSharp/Page.cs
kblok: @jnyrup this matches [upstream](https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer-core/src/common/Page.ts#L1506)

lib/PuppeteerSharp/ChromeTargetManager.cs
leonardo-fernandes: Note that `_targetDiscoveryCompletionSource` will be set if and only if the message `Target.setDiscoverTargets` completes. I can't think of any scenario where it wouldn't complete.

There's no code that uses a timeout value in ChromeTargetManager. Where should I derive the timeout value from?

kblok: Let's use `Puppeteer.DefaultTimeout`. It's quite a long timeout. We want to prevent the code from hanging.

lib/PuppeteerSharp/ChromeTargetManager.cs
leonardo-fernandes: @jnyrup had a closer look at your code above.

The change seems fine, and it simplifies a few things - for example it uses a try/catch which is more natural than checking `t.IsFaulted`. It's also closer to how FirefoxTargetManager works.

However, the `_targetDiscoveryCompletionSource` is still needed, to signal other threads that may be processing a message and executing `OnAttachedToTarget`.

It also has the downside of diverging from upstream.

I can commit these changes if you agree it's a better solution.

jnyrup: Ahh, it's those events messing with my brain again...
I now (again) see why `_targetDiscoveryCompletionSource` still is necessary.

I'll let kblok decide whether to call `_connection.SendAsync("Target.setDiscoverTargets"` in the constructor of `InitializeAsync`.

jnyrup:
@leonardo-fernandes Since you seem to excel at this concurrency stuff, you might also want to give `FrameManager` a similar look, to see if it suffers a similar race-condition as the one you found in `ChromeTargetManager`.

lib/PuppeteerSharp/ChromeTargetManager.cs
```diff
@@ -313,7 +329,7 @@ private async Task OnAttachedToTarget(object sender, TargetAttachedToTargetRespo
                 }
             }
 
-            await _targetDiscoveryCompletionSource.Task.ConfigureAwait(false);
+            await EnsureTargetsIdsForInit().ConfigureAwait(false);
```
kblok: This is way better than looking for the variable every time. Thanks! Let's wait for the green!

kblok:
@leonardo-fernandes @jnyrup I consider this PR approved. Let's create a follow-up PR if you find more things to improve.

Side Note: You can't image, well, I bet you @jnyrup can imagine what it means for a community project to have more developers looking at this code. Thank you!

lib/PuppeteerSharp/ChromeTargetManager.cs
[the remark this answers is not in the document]
jnyrup: Did this one slip through?
It would seem more safe to do it like `OnAttachedToTargetHandlingExceptionsAsync` to avoid any exceptions being thrown in the wild.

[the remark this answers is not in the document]
jnyrup: E.g. try running the code below.
Calling `AsyncVoid.A()` I get
image

While when calling `AsyncTask.A()` I get
image


```cs
using System;
using System.Threading.Tasks;

TaskScheduler.UnobservedTaskException += TaskScheduler_UnobservedTaskException;

try
{
    Console.WriteLine("Main try");
    AsyncVoid.A();

    Console.WriteLine("---");
    await Task.Delay(1000);
    Console.WriteLine("---");

    AsyncTask.A();

    Console.WriteLine("---");
    await Task.Delay(1000);
}
catch (Exception e)
{
    Console.WriteLine("Main catch");
    Console.WriteLine(e);
}
finally
{
    Console.WriteLine("Main finally");
}

static void TaskScheduler_UnobservedTaskException(object sender, UnobservedTaskExceptionEventArgs e)
{
    Console.WriteLine("Unobserved exception!");
    Console.WriteLine(e.Exception);
}

static class AsyncVoid
{
    public static void A()
    {
        try
        {
            Console.WriteLine("AsyncVoid sync try");
            B();
        }
        catch (Exception e)
        {
            Console.WriteLine("AsyncVoid sync catch");
            Console.WriteLine(e);
        }
        finally
        {
            Console.WriteLine("AsyncVoid sync finally");
        }
    }

    private static async void B()
    {
        try
        {
            Console.WriteLine("AsyncVoid async try");
            await Task.Delay(100);
            throw new TimeoutException();
        }
        catch
        {
            Console.WriteLine("AsyncVoid async catch");
            throw;
        }
    }
}

static class AsyncTask
{
    public static void A()
    {
        try
        {
            Console.WriteLine("AsyncTask sync try");
            _ = B();
        }
        catch (Exception e)
        {
            Console.WriteLine("AsyncTask sync catch");
            Console.WriteLine(e);
        }
        finally
        {
            Console.WriteLine("AsyncTask sync finally");
        }
    }

    private static async Task B()
    {
        try
        {
            Console.WriteLine("AsyncTask async try");
            await Task.Delay(100);
            throw new TimeoutException();
        }
        catch
        {
            Console.WriteLine("AsyncTask async catch");
            throw;
        }
    }
}
```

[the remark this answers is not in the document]
kblok: @jnyrup `OnAttachedToTargetHandlingExceptionsAsync` has a try/catch. It won't throw an exception.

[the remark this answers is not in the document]
jnyrup: Sorry, meant `OnTargetDestroyed`.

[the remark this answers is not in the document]
kblok: Yes. We do need a try/catch there @jnyrup :)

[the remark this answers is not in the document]
jnyrup: Already done
https://github.com/hardkoded/puppeteer-sharp/pull/2289/commits/f5ac62fee3c0f6d5aec0227e70d71b2ef9bcc346?w=1

PR #2220, opened 2023-06-07
Ensure members are initialized before attaching event handlers

jnyrup:
Inspired by #2218 I found some more places, where we attach event handlers, before initializing instance members used by the event handlers.

Note, I haven't investigated if these scenarios can happen in practice, so this is a better-safe-than-sorry change.

```
Connection.ctor
    attaching Transport_Closed
        Close(string closeReason)
            using _sessions
            using _callbacks
            using MessageQueue

    initialize _callbacks
    initialize _sessions
    initialize MessageQueue
```    

```    
Worker.ctor
    attaching OnMessageReceived
        OnExecutionContextCreated(RuntimeExecutionContextCreatedResponse e)
            using _executionContextCallback
    initialize _executionContextCallback
```

### DIFF
--- a/lib/PuppeteerSharp/Connection.cs
+++ b/lib/PuppeteerSharp/Connection.cs
@@ -37,12 +37,13 @@ namespace PuppeteerSharp
 
             _logger = LoggerFactory.CreateLogger<Connection>();
 
-            Transport.MessageReceived += Transport_MessageReceived;
-            Transport.Closed += Transport_Closed;
             _callbacks = new ConcurrentDictionary<int, MessageTask>();
             _sessions = new ConcurrentDictionary<string, CDPSession>();
             MessageQueue = new AsyncMessageQueue(enqueueAsyncMessages, _logger);
             _asyncSessions = new AsyncDictionaryHelper<string, CDPSession>(_sessions, "Session {0} not found");
+
+            Transport.MessageReceived += Transport_MessageReceived;
+            Transport.Closed += Transport_Closed;
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


lib/PuppeteerSharp/Connection.cs
```diff
@@ -37,12 +37,13 @@ internal Connection(string url, int delay, bool enqueueAsyncMessages, IConnectio
 
             _logger = LoggerFactory.CreateLogger<Connection>();
 
-            Transport.MessageReceived += Transport_MessageReceived;
-            Transport.Closed += Transport_Closed;
             _callbacks = new ConcurrentDictionary<int, MessageTask>();
             _sessions = new ConcurrentDictionary<string, CDPSession>();
             MessageQueue = new AsyncMessageQueue(enqueueAsyncMessages, _logger);
             _asyncSessions = new AsyncDictionaryHelper<string, CDPSession>(_sessions, "Session {0} not found");
```
kblok, line 43: How about also moving these to the variable declaration?

lib/PuppeteerSharp/Worker.cs
```diff
@@ -45,9 +45,8 @@ public class Worker
             Url = url;
             _consoleAPICalled = consoleAPICalled;
             _exceptionThrown = exceptionThrown;
-            _client.MessageReceived += OnMessageReceived;
-
             _executionContextCallback = new TaskCompletionSource<ExecutionContext>(TaskCreationOptions.RunContinuationsAsynchronously);
```
kblok, line 48: same here

PR #2225, opened 2023-06-12
Fix FrameWaitForFunctionTests.ShouldPollOnMutation

leonardo-fernandes:
The additional `EvaluateExpressionAsync` ensures that the poller has started before the test proceeds.

I also took the liberty to refactor and simplify the `EvaluateExpressionAsync` to avoid any changes to DOM. This seems to have the same effect and wouldn't interfere with the mutation observer.


lib/PuppeteerSharp.Tests/WaitTaskTests/FrameWaitForFunctionTests.cs
```diff
@@ -1,17 +1,31 @@
...
 {
     [Collection(TestConstants.TestFixtureCollectionName)]
-    public class FrameWaitForFunctionTests : PuppeteerPageBaseTest
+    public sealed class FrameWaitForFunctionTests : PuppeteerPageBaseTest, IDisposable
     {
+        private ConnectionTransportInterceptor _connectionTransportInterceptor;
+
         public FrameWaitForFunctionTests(ITestOutputHelper output) : base(output)
         {
+            DefaultOptions = TestConstants.DefaultBrowserOptions();
```
kblok, line 18: This is clever. Can would document what we're trying to accomplish here?

lib/PuppeteerSharp.Tests/WaitTaskTests/FrameWaitForFunctionTests.cs
```diff
@@ -216,5 +232,59 @@ public async Task ShouldSurviveNavigations()
...
+
+        private TaskCompletionSource<bool> ListenForStartPolling()
+        {
+            TaskCompletionSource<bool> startedPolling = new TaskCompletionSource<bool>();
+
+            // Wait for function will release the execution faster than in node.
+            // We intercept the poller.start() call to prevent tests from continuing before the polling has started.
+            _connectionTransportInterceptor.MessageSent += (_, message) =>
+            {
+                if (message.Contains("poller => poller.start()"))
```
kblok, line 244: Let's add a comment where we are executing the `poller => poller.start()`. Mentioning that if we change the code there we need to update this test.

lib/PuppeteerSharp.Tests/WaitTaskTests/FrameWaitForFunctionTests.cs
```diff
@@ -216,5 +232,59 @@ public async Task ShouldSurviveNavigations()
...
+                {
+                    startedPolling.SetResult(true);
+                }
+            };
+
+            return startedPolling;
+        }
+
+        private sealed class ConnectionTransportInterceptor : IConnectionTransport
```
kblok, line 253: Let's move this to its own file and make it public. This is a great idea.

lib/PuppeteerSharp.Tests/WaitTaskTests/FrameWaitForFunctionTests.cs
```diff
@@ -216,5 +232,59 @@ public async Task ShouldSurviveNavigations()
             await Page.EvaluateFunctionAsync("() => window.__done = true");
             await watchdog;
         }
+
+        private TaskCompletionSource<bool> ListenForStartPolling()
```
kblok, line 236: This can return only the task and call this `WaitForStartPollingAsync`

lib/PuppeteerSharp.Tests/WaitTaskTests/FrameWaitForFunctionTests.cs
leonardo-fernandes: Should I create a new folder, or leave it in WaitTaskTests?

kblok: Let's move it to the root folder by now. There are many classes there that could be moved to an `Utils` directory.

PR #2230, opened 2023-06-12
Use xpath query handler

### LINKED ISSUE #2085: Add missing Xpath Selector engine

kblok:
closes #2085

### DIFF
--- a/lib/PuppeteerSharp/CustomQueriesManager.cs
+++ b/lib/PuppeteerSharp/CustomQueriesManager.cs
@@ -10,6 +10,8 @@ namespace PuppeteerSharp
 {
     internal class CustomQueriesManager
     {
+        private static readonly string[] CustomQuerySeparators = new[] { "=", "/" };
+        private readonly Dictionary<string, PuppeteerQueryHandler> _internalQueryHandlers;
         private readonly Dictionary<string, PuppeteerQueryHandler> _queryHandlers = new();
         private readonly PuppeteerQueryHandler _pierceHandler = CreatePuppeteerQueryHandler(new CustomQueryHandler
         {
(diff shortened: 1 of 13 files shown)


lib/PuppeteerSharp/CustomQueriesManager.cs
```diff
@@ -155,21 +190,23 @@ internal void RegisterCustomQueryHandler(string name, CustomQueryHandler queryHa
...
-            {
-                return (selector, _defaultHandler);
-            }
-
-            var name = customQueryHandlerMatch.Groups["query"].Value;
-            var updatedSelector = customQueryHandlerMatch.Groups["selector"].Value;
+            var handlers = _internalQueryHandlers.Concat(_queryHandlers);
 
-            if (!_queryHandlers.TryGetValue(name, out var queryHandler))
+            foreach (var kv in handlers)
```
jnyrup, line 195: Use 4 spaces instead of 2 🧹

lib/PuppeteerSharp/CustomQueriesManager.cs
```diff
@@ -10,6 +10,8 @@ namespace PuppeteerSharp
 {
     internal class CustomQueriesManager
     {
+        private static readonly string[] CustomQuerySeparators = new[] { "=", "/" };
+        private readonly Dictionary<string, PuppeteerQueryHandler> _internalQueryHandlers;
```
jnyrup, line 14: Suggestion:
Besides `_defaultHandler`, we don't need to store the internal handlers in separate fields as they are only used in the constructor to initialize `_internalQueryHandlers`.

By changing them from fields to private static creator methods, we can move initialization of `_internalQueryHandlers` to declaration and remove the then empty public constructor.

```cs
private readonly Dictionary<string, PuppeteerQueryHandler> _internalQueryHandlers = new()
{
    ["aria"] = AriaHandler(),
    ["pierce"] = PierceHandler(),
    ["text"] = TextQueryHandler(),
    ["xpath"] = XpathHandler(),
};
```

lib/PuppeteerSharp/CustomQueriesManager.cs
kblok: ohhh just merged! Let's create a follow up PR

jnyrup: I'll create a PR right away

kblok: I have one ready to go :)

PR #2231, opened 2023-06-12
Single lookups into concurrent dictionary

jnyrup:
Found some more places where we can use built-in methods on `ConcurrentDictionary`

### DIFF
--- a/lib/PuppeteerSharp/ChromeTargetManager.cs
+++ b/lib/PuppeteerSharp/ChromeTargetManager.cs
@@ -89,18 +89,8 @@ namespace PuppeteerSharp
 
         public void AddTargetInterceptor(CDPSession session, TargetInterceptor interceptor)
         {
-            lock (_targetInterceptors)
-            {
-                _targetInterceptors.TryGetValue(session, out var interceptors);
-
-                if (interceptors == null)
-                {
-                    interceptors = new List<TargetInterceptor>();
-                    _targetInterceptors.TryAdd(session, interceptors);
-                }
-
-                interceptors.Add(interceptor);
-            }
+            var interceptors = _targetInterceptors.GetOrAdd(session, static _ => new());
+            interceptors.Add(interceptor);
         }
 
         public void RemoveTargetInterceptor(CDPSession session, TargetInterceptor interceptor)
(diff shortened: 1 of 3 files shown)


lib/PuppeteerSharp/ChromeTargetManager.cs
```diff
@@ -89,18 +89,8 @@ await _connection.SendAsync("Target.setAutoAttach", new TargetSetAutoAttachReque
...
-                if (interceptors == null)
-                {
-                    interceptors = new List<TargetInterceptor>();
-                    _targetInterceptors.TryAdd(session, interceptors);
-                }
-
-                interceptors.Add(interceptor);
-            }
+            var interceptors = _targetInterceptors.GetOrAdd(session, static _ => new());
+            interceptors.Add(interceptor);
```
jnyrup, line 93: @kblok Do you recall if the lock in `AddTargetInterceptor` intended to lock both the usage of `_targetInterceptors` and adding an element to `interceptors`?

A colleague of mine pointed out that I might have introduced a race-condition here
If two threads get the same `interceptors` out of `_targetInterceptors` and concurrently calls the thread-unsafe `List<T>.Add`.

On the other hand the other usages of `interceptors` are also not locked
* `RemoveTargetInterceptor` removes an item from the list
* `OnAttachedToTarget` enumerates the list

kblok: > A colleague of mine pointed out that I might have introduced a race condition here

Is that a theory or a suspicion? We are talking about multiple threads, but I don't see how those events coming from the browser could be racy.

But yeah. The goal of the lock was to make those two actions thread-safe.
We could restore the lock. We could also turn the list into a `ConcurrentSet`.

> On the other hand the other usages of interceptors are also not locked
> * RemoveTargetInterceptor removes an item from the list

Again, I don't see how we can get a race in `RemoveTargetInterceptor`

> OnAttachedToTarget enumerates the list

I can see the need for a lock there.

jnyrup: > > A colleague of mine pointed out that I might have introduced a race condition here
> 
> Is that a theory or a suspicion? We are talking about multiple threads, but I don't see how those events coming from the browser could be racy.

Before this PR the non-thread safe `interceptors.Add(interceptor);` was inside a lock and now it isn't.
I didn't notice that slight change of semantics when I created the PR, that's why I'm bringing it to your attention.
We haven't experienced that any problems that I suspect stem from this PR, so for now it's a theoretical race-condition.
I assume that's what you meant by "theory or a suspicion"?

kblok: @jnyrup yes. I was curious whether you did got a race condition or you found it could be a problem.
I think we could make it a `ConcurrentSet` just in case.

PR #2232, opened 2023-06-12
Skip `ShouldDownloadAndExtractFirefoxLinuxBinary` on Windows

jnyrup:
The built-in Windows command `tar` delegates handling of .tar.bzip2 archives to another command `bzip2`.
Windows 10 does not come with `bzip2` pre-installed.

Tests can only be decorated with a single attribute deriving from `FactAttribute`, that's the reason I replaced `PuppeteerFact`.

Fixes https://github.com/hardkoded/puppeteer-sharp/discussions/2130


kblok (approved):

kblok:
It's a shame that it does work on AppVeyor (Windows), but not on your local. But I get the point. It's more important to run this on Linux.

PR #2236, opened 2023-06-13
Atomically get Puppeteer utilities

kblok:
It matches [stream code](https://github.com/puppeteer/puppeteer/commit/050a7b062415ebaf10bcb71c405143eacc4e5d4b)

### DIFF
--- a/lib/PuppeteerSharp/CustomQueriesManager.cs
+++ b/lib/PuppeteerSharp/CustomQueriesManager.cs
@@ -10,7 +10,7 @@ namespace PuppeteerSharp
 {
     internal class CustomQueriesManager
     {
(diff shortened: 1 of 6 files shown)


lib/PuppeteerSharp/CustomQueriesManager.cs
```diff
@@ -235,7 +235,7 @@ private static PuppeteerQueryHandler CreatePuppeteerQueryHandler(CustomQueryHand
                     var jsHandle = await element.EvaluateFunctionHandleAsync(
```
jnyrup, line 235: `handle` on line 234 and 283 are now unused

kblok: thank you!

PR #2238, opened 2023-06-14
Fix PageGotoTests

leonardo-fernandes:
I started investigating why PageGotoTests were failing sometimes with NullReferenceException (e.g. https://github.com/hardkoded/puppeteer-sharp/actions/runs/5257640661/jobs/9500723964).

I found that, when the tests fails, the Response object wasn't initialized yet in the test's thread, but another thread was concurrently processing a message that would initialize it. More investigation indicated that the timing and order of two messages would determine if the test passed/failed.
* The `Page.navigate` message is sent to initiate the navigation. This message awaits for `NewDocumentNavigationTask` from LifecycleWatcher.
* Two messages, `Network.responseReceived` and `Network.responseReceivedExtraInfo`, are received as a result of the navigation. It seems that `Network.responseReceived` is always delivered before the `Page.navigate` resolves. But the order that the two messages are delivered varies with each test execution.
* When `Network.responseReceivedExtraInfo` is delivered before `Network.responseReceived`, the code in `NetworkManager.OnResponseReceived` is able to find the `extraInfo` and proceeds to execute `EmitResponseEvent()` which sets the Response object.
* When `Network.responseReceived` is delivered before `Network.responseReceivedExtraInfo`, the execution of `NetworkManager.OnResponseReceived` does not find the `extraInfo` object and queues the response event. In this state, the `NewDocumentNavigationTask` can resolve and continue the test execution before the Response object is set.

I compared the relevant code with upstream, and found some differences that were committed in https://github.com/puppeteer/puppeteer/pull/8717. My understanding is that the `_newDocumentNavigation` flag was being set too early by a same-document navigation (somehow related to beforeunload events), and thus causing the `NewDocumentNavigationTask` to resolve before it should have. I am still investigating if that is sufficient to make sure the Response object is set.

I also found that ShouldWorkWhenNavigatingTo404 was failing consistently when executed in headful mode, and this had been fixed upstream by https://github.com/puppeteer/puppeteer/pull/9577.

### DIFF
--- a/lib/PuppeteerSharp.Tests/NavigationTests/PageGotoTests.cs
+++ b/lib/PuppeteerSharp.Tests/NavigationTests/PageGotoTests.cs
@@ -294,6 +294,36 @@ namespace PuppeteerSharp.Tests.NavigationTests
             Assert.Equal(HttpStatusCode.NotFound, response.Status);
         }
 
+        [PuppeteerTest("navigation.spec.ts", "Page.goto", "should not throw an error for a 404 response with an empty body")]
+        [PuppeteerFact]
+        public async Task ShouldNotThrowAnErrorForA404ResponseWithAnEmptyBody()
+        {
+            Server.SetRoute("/404-error", context =>
+            {
+                context.Response.StatusCode = 404;
+                return Task.CompletedTask;
+            });
+
+            var response = await Page.GoToAsync(TestConstants.ServerUrl + "/404-error");
+            Assert.False(response.Ok);
+            Assert.Equal(HttpStatusCode.NotFound, response.Status);
+        }
+
+        [PuppeteerTest("navigation.spec.ts", "Page.goto", "should not throw an error for a 500 response with an empty body")]
+        [PuppeteerFact]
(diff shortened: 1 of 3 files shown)


kblok:
> Two messages, Network.responseReceived and Network.responseReceivedExtraInfo, are received as a result of the navigation. It seems that Network.responseReceived is always delivered before the Page.navigate resolves. But the order that the two messages are delivered varies with each test execution.

Do you think that could be a problem? Do you see that in the logs? Because if that's coming out of order from the browser, puppeteer might be dealing with the same.

kblok:
> I compared the relevant code with upstream, and found some differences that were committed in https://github.com/puppeteer/puppeteer/pull/8717.

We just solved the same thing :)

leonardo-fernandes:
> Do you think that could be a problem? Do you see that in the logs? Because if that's coming out of order from the browser, puppeteer might be dealing with the same.

For reference, I forced a `Thread.Sleep()` just before the assignment of the Response object. Without the fix, the test fails with NullPointerException in about 2% of executions (the failure still depends on the order of the `Network.responseReceived` and `Network.responseReceivedExtraInfo` messages). With the fix, it didn't fail in 200 executions.

PR #2239, opened 2023-06-17
Fix flakyness of ShouldWaitForNetworkIdleToSucceedNavigation test

leonardo-fernandes:
The `responses` collection is accessed by multiple concurrent requests, so we should use a thread-safe collection. We should also be careful to clear the collection before the next element has any chance of being added to it.


kblok (approved):
Thank you!

PR #2259, opened 2023-07-21
Upgrade chrome to 1108766

### LINKED ISSUE #2258: BrowserFetcher.DefaultChromiumRevision Exhibits Hang Bug When Page Has Odd Number column-count
### Description

From what we can tell Chromium has a bug where if the `column-count` is odd for this page (and possibly others), it'll thread spin hard and take _hours_ to complete a screenshot in some cases. Updating to `1115057` fixes it.

There is a block of code that allows us to replace `column-count: 3` with `column-count: 4`, this works, replacing it with 5 does not, 6 does, 7 does not, 8 does.

_Ideally_ moving up to a newer stable release may be ideal, I haven't been able to find a logged Chroium bug regarding this, it may have been silently fixed at some point.

See application here:

https://github.com/StrangeWill/PuppeteerSharpStall/blob/master/Program.cs

### Expected behavior:

A screenshot to be taken

### Actual behavior:

A stall and CPU spin for possibly _hours_ before screenshot is taken

### Versions

PuppeteerSharp 10.1.2

Edit:

We're just running with this version for now -- but I wanted visibility on this and a discussion around updating the default to prevent `column-count` related issues for anyone else.

kblok:
closes #2258

### DIFF
--- a/lib/PuppeteerSharp.Tests/NavigationTests/FrameGoToTests.cs
+++ b/lib/PuppeteerSharp.Tests/NavigationTests/FrameGoToTests.cs
@@ -41,13 +41,16 @@ namespace PuppeteerSharp.Tests.NavigationTests
             await Page.QuerySelectorAsync("iframe").EvaluateFunctionAsync("frame => frame.remove()");
             var exception = await Assert.ThrowsAsync<NavigationException>(async () => await navigationTask);
             Assert.Contains(
-                "Navigating frame was detached",
                 new[]
+                {
+                    exception.Message
+                },
+                message => new[]
                 {
                     "Navigating frame was detached",
                     "Error: NS_BINDING_ABORTED",
-                    "net::ERR_ABORTED",
-                });
+                    "net::ERR_ABORTED"
+                }.Any(error => message.Contains(error)));
         }
 
         [PuppeteerTest("navigation.spec.ts", "Frame.goto", "should return matching responses")]
--- a/lib/PuppeteerSharp.Tests/NavigationTests/FrameGoToTests.cs
+++ b/lib/PuppeteerSharp.Tests/NavigationTests/FrameGoToTests.cs
@@ -1,4 +1,3 @@
-using System.Collections.Generic;
 using System.Linq;
 using System.Net;
 using System.Threading.Tasks;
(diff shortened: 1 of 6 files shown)


lib/PuppeteerSharp.Tests/NavigationTests/FrameGoToTests.cs
```diff
@@ -41,7 +40,14 @@ public async Task ShouldRejectWhenFrameDetaches()
...
             var exception = await Assert.ThrowsAsync<NavigationException>(async () => await navigationTask);
-            Assert.Equal("Navigating frame was detached", exception.Message);
+            Assert.Contains(
+                "Navigating frame was detached",
+                new[]
+                {
+                    "Navigating frame was detached",
+                    "Error: NS_BINDING_ABORTED",
+                    "net::ERR_ABORTED",
+                });
```
jnyrup, line 50: Is this change intentional? The assertion is now always true as both subject and expectation are fixed values.

kblok: This matches the change [upstream](https://github.com/puppeteer/puppeteer/blob/d0c5b8e08905f3802705a1a90d7cc8fa04bc82db/test/src/navigation.spec.ts#L808) 🤷

jnyrup: No :)
The upstream asserts that `error.message` contains at least one of the three elements.
This change asserts that either of the three elements is `"Navigating frame was detached"`.

jnyrup: This is the best I can come up with xUnit
```cs
Assert.Contains(new[] { exception.Message }, message => new[]
{
    "Navigating frame was detached",
    "Error: NS_BINDING_ABORTED",
    "net::ERR_ABORTED"
}.Any(error => message.Contains(error)));
```

jnyrup: With FluentAssertions it can be written as
```cs
exception.Message.Should().ContainAny(
    "Navigating frame was detached",
    "Error: NS_BINDING_ABORTED",
    "net::ERR_ABORTED"
);
```

kblok: Oops you're so right 😅
The test passes right? 😅

PR #2288, opened 2023-08-23
Fix URL for Chromium ResolveBuildIdAsync()

mmalvik:
There is an issue where the URL has an extra dollar sign (`$`) in the string interpolation, breaking the URL.

This results in a URL like this, giving 404 Not Found
- https://storage.googleapis.com/chromium-browser-snapshots/$Mac_Arm/LAST_CHANGE

Removing the extra dollar sign fixes the issue, giving a correct URL:
- https://storage.googleapis.com/chromium-browser-snapshots/Mac_Arm/LAST_CHANGE


kblok:
Thank you @mmalvik. Do you want to update all the versions to 11.0.2 in the csproj so we can ship it as soon as it gets merged?

mmalvik:
Just updated to `11.0.2` in the cspoj now :+1:

PR #2296, opened 2023-08-30
Remove unused code

kblok:
Added `Microsoft.CodeAnalysis.CSharp.CodeStyle` to get some extra validations on build time.


lib/PuppeteerSharp/WaitTask.cs
```diff
@@ -79,6 +79,13 @@ public void Dispose()
                 return;
             }
 
+            if (_timeoutTimer?.Status == TaskStatus.RanToCompletion ||
+                _timeoutTimer?.Status == TaskStatus.Faulted ||
+                _timeoutTimer?.Status == TaskStatus.Canceled)
+            {
+                _timeoutTimer.Dispose();
+            }
```
jnyrup, line 87:
```suggestion
            if (_timeoutTimer is { Status: TaskStatus.RanToCompletion or TaskStatus.Faulted or TaskStatus.Canceled } timeoutTimer)
            {
                timeoutTimer.Dispose();
            }
```

kblok: Fancy!

PR #2297, opened 2023-08-30
Fix memory leaks with `AsyncDictionaryHelper`

jnyrup:
To ensure that entries from `_pendingRequests` are removed when removing a key from `_dictionary`.

See [discussion](https://github.com/hardkoded/puppeteer-sharp/discussions/2283) for more details.

### DIFF
--- a/lib/PuppeteerSharp/Helpers/AsyncDictionaryHelper.cs
+++ b/lib/PuppeteerSharp/Helpers/AsyncDictionaryHelper.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Concurrent;
+using System.Collections.Generic;
 using System.Globalization;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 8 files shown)


kblok (approved):
I like it. Let's see if GH actions has the same opinion.

lib/PuppeteerSharp/Helpers/AsyncDictionaryHelper.cs
```diff
@@ -58,5 +58,27 @@ internal void AddItem(TKey key, TValue value)
                 tcs.TrySetResult(value);
             }
         }
+
+        internal bool TryRemove(TKey key, out TValue value)
+        {
+            var result = _dictionary.TryRemove(key, out value);
+            _ = _pendingRequests.TryRemove(key, out _);
```
kblok, line 65: Do we need `_ =`?

jnyrup: Strictly speaking, no. 
I usually assign to a discard variable to explicitly ignore the return value.

PR #2448, opened 2024-02-15
Fix Chrome for testing download URL

kblok:
See upstream change: https://github.com/GoogleChromeLabs/chrome-for-testing/pull/108


kblok:
@durandt, `BaseUrl` is used to build the final URL so that won't help.
I think the only way at the moment is implementing a custom download action with the `CustomFileDownloadAction` property. But that would involved downloading the file manually.

durandt:
Ok thank you for your answer.

I was wondering what would be a good way to fix this kind of issues going forward, and I see that you have already created #2460 that I guess should address that.
I would be willing to help, but do you usually wait for features to be implemented in puppeteer before porting here so that features have parity with the original repo?

PR #2637, opened 2024-05-17
Roll chrome to 125

kblok:
Unfortunately, Chrome M125 might require extra permissions on Windows (see https://pptr.dev/troubleshooting#chrome-reports-sandbox-errors-on-windows).

If you are a Windows user checking this PR, I would appreciate it if you could confirm this change in your environment.


lib/PuppeteerSharp/BrowserFetcher.cs
```diff
@@ -405,6 +405,36 @@ private async Task UnpackArchiveAsync(string archivePath, string outputPath, str
                     }
                 }
             }
+
+            if (browser == SupportedBrowser.Chrome && (GetCurrentPlatform() == Platform.Win64 || GetCurrentPlatform() == Platform.Win32))
+            {
+                if (int.TryParse(buildId.Split('.').First(), out var majorVersion) &&
+                    majorVersion > Chrome.ChromeVersionRequiringPermissionsFix)
```
kblok, line 412:
```suggestion
                    majorVersion >= Chrome.ChromeVersionRequiringPermissionsFix)
```

ezanker:
Interestingly, if I go back to PuppeteerSharp 10.1.0 and point at the Chrome V126 exe, it works fine with no errors... So, are there any change within Puppeteer that could solve this issue?

PR #2668, opened 2024-06-28
Fix: `BrowserContext.IsClosed` returns true when a context is not closed

IliaBrahinets:
The PR fixes `BrowserContext.IsClosed`. The property returns true when a context is not closed.


kblok (approved):
Thank you!

PR #2689, opened 2024-07-12
Ensure temp directory is deleted on disposal

### LINKED ISSUE #2685: PuppeteerSharp "leaks" temporary files on exit
### Description

PuppeteerSharp (or the browser process) leaves ~8mb of files in temp folder. This happens when the Browser object is disposed (async) immediately before process exit. We have a commandline utility that uses PuppeteerSharp to generate PDF documents. We found that calling it repeatedly to process a large number of PDFs would eventually fill the C: drive with temp file detritus. We discovered a workaround, which was to add a delay (`await Task.Delay(1000)`) after the browser disposal before exiting the process. With this delay, the temp files were deleted.

The symptom makes me think there is some un-awaited Task in the browser dispose code that is still pending after DisposeAsync completes. Reviewing the code, I don't see any obvious culprits.

This code example is a C# top-level program. The only non-minimal part is adjusting the temp directory to make it easier to see the files that are created by isolating them from the base temp folder. I don't know how to make this a unit test, since it requires quick process termination to illustrate the issue.

```csharp
using PuppeteerSharp;

// set temp folder env var to cause browser to write to a 
// specific location to make it easier to isolate the files
// issue repros without this, this just makes it easier to see
// and cleanup
var temp = Path.GetTempPath();
temp = Path.Combine(temp, "Puppeteer");
Directory.CreateDirectory(temp);
Environment.SetEnvironmentVariable("TMP", temp);

// repros with both msedge and chrome, I just happen to have edge installed
var opts = new LaunchOptions { ExecutablePath = "C:\\Program Files (x86)\\Microsoft\\Edge\\Application\\msedge.exe" };
await using (var b = await Puppeteer.LaunchAsync(opts))
{

} // <-- browser cleanup

// without the delay, the temp folder will have a subfolder with ~8mb
// with this delay, those files are deleted before exit
//await Task.Delay(1000); // <-- uncomment this line to demonstrate the workaround
```

### Expected behavior:

The files in temp to be deleted without requiring a delay.

### Actual behavior:

Each invocation of this program (without the delay) will leave ~8mb of files in the temp folder.

### Versions
OS: both windows 10 and windows server 2022.
PuppeteerSharp: 18.0.3 (latest), but also repros on 10.1.0.
Framework: net framework 4.8 and net 8.0.

### Additional Information

This isn't a high priority issue for us, since the workaround seems to resolve the issue. The behavior does feel like a bug though, so thought it worth reporting.

kblok:
closes #2685

### DIFF
--- a/lib/PuppeteerSharp/Browser.cs
+++ b/lib/PuppeteerSharp/Browser.cs
@@ -181,11 +181,6 @@ namespace PuppeteerSharp
             else
             {
                 await CloseAsync().ConfigureAwait(false);
-
-                if (Launcher.TempUserDataDir != null)
-                {
-                    await Launcher.TempUserDataDir.DisposeAsync().ConfigureAwait(false);
-                }
             }
 
             await ScreenshotTaskQueue.DisposeAsync().ConfigureAwait(false);
--- a/lib/PuppeteerSharp/Browser.cs
+++ b/lib/PuppeteerSharp/Browser.cs
@@ -4,7 +4,6 @@ using System.Diagnostics;
 using System.Linq;
 using System.Threading.Tasks;
 using PuppeteerSharp.Cdp;
-using PuppeteerSharp.Cdp.Messaging;
 using PuppeteerSharp.Helpers;
 
 namespace PuppeteerSharp
@@ -46,7 +45,7 @@ namespace PuppeteerSharp
         public abstract bool IsClosed { get; }
 
         /// <inheritdoc/>
-        public IBrowserContext DefaultContext { get; protected set; }
+        public IBrowserContext DefaultContext { get; protected init; }
 
         /// <inheritdoc/>
         public int DefaultWaitForTimeout { get; set; } = Puppeteer.DefaultTimeout;
@@ -59,13 +58,13 @@ namespace PuppeteerSharp
 
         internal TaskQueue ScreenshotTaskQueue { get; } = new();
 
-        internal Connection Connection { get; set; }
+        internal Connection Connection { get; init; }
 
-        internal ViewPortOptions DefaultViewport { get; set; }
+        internal ViewPortOptions DefaultViewport { get; init; }
 
-        internal LauncherBase Launcher { get; set; }
+        internal LauncherBase Launcher { get; init; }
 
-        internal Func<Target, bool> IsPageTargetFunc { get; set; }
+        internal Func<Target, bool> IsPageTargetFunc { get; init; }
 
         /// <inheritdoc/>
         public abstract Task<IPage> NewPageAsync();
@@ -175,6 +174,12 @@ namespace PuppeteerSharp
         public async ValueTask DisposeAsync()
         {
             await CloseAsync().ConfigureAwait(false);
+
+            if (Launcher.TempUserDataDir != null)
+            {
+                await Launcher.TempUserDataDir.DisposeAsync().ConfigureAwait(false);
+            }
+
             await ScreenshotTaskQueue.DisposeAsync().ConfigureAwait(false);
             GC.SuppressFinalize(this);
         }
--- a/lib/PuppeteerSharp/Browser.cs
+++ b/lib/PuppeteerSharp/Browser.cs
@@ -173,8 +173,6 @@ namespace PuppeteerSharp
         /// <returns>ValueTask.</returns>
         public async ValueTask DisposeAsync()
         {
-            await CloseAsync().ConfigureAwait(false);
-
             // On disposal, the browser doesn't get closed. It gets disconnected.
             if (Launcher == null)
             {
@@ -183,6 +181,11 @@ namespace PuppeteerSharp
             else
             {
                 await CloseAsync().ConfigureAwait(false);
+
+                if (Launcher.TempUserDataDir != null)
+                {
+                    await Launcher.TempUserDataDir.DisposeAsync().ConfigureAwait(false);
+                }
             }
 
             await ScreenshotTaskQueue.DisposeAsync().ConfigureAwait(false);
--- a/lib/PuppeteerSharp/Browser.cs
+++ b/lib/PuppeteerSharp/Browser.cs
@@ -45,7 +45,7 @@ namespace PuppeteerSharp
         public abstract bool IsClosed { get; }
 
         /// <inheritdoc/>
-        public IBrowserContext DefaultContext { get; protected init; }
+        public IBrowserContext DefaultContext { get; protected set; }
 
         /// <inheritdoc/>
         public int DefaultWaitForTimeout { get; set; } = Puppeteer.DefaultTimeout;
--- a/lib/PuppeteerSharp/Browser.cs
+++ b/lib/PuppeteerSharp/Browser.cs
@@ -179,14 +179,15 @@ namespace PuppeteerSharp
             if (Launcher == null)
             {
                 Disconnect();
-                if (Launcher.TempUserDataDir != null)
-                {
-                    await Launcher.TempUserDataDir.DisposeAsync().ConfigureAwait(false);
-                }
             }
             else
             {
                 await CloseAsync().ConfigureAwait(false);
+
+                if (Launcher.TempUserDataDir != null)
+                {
+                    await Launcher.TempUserDataDir.DisposeAsync().ConfigureAwait(false);
+                }
             }
 
             await ScreenshotTaskQueue.DisposeAsync().ConfigureAwait(false);
--- a/lib/PuppeteerSharp/Browser.cs
+++ b/lib/PuppeteerSharp/Browser.cs
@@ -183,11 +183,6 @@ namespace PuppeteerSharp
             else
             {
                 await CloseAsync().ConfigureAwait(false);
-
-                if (Launcher.TempUserDataDir != null)
-                {
-                    await Launcher.TempUserDataDir.DisposeAsync().ConfigureAwait(false);
-                }
             }
 
             await ScreenshotTaskQueue.DisposeAsync().ConfigureAwait(false);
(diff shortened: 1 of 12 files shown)


lib/PuppeteerSharp/Browser.cs
```diff
@@ -182,6 +181,11 @@ public async ValueTask DisposeAsync()
             else
             {
                 await CloseAsync().ConfigureAwait(false);
+
+                if (Launcher.TempUserDataDir != null)
+                {
+                    await Launcher.TempUserDataDir.DisposeAsync().ConfigureAwait(false);
+                }
```
jnyrup, line 188: It seems a bit odd to me why `Browser` should dispose a member of `Launcher` 🤔  
I experimented with this and how about we instead change `ExitedState` from

https://github.com/hardkoded/puppeteer-sharp/blob/eca16403b86f464f9182afae1b3bf2e4500950c1/lib/PuppeteerSharp/States/ExitedState.cs#L26-L27

to something like this?
```cs
if (p.TempUserDataDir is { } tempUserDataDir)
{
    tempUserDataDir
        .DisposeAsync()
        .AsTask()
        .ContinueWith(
            t => p.ExitCompletionSource.TrySetResult(true),
            TaskScheduler.Default);
}
else
{
    p.ExitCompletionSource.TrySetResult(true);
}
```

kblok: I agree. I don't know why I accepted that State Machine refactor that now I have to maintain many years ago.

kblok: Let me try to test that snippet. Thank you!

jnyrup:
And here's a test that fails before the fix above, but passes with it.

```cs
[Test]
public async Task DeleteTempUserDataDirWhenDisposingBrowser()
{
    var options = TestConstants.DefaultBrowserOptions();
    var launcher = new Launcher(TestConstants.LoggerFactory);
    await using var browser = await launcher.LaunchAsync(options);

    var tempUserDataDir = ((Browser)browser).Launcher.TempUserDataDir;
    DirectoryAssert.Exists(tempUserDataDir.Path);

    await browser.DisposeAsync();
    DirectoryAssert.DoesNotExist(tempUserDataDir.Path);
}
```

kblok:
@jnyrup do you want to give it another look?

jnyrup:
I'll have a look later today

lib/PuppeteerSharp/Helpers/TempDirectory.cs
```diff
@@ -29,37 +29,44 @@ private TempDirectory(string path)
...
+
+        public override string ToString() => Path;
+
+        /// <inheritdoc />
+        public async ValueTask DisposeAsync()
         {
+            lock (_disposingLock)
+            {
+                _deleteTask ??= DeleteAsync();
+            }
```
jnyrup, line 48: TLDR; using a lock  and ensuring we have a single task of `DeleteAsync` looks good.

My thoughts while looking at this:

When looking at other implementations of `DisposeAsync` we have different mindsets about how to handle re-entrance.

`TaskQueue.DisposeAsync` returns immediately on re-entrance and in a thread-safe manner.
https://github.com/hardkoded/puppeteer-sharp/blob/7d7e6f01e5af831d1b03986746d00af04b311da8/lib/PuppeteerSharp/Helpers/TaskQueue.cs#L32-L35

`DisposableTasksStack.DisposeAsync` also immediately returns on re-entrance, but in thread-unsafe manner, which seems fine when looking at its two usages.
https://github.com/hardkoded/puppeteer-sharp/blob/7d7e6f01e5af831d1b03986746d00af04b311da8/lib/PuppeteerSharp/DisposableTasksStack.cs#L44-L58

`ExecutionContext.DisposeAsync` on the other hand does not have special re-entrance handling.
https://github.com/hardkoded/puppeteer-sharp/blob/7d7e6f01e5af831d1b03986746d00af04b311da8/lib/PuppeteerSharp/ExecutionContext.cs#L72-L90

We for sure don't want multiple _different_ invocations of `DeleteAsync`.
I added the thread-safe variant using `Interlocked` in 45ed534343989ee358bd6f8d90bdf564924aab9b.
I'm not sure we _really_ need the thread-safety, but I suggest we keep it thread-safe.

Since `DeleteAsync` does sync IO (`Directory.Exists` and `Directory.Delete`) before reaching its first `await`, calling it inside a `lock` could be sub-optimal if the lock is expected to be congested.

So the remaining question I see is: how should `TempDirectory.DisposeAsync` behave on re-entrance?

Can re-entrant callers return immediately, expecting the first caller to complete the cleanup?
```cs
private int _disposed;
...

{
    {
        return;
    }

    await DeleteAsync().ConfigureAwait(false);
    _tcs.SetResult(true);
    GC.SuppressFinalize(this);
}
```

Do concurrent callers need to still await the call to the single `DeleteAsync` task to ensure that `Path` has been deleted when returning?

```cs
private readonly TaskCompletionSource<bool> _tcs = new();
private int _disposed;
...

{
    {
        await _tcs.Task.ConfigureAwait(false);
        return;
    }

    await DeleteAsync().ConfigureAwait(false);
    _tcs.SetResult(true);
    GC.SuppressFinalize(this);
}
```

lib/PuppeteerSharp/Helpers/TempDirectory.cs
```diff
@@ -29,37 +29,44 @@ private TempDirectory(string path)
...
             const int maxDelayInMillis = 8000;
 
             var retryDelay = minDelayInMillis;
             while (true)
             {
-                if (!Directory.Exists(path))
+                if (!Directory.Exists(Path))
                 {
                     return;
                 }
```
jnyrup, line 65: Can be simplified to 
`while (Directory.Exists(Path))`

lib/PuppeteerSharp/Helpers/TempDirectory.cs
```diff
@@ -29,37 +29,44 @@ private TempDirectory(string path)
...
-            DisposeCore();
-        }
+        ~TempDirectory() => Dispose();
 
         public string Path { get; }
 
-        public void Dispose()
+#pragma warning disable CA1816
+        public void Dispose() => _ = DisposeAsync();
+#pragma warning restore CA1816
```
jnyrup, line 38: We could delete `Dispose` and `IDisposable`.
This helps ensuring we call and await `DeleteAsync`, except if `~TempDirectory()` is called.

jnyrup (commented):
Disposal, async and concurrency combined is challenging to grasp.
(At least to me, but also good learning experience)

lib/PuppeteerSharp/States/DisposedState.cs
```diff
@@ -3,24 +3,18 @@
...
+        public override Task EnterFromAsync(LauncherBase launcher, State fromState, TimeSpan timeout)
         {
             if (fromState == StateManager.Exited)
             {
                 return Task.CompletedTask;
             }
 
-            Kill(p);
-
-            p.ExitCompletionSource.TrySetException(new ObjectDisposedException(p.ToString()));
```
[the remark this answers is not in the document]
kblok: Good catch. I though it was the same code

lib/PuppeteerSharp/Helpers/TempDirectory.cs
kblok: > TLDR; using a lock and ensuring we have a single task of `DeleteAsync` looks good.
> 
> My thoughts while looking at this:
> 
> When looking at other implementations of `DisposeAsync` we have different mindsets about how to handle re-entrance.

I think that, functionally speaking, all the calls to `DisposeAsync` should return the same task. But I couldn't find any code in `dotnet/runtime` to back me up.
That left me thinking that we are breaking some pattern here. 

The good news is that we could just simply call `DeleteAsync` instead of `DisposeAsync`.
So, I'm removing `IDisposableAsync`. The Launcher should call `DeleteAsync`. If everything goes as expected, the directory should be gone when disposed.

Thanks for your help man.

lib/PuppeteerSharp/Helpers/TempDirectory.cs
kblok: rollback

lib/PuppeteerSharp/Browser.cs
```diff
@@ -46,7 +45,7 @@ public abstract class Browser : IBrowser
         public abstract bool IsClosed { get; }
 
         /// <inheritdoc/>
-        public IBrowserContext DefaultContext { get; protected set; }
+        public IBrowserContext DefaultContext { get; protected init; }
```
jnyrup, line 48: FYI: Since `Browser` is `public` this is a breaking change, so you _might_ want to leave out the changes from `set` to `init` from this PR.

kblok: Yeah...

kblok, line 48:
```suggestion
        public IBrowserContext DefaultContext { get; protected set; }
```

lib/PuppeteerSharp/LauncherBase.cs
```diff
@@ -146,13 +146,19 @@ public async Task<bool> WaitForExitAsync(TimeSpan? timeout)
...
+        protected virtual void Dispose(bool disposing) => _stateManager.CurrentState.Dispose(this);
+
         /// <summary>
         /// Set Env Variables.
         /// </summary>
         /// <param name="environment">The environment.</param>
         /// <param name="customEnv">The customEnv.</param>
         /// <param name="realEnv">The realEnv.</param>
-        protected static void SetEnvVariables(IDictionary<string, string> environment, IDictionary<string, string> customEnv, IDictionary realEnv)
+        private static void SetEnvVariables(IDictionary<string, string> environment, IDictionary<string, string> customEnv, IDictionary realEnv)
```
jnyrup, line 161: FYI: This is also a breaking change

kblok: hahaha fair. I have to `seal` some classes.

kblok: changed

jnyrup (approved):

PR #2697, opened 2024-07-17
Align per-test and per-class disposal

jnyrup:
In NUnit the constructor and Dispose(Async) methods are run once per _class_, whereas methods marked with `[SetUp]` or `[TearDown]` are run once per _test_.

In `PuppeteerBrowserBaseTest` we hence:
* Before each test: construct a new `Browser`
* After each test: close `Browser`
* At the last test: dispose the last created `Browser`

Unless I'm missing something, each `Browser` instance should be handled equally.
b957d9ba2b38d366963cc482ed3caf504248ed54 introduced this asymmetry but I  don't understand why.

If we want to keep this asymmetry, let's at least add a comment why we close all browser, but only dispose the last one.

### DIFF
--- a/lib/PuppeteerSharp.Tests/PuppeteerBrowserBaseTest.cs
+++ b/lib/PuppeteerSharp.Tests/PuppeteerBrowserBaseTest.cs
@@ -1,42 +1,27 @@
-using System;
 using System.Threading.Tasks;
 using NUnit.Framework;
 
 namespace PuppeteerSharp.Tests
 {
-    public class PuppeteerBrowserBaseTest : PuppeteerBaseTest, IDisposable, IAsyncDisposable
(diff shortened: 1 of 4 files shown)


kblok:
@jnyrup ideally, I would love to keep one browser per `LaunchOption`! I think that's how it worked back in the `XUnit`. days.
If we get to do that, it would be awesome. If not. I would like to keep one browser per class rather than per test.

kblok:
@jnyrup  Let's go for NUnit4 when you can :)

jnyrup:
I won't get around looking at this any time soon.
(It also seems to me that the tests have been more stable the past months)

kblok:
Thank you, @jnyrup. I tried to make it work. But tests weren't happy about it

PR #2713, opened 2024-07-24
Migrate to system.text.json

kblok:
# Description

Work in progress. Early review is welcome


lib/PuppeteerSharp/PuppeteerSharp.csproj
```diff
@@ -29,13 +29,12 @@
   </PropertyGroup>
   <Import Project="../Common/CommonProps.props" />
   <ItemGroup>
+    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="8.0.0" />
```
jnyrup, line 32: The two new libaries added here should most likely only be included for `netstandard2.0`

lib/PuppeteerSharp.Tests/PuppeteerSharp.Tests.csproj
```diff
@@ -14,9 +14,9 @@
 		<PackageReference Include="NUnit" Version="3.13.3" />
 		<PackageReference Include="NUnit3TestAdapter" Version="4.5.0" />
 		<PackageReference Include="System.Net.Http" Version="4.3.4" />
-		<PackageReference Include="SixLabors.ImageSharp" Version="3.1.4" />
+		<PackageReference Include="SixLabors.ImageSharp" Version="3.1.5" />
 		<PackageReference Include="Microsoft.AspNetCore.Connections.Abstractions" Version="2.2.0" />
-		<PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="7.0.0" />
+		<PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="8.0.0" />
```
jnyrup, line 19: As we'we added `Microsoft.Bcl.AsyncInterfaces` in `PuppeteerSharp.csproj` we can delete it from here.

lib/PuppeteerSharp/Helpers/EnumHelper.cs
```diff
@@ -77,4 +78,39 @@ public static string ToValueString<TEnum>(this TEnum value)
...
+                if (field.GetCustomAttribute<EnumMemberAttribute>()?.Value is { } enumMember)
+                {
+                    dictionary[enumMember] = fieldValue;
+                }
+            }
+
+            return dictionary;
+        });
+
+        var defaultEnumAttribute = typeof(TEnum).GetCustomAttribute<DefaultEnumValueAttribute<TEnum>>();
```
jnyrup, line 103: Can this call to `GetCustomAttribute` be moved to after the lookup in `enumValues`?

lib/PuppeteerSharp/Helpers/Json/JsonStringEnumMemberConverter.cs
```diff
@@ -0,0 +1,116 @@
...
+/// <summary>
+/// Converts an enum value to or from a JSON string.
+/// </summary>
+public class JsonStringEnumMemberConverter : JsonConverterFactory
+{
+    /// <inheritdoc />
+    public override bool CanConvert(Type typeToConvert)
+        => typeToConvert != null &&
+           (typeToConvert.IsEnum || Nullable.GetUnderlyingType(typeToConvert)?.IsEnum == true) &&
+           typeToConvert.GetCustomAttributes(typeof(JsonStringIgnoreAttribute), true).Length == 0;
```
jnyrup, line 40: Instead of checking if `Length` of `GetCustomAttributes` is zero, can we use `Attribute.IsDefined`?

lib/PuppeteerSharp/Helpers/RemoteObjectHelper.cs
```diff
@@ -92,15 +94,15 @@ private static object ValueFromType<T>(JToken value, RemoteObjectType objectType
...
                         return null;
                     case RemoteObjectType.Number:
-                        return value.Value<T>();
+                        return value.GetInt32();
                     case RemoteObjectType.Boolean:
-                        return value.Value<bool>();
+                        return value.GetBoolean();
                     case RemoteObjectType.Bigint:
-                        return value.Value<double>();
+                        return value.GetInt64();
```
jnyrup, line 105: Just double checking that `Number` and `Bigint` should now be mapped to `Int32` and `Int64`, respectively

lib/PuppeteerSharp/PuppeteerSharp.csproj
```diff
@@ -29,13 +29,12 @@
...
     <PackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="8.0.0">
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
     <PackageReference Include="Microsoft.Extensions.Logging" Version="2.0.2" />
```
jnyrup, line 37: Do we need this library for .net8.0?

lib/PuppeteerSharp/Helpers/RemoteObjectHelper.cs
kblok: Good one. I restored `Bigint` to `GetDouble` and `Number` to `ToObject<T>`.

lib/PuppeteerSharp/PuppeteerSharp.csproj
kblok: Do you mean upgrading it to v8?

jnyrup: I wasn't on my dev machine, so I was just unsure if the types in this package are already available in net8

lib/PuppeteerSharp/Helpers/Json/HttpMethodConverter.cs
```diff
@@ -1,20 +1,59 @@
...
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace PuppeteerSharp.Helpers.Json;
 
-namespace PuppeteerSharp.Helpers.Json
+/// <summary>
+/// <see cref="HttpMethodConverter"/> will convert a <see cref="HttpMethod"/> to and from a string.
+/// </summary>
+public class HttpMethodConverter : JsonConverter<HttpMethod>
```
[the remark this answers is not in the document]
kblok: Good thought.

kblok:
@jnyrup @campersau build is (pretty) much green. If you have some time to review it I will really appreciate it!

My plan create a pre-release doc and  publish a beta version out of this branch, hoping for some users willing to test this out.

.github/workflows/dotnet.yml
```diff
@@ -84,7 +84,7 @@ jobs:
         Get-ChildItem -Path cert:\CurrentUSer\my | where { $_.friendlyname -eq "Puppeteer" } | Export-Certificate -FilePath $env:GITHUB_WORKSPACE\lib\PuppeteerSharp.TestServer\testCert.cer
     - name: Check formatting
       if: ${{ matrix.os == 'ubuntu-latest' && matrix.browser == 'CHROME' && matrix.mode == 'headless' }}
-      run: dotnet format ./lib/PuppeteerSharp.sln --verify-no-changes
+      run: dotnet format ./lib/PuppeteerSharp.sln --verify-no-changes --exclude-diagnostics CA1865
```
kblok, line 87: I had to exclude CA1865 because if generates invalid code on netstandard2.0

campersau:
Searching for "Newtonsoft" currently yields these results, which could be removed:

https://github.com/hardkoded/puppeteer-sharp/blob/aef54b35b6943c2cc621986d74709e22f3e959f3/lib/PuppeteerSharp/IPage.cs#L539

https://github.com/hardkoded/puppeteer-sharp/blob/aef54b35b6943c2cc621986d74709e22f3e959f3/lib/PuppeteerSharp.DevicesFetcher/PuppeteerSharp.DevicesFetcher.csproj#L9

https://github.com/hardkoded/puppeteer-sharp/blob/aef54b35b6943c2cc621986d74709e22f3e959f3/lib/PuppeteerSharp.Nunit/PuppeteerSharp.Nunit.csproj#L8

https://github.com/hardkoded/puppeteer-sharp/blob/aef54b35b6943c2cc621986d74709e22f3e959f3/lib/PuppeteerSharp.Tests/DeviceRequestPromptTests/MockCDPSession.cs#L27

https://github.com/hardkoded/puppeteer-sharp/blob/aef54b35b6943c2cc621986d74709e22f3e959f3/lib/PuppeteerSharp.Tests/DeviceRequestPromptTests/WaitForDevicePromptTests.cs#L6

https://github.com/hardkoded/puppeteer-sharp/blob/aef54b35b6943c2cc621986d74709e22f3e959f3/lib/PuppeteerSharp.Tooling/PuppeteerSharp.Tooling.csproj#L17

https://github.com/hardkoded/puppeteer-sharp/blob/aef54b35b6943c2cc621986d74709e22f3e959f3/samples/PupppeterSharpAspNetFrameworkSample/PupppeterSharpAspNetFrameworkSample/packages.config#L26

https://github.com/hardkoded/puppeteer-sharp/blob/aef54b35b6943c2cc621986d74709e22f3e959f3/samples/PupppeterSharpAspNetFrameworkSample/PupppeterSharpAspNetFrameworkSample/PupppeterSharpAspNetFrameworkSample.csproj#L127-L129

https://github.com/hardkoded/puppeteer-sharp/blob/aef54b35b6943c2cc621986d74709e22f3e959f3/samples/PupppeterSharpAspNetFrameworkSample/PupppeterSharpAspNetFrameworkSample/Web.config#L47-L50

lib/PuppeteerSharp/Helpers/Json/AnyTypeToStringConverter.cs
```diff
@@ -0,0 +1,71 @@
...
+            reader.Skip();
+            return "(not supported)";
+        }
+
+        throw new JsonException($"Unsupported token type: {reader.TokenType}");
+    }
+
+    public override void Write(Utf8JsonWriter writer, string value, JsonSerializerOptions options)
+    {
+        if (int.TryParse(value, out var number))
```
campersau, line 62: Also use `CultureInfo.InvariantCulture` here.

lib/PuppeteerSharp/Helpers/Json/DefaultEnumValueAttribute.cs
```diff
@@ -0,0 +1,31 @@
...
+//  * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+//  * SOFTWARE.
+
+using System;
+
+namespace PuppeteerSharp.Helpers.Json;
+
+[AttributeUsage(AttributeTargets.Enum)]
+internal class DefaultEnumValueAttribute<T>(T value) : Attribute
```
campersau, line 28: Could also be `sealed` maybe enable https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1852 for that.

lib/PuppeteerSharp/PuppeteerSharp.csproj
```diff
@@ -45,6 +42,11 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
+    <PackageReference Include="System.Text.Json" Version="8.0.4" />
```
campersau, line 45: Could this also be moved to `netstandard2.0`?
`System.Text.Json` is alread there for `net8.0`.

lib/PuppeteerSharp/TargetInfo.cs
```diff
@@ -7,54 +9,51 @@ namespace PuppeteerSharp
     /// </summary>
     public class TargetInfo
     {
-        internal TargetInfo()
+        /// <summary>
+        /// Initializes a new instance of the <see cref="TargetInfo"/> class.
+        /// </summary>
+        public TargetInfo()
```
campersau, line 15: Remove to use implicit default constructor.

campersau (commented):
Would be nice to remove the explicit STJ PackageReference on net8.0.

Also some other comments and suggestions since this is the first time I reviewed this code base in a bit more detail :)

lib/PuppeteerSharp.Tooling/PuppeteerSharp.Tooling.csproj
```diff
@@ -12,9 +12,8 @@
 		<PackageReference Include="CommandLineParser" Version="2.9.1" />
 		<PackageReference Include="System.CodeDom" Version="5.0.0" />
 		<PackageReference Include="Macross.Json.Extensions" Version="1.5.0" />
```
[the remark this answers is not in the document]
kblok: removed

lib/PuppeteerSharp/DragOperation.cs
```diff
@@ -1,8 +1,12 @@
+using System.Runtime.Serialization;
```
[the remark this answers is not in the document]
kblok: changed!

lib/PuppeteerSharp/Helpers/Json/AnyTypeToStringConverter.cs
kblok: Interesting... it's enabled

lib/PuppeteerSharp/Helpers/Json/DefaultEnumValueAttribute.cs
kblok: done!

lib/PuppeteerSharp/PuppeteerSharp.csproj
kblok: done!

lib/PuppeteerSharp/TargetInfo.cs
kblok: done!

jnyrup (commented):
I didn't look at the uses of attributes on types

lib/PuppeteerSharp.Nunit/PuppeteerTestAttribute.cs
```diff
@@ -177,7 +178,14 @@ private static TestExpectation[] LoadExpectationsFromResource(string resourceNam
...
             var fileContent = reader.ReadToEnd();
-            return JsonConvert.DeserializeObject<TestExpectation[]>(fileContent);
+            return JsonSerializer.Deserialize<TestExpectation[]>(fileContent, new JsonSerializerOptions()
+            {
+                PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
+                Converters =
+                {
+                    new JsonStringEnumMemberConverter(),
+                },
+            });
```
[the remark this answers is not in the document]
kblok: done!

lib/PuppeteerSharp.Tooling/PuppeteerSharp.Tooling.csproj
```diff
@@ -11,11 +11,8 @@
 		<FrameworkReference Include="Microsoft.AspNetCore.App" />
 		<PackageReference Include="CommandLineParser" Version="2.9.1" />
 		<PackageReference Include="System.CodeDom" Version="5.0.0" />
-		<PackageReference Include="Macross.Json.Extensions" Version="1.5.0" />
-		<PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="5.0.0" />
-		<PackageReference Include="Microsoft.Extensions.Primitives" Version="6.0.0" />
-		<PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="7.0.0" />
-		<PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
+		<PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="8.0.0" />
+		<PackageReference Include="Microsoft.Extensions.Primitives" Version="8.0.0" />
```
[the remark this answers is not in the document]
kblok: done!

lib/PuppeteerSharp/Helpers/EnumHelper.cs
```diff
@@ -77,4 +78,39 @@ public static string ToValueString<TEnum>(this TEnum value)
 
         return enumValues[value];
     }
+
+    public static TEnum FromValueString<TEnum>(string value)
+        where TEnum : struct, Enum
+    {
+        var enumValues = _stringToEnumCache.GetOrAdd(typeof(TEnum), type =>
```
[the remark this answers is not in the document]
kblok: Follow up PR :)

lib/PuppeteerSharp/Helpers/Json/JsonHelper.cs
```diff
@@ -1,14 +1,74 @@
...
+        public static readonly JsonSerializerOptions DefaultJsonSerializerSettings = new()
         {
-            ContractResolver = new CamelCasePropertyNamesContractResolver(),
-            NullValueHandling = NullValueHandling.Ignore,
+            PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
+            DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull,
+#if NET8_0_OR_GREATER
+            TypeInfoResolver = JsonSerializer.IsReflectionEnabledByDefault
+                ? new DefaultJsonTypeInfoResolver()
+                : SystemTextJsonSerializationContext.Default,
```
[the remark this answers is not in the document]
kblok: I think that with the conditional build that makes sense.

lib/PuppeteerSharp.Nunit/PuppeteerTestAttribute.cs
```diff
@@ -170,14 +171,24 @@ private static TestExpectation.TestExpectationPlatform GetCurrentExpectationPlat
         private static TestExpectation[] GetUpstreamExpectations() =>
             _upstreamExpectations ??= LoadExpectationsFromResource("PuppeteerSharp.Nunit.TestExpectations.TestExpectations.upstream.json");
 
+        private static JsonSerializerOptions DefaultJsonSerializerOptions =>
```
campersau, line 174: Needs to be a field to be a cached instance just like
https://github.com/hardkoded/puppeteer-sharp/pull/2713/files#diff-af9c020cb3f1613906500f6b77a8dd95e2f19221ed17509db28c47a5ff5dd55dR10
Since this is currently just a shortcut for `DefaultJsonSerializerOptions { get { return new() { ... }; } }` which would return a new instance every time.

```suggestion
        private static readonly JsonSerializerOptions DefaultJsonSerializerOptions =
```

kblok: oh yeah, good catch

lib/PuppeteerSharp/Helpers/Json/JsonHelper.cs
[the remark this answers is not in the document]
kblok: It seems that with this change, I would be forcing people to use declared values in the Context. 

In fact, I think that if we want `Evaluate` calls to work on AOT, you might need to use basic types or provide an extended context.

lib/PuppeteerSharp/Helpers/Json/SystemTextJsonSerializationContext.cs
```diff
@@ -0,0 +1,208 @@
...
+using System.Collections.Generic;
+using System.Text.Json.Serialization;
+using PuppeteerSharp.Cdp.Messaging;
+
+namespace PuppeteerSharp.Helpers.Json;
+
+/// <summary>
+/// We need this class for AOT.
+/// </summary>
+[JsonSourceGenerationOptions(WriteIndented = true)]
```
jnyrup, line 32: Do we need to write indented JSON?

kblok: not at all :)

lib/PuppeteerSharp/Helpers/Json/JsonHelper.cs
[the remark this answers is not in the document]
kblok: I confirm this. If the consumer doesn't enforce this. There is no need to do this.
If the consumer does enable this. They will also have to provide a Context for any extra types they use.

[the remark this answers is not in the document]
jnyrup: Fine with me 👍 
As a follow-up later, we could perhaps change it such that all types that don't rely on user types, are (de)serialized using a source-generated JSON serializer when using the .NET8 target.

kblok:
@jnyrup some updates:

We can't set the JsonTypeInfoResolver by default. That would force the user to do the same thing.
If you build your project with AOT, and you use some special types when you call some `Evaluate` function. You have to provide a serialization context using `Puppeteer.ExtraJsonSerializerContext`. You should do that as a first step. After that, the `DefaultJsonSerializerSettings` will be cached.

I added a demo in the console app.

Thoughts?

jnyrup:
I don't have that much experience with AOT and `JsonTypeInfoResolver`, so can't provide the best feedback here.

If I have two different types `A` and `B` and also two different source-generated `JsonSerializerContext`s, does the current design allow me two use them without using `Puppeteer.ExtraJsonSerializerContext`?

kblok:
> If I have two different types `A` and `B` and also two different source-generated `JsonSerializerContext`s, does the current design allow me two use them without using `Puppeteer.ExtraJsonSerializerContext`?

You could pass the result a `JsonTypeInfoResolver.Combine()` to `Puppeteer.ExtraJsonSerializerContext`.

PR #2717, opened 2024-08-04
Bump chrome to 127

kblok:


lib/PuppeteerSharp/BrowserFetcher.cs
```diff
@@ -261,7 +263,41 @@ private async Task<InstalledBrowser> DownloadAsync(SupportedBrowser browser, str
...
+                    process.StartInfo.FileName = setupExePath;
+                    process.StartInfo.Arguments = $"--configure-browser-in-directory=\"{browserDir.FullName}\"";
+                    process.StartInfo.RedirectStandardOutput = true;
+                    process.StartInfo.UseShellExecute = false;
+                    process.Start();
+                    process.WaitForExit();
+                }
+                finally
+                {
+                    // swallow
```
campersau, line 298: What does a finally swallow?
Upstream just does some logging there: 
https://github.com/puppeteer/puppeteer/blob/79a3d049608056ab8dd50756e01103ba6e76b35c/packages/browsers/src/install.ts#L301

kblok: Yeah. The main difference is that in puppeteer that's a CLI tool and here it's part of the code flow.

I added an optional logger :)

PR #2744, opened 2024-08-18
Replace small dictionaries with switch

jnyrup:
Instead of initializing and looking up values in small dictionaries, it's faster to simply use a switch expression and we can also (I would say) better name the lookup methods.

I implemented them as private static functions, but they also have been static local functions.

### DIFF
--- a/lib/PuppeteerSharp/BrowserData/InstalledBrowser.cs
+++ b/lib/PuppeteerSharp/BrowserData/InstalledBrowser.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Collections.Generic;
 using System.IO;
 
 namespace PuppeteerSharp.BrowserData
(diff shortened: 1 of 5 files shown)


lib/PuppeteerSharp/BrowserData/InstalledBrowser.cs
```diff
@@ -62,7 +53,16 @@ public string GetExecutablePath()
             var installationDir = Cache.GetInstallationDir(Browser, Platform, BuildId);
             return Path.Combine(
                 installationDir,
-                _executablePathByBrowser[Browser](Platform, BuildId));
+                GetExecutablePath(Browser)(Platform, BuildId));
         }
+
+        private static Func<Platform, string, string> GetExecutablePath(SupportedBrowser browser) => browser switch
```
jnyrup, line 59: We could also add the `platform` and `buildId` parameters and avoid the intermediate `Func<Platform, string, string>` which also makes the method signature prettier.

kblok:
Love it!

PR #2749, opened 2024-08-21
Use byte array in message send to avoid some unnecessary encoding

campersau:
Similar to #2748 System.Text.Json can also serialize directly to bytes to avoid some unnecessary [en](https://github.com/dotnet/runtime/blob/d3154144ae65a77585de29892ec41517f0add61e/src/libraries/System.Text.Json/src/System/Text/Json/Serialization/JsonSerializer.Write.String.cs#L164)/[decoding](https://github.com/campersau/puppeteer-sharp/blob/39f23de3c9b9f71178eb9d612bf66f3531bc0ec5/lib/PuppeteerSharp/Transport/WebSocketTransport.cs#L75).

### DIFF
--- a/lib/PuppeteerSharp.Tests/PollerInterceptor.cs
+++ b/lib/PuppeteerSharp.Tests/PollerInterceptor.cs
@@ -6,9 +6,9 @@ namespace PuppeteerSharp.Tests
 {
     public sealed class PollerInterceptor(IConnectionTransport connectionTransport) : IConnectionTransport
     {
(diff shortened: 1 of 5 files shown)


kblok (commented):
I'm afraid that I need to revert your previous PR.
Both are breaking changes. `IConnectionTransport` and `MessageReceivedEventArgs` are public.
I will point this PR to a `v20` branch. and create a new PR with the other change also pointing to `v20`.

Makes sense?

campersau:
Yes, makes sense.

PR #2767, opened 2024-09-03
Roll chrome to 128.0.6613.119

kblok:


ognjenst:
It would be nice to have this asap 👍

kblok:
> It would be nice to have this asap 👍

On the way :)

PR #2770, opened 2024-09-03
Use byte array in message received to avoid some unnecessary encoding (#2748)

campersau:
Looks like this commit got lost (reverted and not applied again) with the v20 merge.


kblok:
Oh yeah. It was removed on the  sync merge. Thank you!

PR #2783, opened 2024-09-18
Fix deserialization exception when retrieving cookie with PartitionKey set

AlexCSDev:
At some point CookieParam.PartitionKey [became an object](https://chromedevtools.github.io/devtools-protocol/tot/Network/#type-CookiePartitionKey) in DevTools protocol. This causes deserialization exception because it is currently defined as a string in puppeteer-sharp. This PR implements necessary changes to fix that.


kblok:
Thanks for jumping in @AlexCSDev.
Do you mind matching [this change](https://github.com/puppeteer/puppeteer/commit/8e6fd74de15c773ffd046b313b681a4afd162d38) upstream?

AlexCSDev:
I don't have much experience working with System.Text.Json, but I figured writing JsonConverter for this fix will be a decent solution?

I must mention that I did not test writing as I don't have any way of testing that quickly.

PR #2786, opened 2024-09-29
Implement IAsyncDisposable / IDisposable on IBrowserContext

campersau:
Makes it more convenient to use and aligns with upstream implementation:

https://github.com/puppeteer/puppeteer/blob/3ad2e45c295083de6fc72a5041138c620615b755/packages/puppeteer-core/src/api/BrowserContext.ts#L260-L268

In the second commit I also updated some tests. Looks like this wasn't done upstream so please let me know if I should remove it.


kblok (commented):
Changed the base branch to `v21` because this is a breaking change.

kblok (approved):

PR #2788, opened 2024-09-29
Use object for Top / Left / Bottom / Right in MarginOptions

campersau:
Because:

1. [Upstream `PDFMargin` uses `string | number`](https://github.com/puppeteer/puppeteer/blob/3ad2e45c295083de6fc72a5041138c620615b755/packages/puppeteer-core/src/common/PDFOptions.ts#L10-L15)
2. [`PDFOptions.Width` / `PDFOptions.Height` also use `string | number` upstream](https://github.com/puppeteer/puppeteer/blob/3ad2e45c295083de6fc72a5041138c620615b755/packages/puppeteer-core/src/common/PDFOptions.ts#L125-L132)
3. `PdfOptions.Width` / `PdfOptions.Height` using `object` https://github.com/hardkoded/puppeteer-sharp/blob/e14f59da8d95536e0002efa13f451f2dc5c1c762/lib/PuppeteerSharp/PdfOptions.cs#L67-L75
4. All of these properties were already going to `ConvertPrintParameterToInches` https://github.com/hardkoded/puppeteer-sharp/blob/e14f59da8d95536e0002efa13f451f2dc5c1c762/lib/PuppeteerSharp/Cdp/CdpPage.cs#L791-L807

---

`PdfOptionsShouldBeSerializable` didn't actually test `Width` / `Height` which meant that serialisation was broken by the migration to `System.Text.Json`.
I added a simple `PrimitiveTypeConverter` to support https://github.com/hardkoded/puppeteer-sharp/issues/1001

---

`PdfOptionsShouldWorkWithMarginWithNoUnits` didn't actually test the `ConvertPrintParameterToInches` logic so I have removed it. It was just testing what `PdfOptionsShouldBeSerializable` already did.

---

Also made `MarginOptions` / `PaperFormat` records.

### DIFF
--- a/lib/PuppeteerSharp.Tests/PageTests/PdfTests.cs
+++ b/lib/PuppeteerSharp.Tests/PageTests/PdfTests.cs
@@ -127,6 +127,8 @@ namespace PuppeteerSharp.Tests.PageTests
         {
             var pdfOptions = new PdfOptions
             {
+                Width = 100,
+                Height = 100,
                 Format = PaperFormat.A4,
                 DisplayHeaderFooter = true,
                 MarginOptions = new MarginOptions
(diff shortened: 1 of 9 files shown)


lib/PuppeteerSharp.Tests/PageTests/PdfTests.cs
```diff
@@ -143,27 +145,5 @@ public void PdfOptionsShouldBeSerializable()
             var newPdfOptions = JsonSerializer.Deserialize<PdfOptions>(serialized);
             Assert.That(newPdfOptions, Is.EqualTo(pdfOptions));
         }
-
-        [Test]
-        public void PdfOptionsShouldWorkWithMarginWithNoUnits()
```
kblok, line 148: This was testing something that was causing puppeteer to fail. Let's leave it so we prove that we can pass `"0"`.

campersau: I added i back. I am still not sure that it actually testes the `ConvertPrintParameterToInches` code path.

https://github.com/hardkoded/puppeteer-sharp/blob/cbff8a1d6eee3f2b32fced57406d38ac941fb88a/lib/PuppeteerSharp/Cdp/CdpPage.cs#L1217-L1257

kblok: That test should check [this change](https://github.com/hardkoded/puppeteer-sharp/commit/a4ce0cfab1d376bcb84c534f24a1762cd02db802#diff-b7e03018805bd7ee45bb7f4eec5eed50f785de53ec455aaa07a896af2c259a7aR1231)

kblok: You're right that test is wrong.

kblok: I turned that test into a (valid) more unit test. Let me know what you think.

PR #2801, opened 2024-10-22
Only get build ID if executable path is not supplied.

AshIzat:
fixes https://github.com/hardkoded/puppeteer-sharp/issues/2800

### DIFF
--- a/lib/PuppeteerSharp/Launcher.cs
+++ b/lib/PuppeteerSharp/Launcher.cs
@@ -51,14 +51,19 @@ namespace PuppeteerSharp
 
             EnsureSingleLaunchOrConnect();
             _browser = options.Browser;
-            var buildId = options.Browser switch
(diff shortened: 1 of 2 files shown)


kblok:
Great catch @AshIzat! 
Do you need a new version with this now? I can make it if you need it :)

AshIzat:
> Great catch @AshIzat! Do you need a new version with this now? I can make it if you need it :)

That would be incredibly useful actually thanks.

AshIzat:
> Do you
> 
> > > Great catch @AshIzat! Do you need a new version with this now? I can make it if you need it :)
> > 
> > 
> > That would be incredibly useful actually, thanks.
> 
> Do you want to replace all the `20.0.2` to `20.0.3` in the PuppeteerSharp.csproj file?

Done.

PR #2904, opened 2025-04-27
Restore runtime compatibility with System.Text.Json 6.0

leonardo-fernandes:
A few clients started reporting a MethodNotFound since v20.1.2, related to ValueIsEscaped property. Turned out these clients were using System.Text.Json 6.0 at runtime, due to an older dependency.
The MethodNotFound is thrown when Read is invoked (as it has to resolve all method references in its body before executing), so it wasn't caught by the try/catch in Read itself.
Moving the ValueIsEscaped call to another method, causes the MethodNotFound to be thrown only when that method is invoked, and is now caught by the try/catch. IMO this fixes compatibility with older versions of the dll without disrupting the codebase.

### DIFF
--- a/lib/PuppeteerSharp/Helpers/Json/LowSurrogateConverter.cs
+++ b/lib/PuppeteerSharp/Helpers/Json/LowSurrogateConverter.cs
@@ -20,12 +20,7 @@ internal class LowSurrogateConverter : JsonConverter<string>
 
         try
         {
-            if (reader.ValueIsEscaped)
-            {
-                value = JsonUnescape(value);
-            }
-
-            return value;
+            return JsonUnescapedValue(reader, value);
         }
         catch
         {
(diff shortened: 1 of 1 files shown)


jnyrup:
Just throwing in an alternative, not sure if it's better. 

Wrap the call to `ValueIsEscaped` in a method. 
If the JIT inlines the method, it might need to be annotated with NoInlining.

leonardo-fernandes:
> Wrap the call to ValueIsEscaped in a method.
I didn't write it that way because I thought it would be more artificial in terms of code readability. But I'm open to any changes.

> If the JIT inlines the method, it might need to be annotated with NoInlining.

I believe it wouldn't be inlined on the first invocation, which is when the MethodNotFound is thrown. So I don't think the annotation is strictly needed.

leonardo-fernandes:
@kblok and @jnyrup what is your opinion on this PR? Any questions or suggestions?

kblok (approved):

jnyrup (approved):

PR #2953, opened 2025-09-04
Improve Precision of Paper Sizes in Inches to 4 Decimal Places

jnyrup:
upstream: https://github.com/puppeteer/puppeteer/pull/13087


kblok:
@jnyrup is this blocking you? If so, commit a `csproj` change with a new version number, and I can publish it.

jnyrup:
It's not blocking me, we use a custom `PaperFormat`.
Just noticed that puppeteer had changed their values, while I was debugging why things moved _slightly_ around when upgrading from chrome 119 to 138.

PR #2972, opened 2025-10-19
Bidi: Implement last redirect

kblok:
Improved the `LastRedirect` property in the `Request` class (`Request.cs`) to correctly traverse the redirect chain and return the last redirect, enhancing accuracy in redirect handling.

### DIFF
--- a/lib/PuppeteerSharp/Bidi/Core/Request.cs
+++ b/lib/PuppeteerSharp/Bidi/Core/Request.cs
@@ -67,6 +67,7 @@ internal class Request : IDisposable
             {
                 redirect = redirect._redirect;
             }
+
             return redirect;
         }
     }


lib/PuppeteerSharp/Bidi/Core/Request.cs
```diff
@@ -58,7 +58,18 @@ private Request(BrowsingContext browsingContext, BeforeRequestSentEventArgs args
...
-    public Request LastRedirect { get; set; }
+    public Request LastRedirect
+    {
+        get
+        {
+            var redirect = _redirect;
+            while (redirect?._redirect != null)
+            {
+                redirect = redirect._redirect;
+            }
```
kblok, line 69:
```suggestion
            }

```

PR #3028, opened 2025-12-05
Don't assume Params type

### LINKED ISSUE #2943: [ICDPSession] Page.startScreencast causes subsequent operations to fail
Hey I have been looking at getting [Page.screencast](url) to work and no matter what I try I find that sending `Page.screencast` succeeds, but any other operation after that point fails. 

These two example differ only by this line:

```c#
await session.SendAsync("Page.startScreencast", new { format = "png", everyNthFrame = 1 });
```

The first one passes, the second one throws error.

```c#
using System;
using System.Threading.Tasks;
using NUnit.Framework;
using PuppeteerSharp;

[TestFixture]
public class AboutChromeDevToolsAndScreencasting
{
    [Test]
    public async Task Can_use_cdp_session_without_screencast()
    {
        await using var browser = await Puppeteer.LaunchAsync(new LaunchOptions { Headless = false });

        var session = await page.CreateCDPSessionAsync();

        session.MessageReceived += (_, args) => { Console.WriteLine($"[EVENT] {args.MessageID} {args.MessageData}"); };

        await page.GoToAsync("https://www.rnz.co.nz");

        Assert.That(await page.GetTitleAsync(), Is.EqualTo("RNZ - NZ News, Current Affairs, Audio On Demand"));
    }

    [Test]
    public async Task Page_enable_causes_subsequent_operations_to_fail()
    {
        await using var browser = await Puppeteer.LaunchAsync(new LaunchOptions { Headless = false });

        var session = await page.CreateCDPSessionAsync();

        /*

            Prints:

                [EVENT] Page.screencastVisibilityChanged {"visible":true}

        */
        session.MessageReceived += (_, args) => { Console.WriteLine($"[EVENT] {args.MessageID} {args.MessageData}"); };

        /*
            [!] This line causes all subsequent operations to fail

            PuppeteerSharp.NavigationException : Protocol error(Page.navigate): Target closed. (Connection failed to process System.Byte[].
                The JSON value could not be converted to System.String. Path: $.sessionId | LineNumber: 0 | BytePositionInLine: 10386..

        */
        await session.SendAsync("Page.startScreencast", new { format = "png", everyNthFrame = 1 });

        await page.GoToAsync("https://www.rnz.co.nz");

        Assert.That(await page.GetTitleAsync(), Is.EqualTo("RNZ - NZ News, Current Affairs, Audio On Demand"));
    }
}

```

What am I missing?

bgrainger:
Fixes #2943

Fix an assumption that `obj.Params` can always be deserialized to `ConnectionResponseParams`, and only make that conversion for  the two messages that need it (`Target.attachedToTarget`, `Target.detachedFromTarget`).

### DIFF
--- a/lib/PuppeteerSharp.Tests/ScreencastTests/PageScreencastTests.cs
+++ b/lib/PuppeteerSharp.Tests/ScreencastTests/PageScreencastTests.cs
@@ -1,5 +1,6 @@
 using System.Threading.Tasks;
 using NUnit.Framework;
+using PuppeteerSharp.Nunit;
 
 namespace PuppeteerSharp.Tests.ScreencastTests
 {


kblok:
Let's add a test expectation in `TestExpectations.local.json` to ignore it in firefox.

bgrainger:
Not completely sure of the required syntax in `TestExpecations`, but hopefully I excluded this correctly?

Also bumped the version, since it would be nice to get a fixed version published. Thanks!

lib/PuppeteerSharp.Tests/ScreencastTests/PageScreencastTests.cs
```diff
@@ -0,0 +1,47 @@
+using System.Threading.Tasks;
+using NUnit.Framework;
+
+namespace PuppeteerSharp.Tests.ScreencastTests
+{
+    public class PageScreencastTests : PuppeteerBrowserContextBaseTest
+    {
+        [Test]
```
kblok:
```suggestion
        [Test, PuppeteerTest("puppeteer-sharp", "Screencast", "can start screencast")]
```

lib/PuppeteerSharp.Nunit/TestExpectations/TestExpectations.local.json
```diff
@@ -62,6 +62,13 @@
     "parameters": ["firefox"],
     "expectations": ["FAIL"]
   },
+  {
+    "comment": "TODO Review screencasts in firefox",
+    "testIdPattern": "[screencast.spec] *",
```
kblok:
```suggestion
    "testIdPattern": "[puppeteer-sharp] Screencast*",
```

lib/PuppeteerSharp.Tests/ScreencastTests/PageScreencastTests.cs
```diff
@@ -0,0 +1,47 @@
+using System.Threading.Tasks;
+using NUnit.Framework;
```
kblok, line 2: We also need this :)

```suggestion
using NUnit.Framework;
using PuppeteerSharp.Nunit;
```

PR #3135, opened 2026-02-15
Conditional compilation for CDP implementation only

### LINKED ISSUE #3134: Version 21.0.0 drastically increases native AOT binary size
Hi team. I noticed an alarming increase in my binary size after upgrading to v21.

With no other changes, my binary size went from 10.3MB in v20.2.6 to 14.8MB in v21.0.0. That's a ~43% increase, which isn't trivial by any means.

I found it's because of WebDriverBiDi's JSON type metadata in WebDriverBiDiJsonSerializerContext. This adds ~1MB of code just for that class, plus about 2.5MB of System.Text.Json code.
Image

To resolve this, two (or more) solutions exist that may work well together:

**Specify `GenerationMode` in [WebDriverBiDiJsonSerializerContext.cs](https://github.com/webdriverbidi-net/webdriverbidi-net/blob/c9c7e66cf08239f4710e6924a8d321c8a10d4e2b/src/WebDriverBiDi/JsonConverters/WebDriverBiDiJsonSerializerContext.cs).**

Try adding `[JsonSourceGenerationOptions(GenerationMode = JsonSourceGenerationMode.Serialization)]` there. According to the [docs](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/source-generation-modes), it will increase speeds of serialization and will remove metadata from the generated code. There are listed side-effects, but none of them look relevant to the WebDriverBiDi implementation, since most types are quite simple. Trying this change on the WebDriverBiDi.Demo project, I saw a decent decrease in binary size (~2MB). On my own project, it removes about 700KB. Not fantastic, but it's something.
Image

**Separate WebDriverBiDi implementation.**

I know this isn't trivial but there's just a lot of additional code in my binary in the PuppeteerSharp and System.Text.Json assemblies for an addition that is somewhat redundant at this point. Is it possible to separate the implementation out into a different NuGet package? It might be manageable with a simple preprocessor directive and some conditional compilation.

steveberdy:
Creates a `PuppeteerSharp.Cdp` Nuget package that removes the WebDriverBiDi implementation and only leaves the CDP implementation. Useful for native AOT builds that don't want the excess binary size generated by WebDriverBiDi code.

Note: the build and Nuget push steps are both duplicated in the pipelines, as Nuget packages are cleared between builds.

Fixes #3134


kblok:
Clever

kblok:
How about some tool that validates that we never leak wedriver bidi there?

steveberdy:
@kblok I used ILSpy to decompile the dll from the generated nuget package earlier, and it had nothing in the PuppeteerSharp.Bidi namespace there at all.
image

If you check the PuppeteerSharp.csproj file, you'll see that I put a condition on referencing the WebDriverBiDi library in the first place. It won't even pull it in if we're doing CDP only.

You can also see that it trimmed out the handling of the launching and connecting when using WebdriverBidi in the Launcher class, as I intended with the preprocessor directives:
image
image

kblok:
> @kblok I used ILSpy to decompile the dll from the generated nuget package earlier, and it had nothing in the PuppeteerSharp.Bidi namespace there at all.
> 
> image
> 
> 
> 
> If you check the PuppeteerSharp.csproj file, you'll see that I put a condition on referencing the WebDriverBiDi library in the first place. It won't even pull it in if we're doing CDP only.
> 
> 
> 
> You can also see that it trimmed out the handling of the launching and connecting when using WebdriverBidi in the Launcher class, as I intended with the preprocessor directives:
> 
> image
> 
> image
> 
> 

Im thinking about the future. What can we do so I don't break you in the future?

steveberdy:
@kblok I'm writing a script right now to decompile the generated DLL and check to make sure it doesn't include the PuppeteerSharp.Bidi namespace or the WebDriverBiDi dependency.

steveberdy:
@kblok alright, check what I added. It checks if the resulting assembly include WebDriverBiDi and if we have any types related to such usage.

.github/workflows/PushNugetPackageToIntNugetOrg.yml
```diff
@@ -29,3 +29,16 @@ jobs:
       if: env.NUGET_TOKEN_TEST_EXISTS != ''
       run: |
         dotnet nuget push ./lib/PuppeteerSharp/bin/Release/*.nupkg --api-key ${{secrets.NUGET_INT_API_KEY}} --source https://apiint.nugettest.org/v3/index.json
+    - name: Build (CDP only)
+      run: |
+        dotnet build lib/PuppeteerSharp/PuppeteerSharp.csproj /p:DefineConstants=CDP_ONLY /p:Title=PuppeteerSharp.Cdp /p:PackageId=PuppeteerSharp.Cdp --configuration Release --no-restore
+        ls ./lib/PuppeteerSharp/bin/Release
+    - name: Check for WebDriverBiDi leakage (CDP only)
+      run: |
+        dotnet tools\csharp\check-leaks.cs lib\PuppeteerSharp\bin\Release
```
kblok: This will be too late. We would have already tagged the version. Let's add this as a new workflow (not the tagging, the check) like the Demo one.

kblok:
One more thing.
Can we overwrite the description, so people exploring packages in Nuget knows when to use this package?

.github/workflows/PushNugetPackageToIntNugetOrg.yml
steveberdy: @kblok done

kblok (approved):

kblok:
Great job!

steveberdy:
> One more thing. Can we overwrite the description, so people exploring packages in Nuget knows when to use this package?

Yep, that's overridden now.

kblok:
@steveberdy new build failed.

steveberdy:
@kblok which run? From what I see, they're still in progress and none have failed.

kblok:
> @kblok which run? From what I see, they're still in progress and none have failed.

We are good. I might have seen an old one.

kblok:
I will ship this tomorrow

PR #3149, opened 2026-02-18
Remove other nupkg packages when building PuppeteerSharp.Cdp

steveberdy:
This should prevent what happened in [this run](https://github.com/hardkoded/puppeteer-sharp/actions/runs/22124548441/job/63951625701). The `PuppeteerSharp.Cdp` package failed to publish because of that failure.


kblok:
Thank you!
I already pushed PuppeteerSharp.Cdp manually if you want to give it a try.

PR #3157, opened 2026-02-18
Add local fonts to OverridePermissions enum

ciaranodonnell:
`local-fonts` is a permission that isn't currently in the enum.

See docs for details:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Permissions-Policy/local-fonts

### DIFF
--- a/lib/PuppeteerSharp.Nunit/TestExpectations/TestExpectations.upstream.json
+++ b/lib/PuppeteerSharp.Nunit/TestExpectations/TestExpectations.upstream.json
@@ -481,13 +481,6 @@
     "comment": "Firefox does not support --remote-debugging-pipe argument"
   },
   {
-    "testIdPattern": "[frame.spec] Frame specs Frame.prototype.frameElement should handle shadow roots",
-    "platforms": ["darwin", "linux", "win32"],
-    "parameters": ["webDriverBiDi"],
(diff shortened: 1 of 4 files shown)


lib/PuppeteerSharp/OverridePermission.cs
```diff
@@ -94,5 +94,11 @@ public enum OverridePermission
         /// </summary>
         [EnumMember(Value = "local-network-access")]
         LocalNetworkAccess,
+
```
kblok:
```suggestion

```

lib/PuppeteerSharp/OverridePermission.cs
```diff
@@ -94,5 +94,10 @@ public enum OverridePermission
         /// </summary>
         [EnumMember(Value = "local-network-access")]
         LocalNetworkAccess,
+        /// <summary>
```
kblok, line 98:
```suggestion

        /// <summary>
```

kblok:
@ciaranodonnell You have to update `BidiBrowserContext.GetPermissionName`. I tried to do it myself, but I can't push to your branch.

kblok:
@ciaranodonnell v21.1.1 is up!

PR #3162, opened 2026-02-19
Fix and tune CI/CD for PuppeteerSharp.Cdp

steveberdy:
- Should fix Nuget errors with duplicate publishes (otherwise we can use --skip-duplicate)
- Removes unnecessary property overrides when building for the leak-check


kblok:
Nice. The last failure was due to a NuGet permission.

PR #3409, opened 2026-04-20
Remove RenderDocument from disabled Chrome features

kblok:
Implements changes from https://github.com/puppeteer/puppeteer/pull/14745

Removes `RenderDocument` from the list of disabled Chrome features in ChromeLauncher.


kblok:
Superseded by #3425, which implements the same upstream change without test ignores.

PR #3423, opened 2026-04-24
Implement upstream PR #14821: browser extensions API (#3404)

kblok:
## Summary

Implements [upstream PR #14821](https://github.com/puppeteer/puppeteer/pull/14821) which adds browser extension management APIs, plus fixes a prerender navigation race condition discovered during testing.

### Extensions API
- Renames `GetExtensionsAsync()` → `ExtensionsAsync()` across `IBrowser`/`Browser`/`CdpBrowser`/`BidiBrowser`
- Adds `TriggerExtensionActionAsync(Extension)` to `IPage`/`Page`/`CdpPage`
- Adds `extension-with-page` test asset with popup and service worker background
- Adds 6 new extension tests matching upstream `cdp/extensions.spec`: list extensions/properties, list workers, trigger action, activate popup page, extension pages
- Adds `Pipe = true` to `BrowserWithExtensionOptions` for test stability
- Adds `AutoAttachFilter` to `ChromeTargetManager` to exclude page targets from initial attach (matches upstream `Filter` option)

### Target Handling
- Filters `CdpBrowser.Targets()` to only return exposed, initialized targets (avoids exposing speculative/prerender targets)
- Adds `IsTargetExposed` guard to `OnTargetDestroyed`/`OnTargetCreated` handlers

### Prerender Navigation Race Fix
- **Root cause**: `Client_MessageReceived` is `async void` — `responseReceived` CDP event arrives while `requestWillBeSent` is still awaiting `GetFrameAsync`, so `OnResponseReceived` finds no request and drops the response
- **Fix 1**: In `OnResponseReceived`, when request is null, queue the response event instead of dropping it
- **Fix 2**: In `OnRequestAsync`, after storing the request, check for a queued response and process it immediately before firing the `Request` event
- **Fix 3**: In `OnResponseReceivedExtraInfoAsync`, when a queued group exists but the request isn't stored yet, save the extraInfo and leave the queued group for `OnRequestAsync` to handle
- Adds `FrameSwappedByActivation` subscription and `OnRequestServedFromCache` handler to `LifecycleWatcher`

## Test plan
- [x] Prerender tests: 6/6 pass (1 skipped for unrelated screencast reason)
- [x] Navigation tests: 67/67 pass
- [x] Network + RequestInterception tests: 154/154 pass
- [x] Extension tests: 9/9 pass
- [x] Page tests: 183/183 pass
- [x] Target tests: 20/20 pass
- [x] Build: 0 errors, 0 warnings


kblok:
Closing: this was created in error. The fix has been applied directly to PR #3404 (branch implement-upstream-change-14821).

PR #3472, opened 2026-05-30
Make ICdpHandle interface public

amaitland:
Would it be ok to change `ICdpHandle` access modifier from `Internal` to `Public`?

I was going to upgrade [PuppeteerSharp.Dom](https://github.com/ChromiumDotNet/PuppeteerSharp.Dom) to support the latest version and having access to the interface would make that easier. 

Previously I used `IJSHandle.RemoteObject` to determine details about the object, since that's been moved to moved to `ICdpHandle` using that interface instead would make the transition easy.

Thanks, for the fantastic project 👍


kblok:
Thank you!

PR #3520, opened 2026-07-27
Replace hand-rolled retry/timeout plumbing with RxSharp

kblok:
Upstream Puppeteer solves the "retry an action" and "wait for an event, racing a timeout and session-close" problems with rxjs (`firstValueFrom` + `raceWith` over `fromEmitterEvent` streams). We were reimplementing each of these by hand in C#, per method — `Locator.RunWithRetryAsync` was a manual retry loop with a linked `CancellationTokenSource` and a pile of catch clauses just to tell timeout apart from cancellation apart from "retry again"; `WaitForNetworkIdleAsync`, `WaitForRequestAsync`, `WaitForResponseAsync`, and `WaitForFrameAsync` each hand-rolled their own `TaskCompletionSource` + manual event teardown + timeout shape, once per protocol.

All of it now goes through [ReactiveExtensionsSharp](https://github.com/hardkoded/ReactiveExtensions-Sharp), a faithful RxJS port built for exactly this kind of swap. `Locator` uses its `RetryAndRaceWithSignalAndTimer` combinator; the four `Wait*` methods above (plus `Browser.WaitForTargetAsync`) read close to upstream's own shape — merge the live events with the existing-item snapshot, filter, race against timeout/cancellation/close.

A bigger change landed along the way: `WaitForRequestAsync`/`WaitForResponseAsync`/`WaitForFrameAsync`/`WaitForNetworkIdleAsync` used to be `abstract`, duplicated separately in `CdpPage` and `BidiPage`. Upstream implements these once, in the shared abstract `Page` class, since they only need generic event-emitter access. Comparing our port against upstream's actual code surfaced real gaps from that duplication, all fixed by unifying:

- `WaitForOptions.CancellationToken` was silently ignored by all four methods on both protocols, despite being a documented, public option.
- We raced against an internal session-closed task instead of the page's own `Close` event, which is what upstream actually races against.
- CDP's `WaitForNetworkIdleAsync` ignored the `Concurrency` option entirely — only Bidi's separate implementation honored it.
- CDP's `WaitForFrameAsync` never checked for an already-matching frame before waiting; only Bidi did.

In-flight request tracking for `WaitForNetworkIdleAsync` now lives for the page's whole lifetime (wired once in the constructor) instead of being rebuilt fresh on every call, mirroring upstream's own `#inflight$` — a fresh-per-call counter would miss requests that were already in flight before the call.

Two real bugs also came up mid-port, both from .NET event delivery being able to run on another thread (unlike upstream's single-threaded JS, where the equivalent race can't exist): a matching event firing in the gap between attaching a handler and this code actually subscribing to the Rx pipeline. Fixed with `FromEventBuffered`, a library primitive that attaches immediately and buffers into a `ReplaySubject` — reusable instead of hand-rolled per call site.

Same external behavior as before everywhere else — same exception types/messages, same timing semantics. `ScreenRecorder` was deliberately left out of this pass; its `Channels`-based design doesn't share the shape Rx solves here.

## Test plan
- [x] All `LocatorTests`, `WaitForNetworkIdleTests`, `WaitForRequestTests`, `WaitForResponseTests`, `WaitForFrameTests`, `BrowserWaitForTargetTests`, and `CloseTests` pass, repeatedly, under both Chrome/CDP and Firefox/BiDi
- [x] Full suite run clean (1358 passed, 3 failed / 67 skipped — the 3 failures are pre-existing and unrelated: incognito service-worker detection, Chrome stderr capture, and a golden-image screenshot flake, all confirmed unchanged on unmodified master)
- [x] Builds clean on `netstandard2.0`/`net8.0`/`net10.0`, both the default and `CDP_ONLY` configs

https://claude.ai/code/session_01K92eapPm8e7mX7puT4cF4s


lib/PuppeteerSharp/Browser.cs
```diff
@@ -136,42 +142,46 @@ public async Task<ITarget> WaitForTargetAsync(Func<ITarget, bool> predicate, Wai
...
+            // target created in that gap is never observed - TargetCreated/TargetChanged are driven by the CDP
+            // message pump on its own thread, so that gap is real on .NET even though it can't happen in
+            // upstream's single-threaded JS.
+            //
+            // bufferSize is deliberately more than the default of 1: with just 1, if a matching target's event
+            // and a later non-matching target's event both land in that same narrow pre-subscribe gap, the
+            // size-1 buffer keeps only the most recent (non-matching) one, silently losing the match - unlike
+            // the old TrySetResult-based implementation, which processes every firing and lets the first match
+            // win regardless of how many events arrive before anyone awaits it. EventBufferSize gives enough
+            // headroom to absorb a realistic burst in that gap without buffering unboundedly for the whole wait.
```
kblok:
```suggestion
```

PR #3562, opened 2026-08-20
Nice to use .js files along with .cs files

Markot4:
Before I could do anything, I had to make sure I read Contribution.MD first. It was annoying setting it all up and had to ask one of GPTs to help me with how I should write it.

Since Building solution didn't really help with finding errors so I manually tried to find them. I instead got warnings and decided to fix them anyways.

For injected.js, I got so many warnings like "unexpected var, use let or const instead", "extract the assignment of "desc" from this expression", "make sure that this dynamic injection or execution of code is safe" and "invalid loop. its body allos only one iteration"

Examples of changes:

Summary of changes
• Suppressed S1523 (dynamic code execution) by adding an explicit NOSONAR marker on the dynamic evaluation line and documenting the reason. • Fixed S1751 (loop that only allows one iteration) by replacing the immediate-return-for-of with an iterator.next() extraction to get the first element. • Replaced remaining safe var declarations with const or class declarations and used let where a variable is intentionally reassigned/uninitialized (e.g., textChangeObserver). • Extracted the inline assignment for the property descriptor into a local const (S1121).

And for ScreenOrientationType.cs, I marked the class as static (internal static class ScreenOrientationType), which prevents instantiation and satisfies the analyzer.

If there is anything else you want me to do, feel free to comment on my fork. Please, do not immediately block me in case I did something really wrong. I tried to follow the rules.

### DIFF
--- a/lib/PuppeteerSharp/Injected/injected.js
+++ b/lib/PuppeteerSharp/Injected/injected.js
@@ -1,45 +1,46 @@
 "use strict";
-var __defProp = Object.defineProperty;
-var __getOwnPropDesc = Object.getOwnPropertyDescriptor;
-var __getOwnPropNames = Object.getOwnPropertyNames;
-var __hasOwnProp = Object.prototype.hasOwnProperty;
-var __export = (target, all) => {
-    for (var name in all)
+const __defProp = Object.defineProperty;
+const __getOwnPropDesc = Object.getOwnPropertyDescriptor;
+const __getOwnPropNames = Object.getOwnPropertyNames;
+const __hasOwnProp = Object.prototype.hasOwnProperty;
+const __export = (target, all) => {
+    for (const name in all)
         __defProp(target, name, { get: all[name], enumerable: true });
 };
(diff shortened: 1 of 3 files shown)


kblok:
Thank you!
I will be away from the keyboard for the next few days.
I like the change in the cs file.
The injected.js file is an exact copy of the code puppeteer injects, it's a webpack of many typescript classes. So I wouldn't change that and ignore it in the checker tools if necessary.

lib/PuppeteerSharp/Injected/injected.js
```diff
@@ -1,45 +1,46 @@
 "use strict";
-var __defProp = Object.defineProperty;
```
kblok, line 2: This one is conflicted now. I think we could merge it. I have this [TODO](https://github.com/hardkoded/puppeteer-sharp/blob/c7973c3ce3ef2dad1f97415cb55ddefc71dc845f/lib/PuppeteerSharp/Injected/README.md?plain=1#L7). We should have a tool to keep this file updated automatically.

PR #3589, opened 2026-09-15
Update WebDriverBiDi.NET dependency to v0.0.61

jimevans:
WebDriverBiDI.NET `0.0.61` slightly changed the Connection API, sealing the `IsActive` property, and adding a `protected abstract` property named `IsConnectionOpen`, which `IsActive` reads. To update the dependency, the previous override of `IsActive` is moved to be an implementation of `IsConnectionOpen`.


jimevans:
@kblok This should be ready to merge now.

